PR #46, opened 2021-05-11
As a repo contributor, I want to understand how large decisions are made, so that I can contribute to making them myself

NatHillardUSDS:
First stab at a proposal for an Architecture Decision Record process. I will be emailing about this shortly to solicit feedback on this process itself.


docs/decisions/0003-adr-process.md
```diff
@@ -0,0 +1,109 @@
...
+### Process
+
+Please see the below diagram:
+![ADR Process](./0003-files/adr_process-mmd.svg)
+
+1. **Start**: To create a new ADR, you have a few options:
+
+   1. **Old fashioned file copy**: Copy `template.md` into a new file in this directory, named `XXXX-name-of-issue.md` where the `XXXX` part is incremented from the previous ADR.
+
+   2. **ADR Tools**: Install `adr-tools` with `brew install adr-tools` then `adr new "Issue Name"` to make a new file. Copy contents of `template.md` in its place to get the right fields.
```
switzersc-usds, line 57: I would add a bit about making sure you're in the right directory:

 Install `adr-tools` with `brew install adr-tools`. Then, in the `docs/decisions` directory, run `adr new "Issue Name"` to make a new file. Do not add anything to the opened markdown file yet; instead, copy the contents of `template.md` and paste into the new file to get the right fields.

switzersc-usds: Actually, now that I've done this, I kind of feel like there isn't much benefit to using adr-tools since you still have to copy paste. You just get the right filename. Is there any other reason to use it?

switzersc-usds (commented):
Want to discuss the inclusion of the adr-tools package before moving forward. Not sure if it's worth including here because it may just end up confusing folks.

docs/decisions/0003-adr-process.md
NatHillardUSDS: It's true, not using their default template takes a lot of the benefit away. I can remove for now.

High-level I'm beginning to wonder about capturing at least some of this using a project board that you can move issues around on and issue templates - we can discus this part further also

switzersc-usds (approved):

Issue #6, opened 2021-04-20
Analysis of Alternatives: Front end site options


switzersc-usds:
**Description**
We need to identify what our front end stack is for our informational site and app. 

Here are the features we're looking to prioritize on the front end:

- smooth integration with our map client (open layers)
- accessibility support
- responsive
- fast
- easy content updates by non-technical team member (in markdown at minimum, ideally with a CMS integration)
- offline support / usable with low-bandwidth connections
- can be hosted as a static site, eg Github Pages (this makes deployment and authorization easier esp in short term), but scalable so we can add more interactive, dyanmic app features later as needed
- support for users wanting to print pages as PDFs
- support for localization tools
- use of US Web Design System for styles

**Definition of done**
Architecture Decision Record created and approved

PR #47, opened 2021-05-12
Client-side framework decision 

switzersc-usds:
This is the ADR for our front end client framework.

This addresses https://github.com/usds/justice40-tool/issues/6


NatHillardUSDS (approved):
LGTM!

PR #65, opened 2021-05-19
Update compile_mermaid.yml

NatHillardUSDS:
Our mermaid-compiling github action currently fails with an exit code of 1 if it attempts to commit files but there haven't been any changes. We should be resilient to this by checking the exit code of `git diff-index`


TomNUSDS (approved):
lgtm

Issue #16, opened 2021-04-29
As a member of the public, I want to learn about Justice40.


switzersc-usds:
**Description**
We should put up a lightly-styled, representative-content version of our initial site for the sake of forming an informational landing page. 

**Solution**
Following the creation of our static site (https://github.com/usds/justice40-tool/issues/17), we should add some approved content and some basic styling utilizing the US Web Design System.

**Describe alternatives you've considered**
* Not posting any content to begin with -- we need to give the public a feel for the messaging and design of our eventual site right out of the gate. 

**Links to user research or other resources**
N/A

**General Tasks**
- [ ] Get content approved
- [ ] Get translation approved

**Tech Tasks**
- [x] Re-generate Gatsby base page with minimal plugin setup- @NatHillardUSDS 
- [x] Integrate [gatsby-plugin-intl](https://www.gatsbyjs.com/plugins/gatsby-plugin-intl/?=i18n) for internationalization - @NatHillardUSDS 
- [x] Page Layout - @TomNUSDS 
- [x] Mobile support - @TomNUSDS 
- [x] Code cleanup to remove older map libraries - @NatHillardUSDS 
- [x] Implement accessibility best practices a la [here](https://www.gatsbyjs.com/docs/conceptual/making-your-site-accessible/) - @NatHillardUSDS 
- [x] Add styles via sass and module imports
- [x] Add [@trussworks/react-uswds](https://github.com/trussworks/react-uswds) - @TomNUSDS  
- [x] Configure Typescript - @NatHillardUSDS 

**Definition of "Done"**
- [ ] We've posted content in English and Spanish
- [x] We've styled that content appropriately according to a related spec. 
- [x] We have a website we can spin up locally. [This card is not about deploying]

**Design Needs**
* Reduced version of site spec (perhaps a limited version of earlier-discussed prototype, with a focus on content and content placement?)

switzersc-usds:
Aiming to have wireframes by EOW.

janelien:
Content is still a work in progress

![J40_informational site.png](https://images.zenhubusercontent.com/60998bc7b85eebae77bd2048/9bcef99b-861c-4e02-8f49-da261760295d)

![J40_informational site_mobile.png](https://images.zenhubusercontent.com/60998bc7b85eebae77bd2048/65b0b7c3-4d33-4590-af02-4e9698010abb)

switzersc-usds:
[...]

PR #67, opened 2021-05-20
Adding Trussworks USWDS React Wrapper

NatHillardUSDS:
This change, for issue #16 , covers adding the [Trussworks](https://github.com/trussworks/react-uswds) USWDS react wrapper for the USWDS [design system](https://designsystem.digital.gov/). 

It works, but there are a few things that are less than ideal at the moment:
* I'm wondering whether we can get by without a global styles file?
* Trussworks seems to misbehave with regard to SSR. If you simply include it, on the *second* load of the page you will see "FIle is not defined"
* To work around the above, based on advice [here](https://www.gatsbyjs.com/docs/debugging-html-builds/#fixing-third-party-modules), I added a bit to `gatsby-node.js` to set its loader to `null` when doing SSR. 
* This in turn results in a different error, ```
Warning: React.createElement: type is invalid -- expected a string   (for built-in components) or a class/function (for composite components) but got: undefined.    at IndexPage```. I am not sure what causes this
* To work around the above, I had to temporarily set `DEV_SSR` to `false` in `gatsby-config`. 

Note! This unblocks us from `gatsby develop`, but the developer-side SSR was put in to help make local debug better match the prod environment, as outlined [here](https://github.com/gatsbyjs/gatsby/discussions/28138). You can see this when you run `gatsby build`, which is what we need to run to put the site into production. This throws the same warning that the app does with `DEV_SSR` true - `Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.` . 

Anyhow, the `DEV_SSR: false` workaround works for now, we will need to revisit this before we integrate with CI/CD


NatHillardUSDS:
This is actually caused by [this issue](https://github.com/trussworks/react-uswds/issues/1250) which there is a pending [fix](https://github.com/trussworks/react-uswds/pull/1251) for that should be merged shortly. For now we can live with the workaround!

TomNUSDS (approved):
LGTM

PR #68, opened 2021-05-20
Completed Google Form script generator

esfoobar-usds:
This Google Form Script can be dropped on any Google Drive and run to generate the Justice 40 ingestion Form.

Pending work: assign a Google Sheet by default.

### DIFF
--- a/data-roadmap/data_roadmap/data_set_description_field_descriptions.yaml
+++ b/data-roadmap/data_roadmap/data_set_description_field_descriptions.yaml
@@ -1,39 +1,23 @@
 # There is no method for adding field descriptions to `yamale` schemas.
 # Therefore, we've created a dictionary here of fields and their descriptions.
-name: A short name of the data set.
-source: The URL pointing towards the data set itself or more information about the
-  data set.
-relevance_to_environmental_justice: It's useful to spell out why this data is
-  relevant to EJ issues and/or can be used to identify EJ communities.
-spatial_resolution: Dev team needs to know if the resolution is granular enough to be useful
-public_status: Whether a dataset has already gone through public release process
-  (like Census data) or may need a lengthy review process (like Medicaid data).
-sponsor: Whether there's a federal agency or non-governmental agency that is working
-  to provide and maintain this data.
-subjective_rating_of_data_quality: Sometimes we don't have statistics on data
-  quality, but we know it is likely to be accurate or not. How much has it been
-  vetted by an agency; is this the de facto data set for the topic?
-estimated_margin_of_error: Estimated margin of error on measurement, if known. Often
-  more narrow geographic measures have a higher margin of error due to a smaller sample
-  for each measurement.
-known_data_quality_issues: It can be helpful to write out known problems.
-geographic_coverage_percent: We want to think about data that is comprehensive across
-  America.
-geographic_coverage_description: A verbal description of geographic coverage.
+name: A short name for the data set.
+source: The URL / website where this data set can be found.
+relevance_to_environmental_justice:
+  Reason(s) why this data set is important for the environmental justice
+  community (e.g. helps identify an EJ community)
 data_formats: Developers need to know what formats the data is available in
-last_updated_date: When was the data last updated / refreshed? (In format YYYY-MM-DD.
+spatial_resolution: How specific does this data set get?
+public_status: Has this data set gone through a public release process?
+sponsor: Who is providing and maintaining this data? (e.g. a federal agency, a specific organization)
+subjective_rating_of_data_quality: Has this data set been vetted?
+estimated_margin_of_error: Estimated margin of error on measurement, if known.
+known_data_quality_issues: Are there any known issues with this data set?
+geographic_coverage_percent: What percentage of the United States and US territories does this data set cover? This can be an estimate.
+geographic_coverage_description: Tell us more about the geographic coverage.
+last_updated_date:
+  When was the data last updated or refreshed? (In format YYYY-MM-DD.
   If exact date is not known, use YYYY-01-01.)
-frequency_of_updates: How often is this data updated? Is it updated on a reliable
-  cadence?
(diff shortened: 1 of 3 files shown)


lucasmbrown-usds:
This is the same issue we were talking about earlier today, but I get nervous about having to maintain duplicate copies of the same schema. As I see it now, we've got one copy of the schema in the `data_set_description_schema.yaml`, and then one copy is spread throughout the source code (e.g. where we write the options such as "Esri Shapefile"). 

Is it possible to autogenerate those options from the schema itself, so we don't have to update the options in two places everytime there's a change, we just run a script? 

Thanks!

esfoobar-usds:
> This is the same issue we were talking about earlier today, but I get nervous about having to maintain duplicate copies of the same schema. As I see it now, we've got one copy of the schema in the `data_set_description_schema.yaml`, and then one copy is spread throughout the source code (e.g. where we write the options such as "Esri Shapefile").

> Is it possible to autogenerate those options from the schema itself, so we don't have to update the options in two places everytime there's a change, we just run a script?

Yes. I mentioned on the meeting that we would do this initial phase first of having the script file on the repo so that we can have the form by Monday and then next week start to work on a Python script to generate the Google Script programatically from the YAML definition file.

lucasmbrown-usds:
> Yes. I mentioned on the meeting that we would do this initial phase first of having the script file on the repo so that we can have the form by Monday and then next week start to work on a Python script to generate the Google Script programatically from the YAML definition file.

Ok. I'm also a little nervous that this is only a create script; after it's run the very first time, what is our plan for how we'll run this again? Are we able to use create or update functionality so that you can update the same form "in place", or are we planning to distribute a new link each time the form is updated (perhaps obscured behind a link shortener or something)? 

Thanks! Overall looking good!

esfoobar-usds:
> Ok. I'm also a little nervous that this is only a create script; after it's run the very first time, what is our plan for how we'll run this again? Are we able to use create or update functionality so that you can update the same form "in place", or are we planning to distribute a new link each time the form is updated (perhaps obscured behind a link shortener or something)?

[Per the caveats here](https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/39#issuecomment-846159519), there is no way that I've found where the script run can update the same form or point to the same sheet, so these are factors that push me to believe we're going to run into issues pretty quickly, unless we commit to creating a proper submission app.

esfoobar-usds:
Hi all, I have updated the script with the following changes:

- The script now generates one single Google Form and updates it
- The script writes to one single Google Sheet, although it creates a new Sheet (tab) every time the script is run
- Added descriptions to all fields
- Added instructions on Google Sheet

You can q/a the new form [here](https://docs.google.com/forms/d/e/1FAIpQLSe9EvOe90OWyrBVGo5QbCQX9Orz3lOjWi9rgBAewUQUb6PFoA/viewform).

data-roadmap/data_roadmap/data_set_description_schema.yaml
```diff
@@ -2,23 +2,26 @@
...
 relevance_to_environmental_justice: str(required=False)
-data_formats: enum('GeoJSON', 'Esri Shapefile (SHP, DBF, SHX)', 'GML', 'KML/KMZ',
-  'GPX', 'CSV/XLSX', 'GDB', 'MBTILES', 'LAS', required=True)
-spatial_resolution: enum('State/territory', 'County', 'Zip code', 'Census tract',
-  'Census block group', 'Exact address or lat/long', 'Other', required=True)
-public_status: enum('Not Released', 'Public', 'Public for certain audiences', 'Other',
-  required=True)
-sponsor: str(required=True)
+data_formats:
+  enum('GeoJSON', 'Esri Shapefile (SHP, DBF, SHX)', 'GML', 'KML/KMZ',
```
lucasmbrown-usds, line 6: Nice cleanup!

data-roadmap/data_roadmap/data_set_descriptions/google_form_gen.js
```diff
@@ -1,15 +1,31 @@
 function createForm() {
+  var form = FormApp.openByUrl(
+    "https://docs.google.com/forms/d/1Kb9Hl82I9kPJhqS_3bQDbWWvzCxPDTZBaI7_Eo6339E/edit"
```
lucasmbrown-usds, line 3: Very nice!

ginabeena:
Re: schema -- we could easily have a [type provider](https://docs.microsoft.com/en-us/dotnet/fsharp/tutorials/type-providers/) or [protobuf](https://developers.google.com/protocol-buffers) format and hook host the schema registry. Would be a high value add early on that will amortize work and provide high ROI. That way you don't have stringly type inconsistencies, you can version it, you can do some binary lensing into header values, etc.

data-roadmap/data_roadmap/data_set_description_field_descriptions.yaml
```diff
@@ -1,39 +1,23 @@
 # There is no method for adding field descriptions to `yamale` schemas.
 # Therefore, we've created a dictionary here of fields and their descriptions.
-name: A short name of the data set.
```
switzersc-usds, line 3: We've ended up updating the descriptions directly in the Google spreadsheet. We should update this yaml file with those descriptions.

data-roadmap/data_roadmap/data_set_description_schema.yaml
switzersc-usds: We need to add data formats to this, eg TXT. Maybe this shouldn't be an enum yet.

switzersc-usds (changes requested):
We should update the YAML field descriptions with the latest from the sheet.

switzersc-usds:
I will look into this a bit more in depth before sprint planning to see what is left here and how to update the card.

switzersc-usds:
Closing this as we have removed the data-roadmap directory for now. We can revisit this feature set at a later date.

PR #69, opened 2021-05-21
Typed Sass Styling

NatHillardUSDS:
* Add styles via sass and module imports
* Adds typed scss imports via gatsby-plugin-sass.
* Makes use of .d.scss file generated by gatsby-plugin-scss-typescript, but avoids importing that directly while they work out [an issue](https://github.com/Debens/gatsby-plugin-scss-typescript/issues/24) in gatsby v3.
* Creates basic layout component and two constituent components to illustrate the use of modules

### DIFF
--- a/client/.vscode/launch.json
+++ b/client/.vscode/launch.json
@@ -1,0 +1,34 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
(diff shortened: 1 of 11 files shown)


TomNUSDS (approved):
LTGM.

PR #70, opened 2021-05-21
Bug : staticrender results in page failing to load

NatHillardUSDS:
Temporarily removing the use of useStaticQuery as it seems to have introduced a bug in page rendering. Hard-coding for now; incidentally we will need to load this as a string anyway for localization purposes.

Also removing an attempt to use absolute module paths, to be readdressed later.


TomNUSDS (approved):
LGTM

PR #71, opened 2021-05-21
Integrate Internationalization/Localization via gatsby-plugin-intl

NatHillardUSDS:
* Addresses #16 subtask `Integrate gatsby-plugin-intl for internationalization`
* Replaces static text with `intl.formatMessage` call and links with gatsby-plugin-intl equivalent (we should use this precedent going forward)
* Two things to note:
  * It is necessary now to run `npm install --force` until [this PR](https://github.com/wiziple/gatsby-plugin-intl/pull/173) is addressed. We can consider using a fork as well, such as [this one](https://www.npmjs.com/package/gatsby-plugin-react-intl), or using FormatJS [directly](https://formatjs.io/docs/getting-started/installation/)
  * For easy reference I have translated the title; going forward we'll probably want to keep this branding for all languages

Screenshots:
Screen Shot 2021-05-21 at 12 49 06 PM
Screen Shot 2021-05-21 at 12 49 02 PM


client/src/pages/404.tsx
```diff
@@ -1,5 +1,5 @@
-import * as React from "react"
-import { Link } from "gatsby"
+import * as React from "react";
+import { Link } from "gatsby-plugin-intl";
```
TomNUSDS, line 2: Interesting that Link is replaced this way.

client/src/intl/en.json
```diff
@@ -0,0 +1,6 @@
+{
+  "71L0pp": {
```
TomNUSDS, line 2: Is this ID random?

TomNUSDS (approved):
LGTM

client/src/intl/en.json
NatHillardUSDS: Yeah - I documented this in the readme as well; it's producted by the `formatjs extract` command -- unfortunately this requires you to have to go back and put it back in later. I think there are alternatives to this through `babel` and/or `eslint` we could look into as well but this may have to do in the meantime.

PR #73, opened 2021-05-21
Initial attempt at static deploy to github pages

NatHillardUSDS:
First stab at a deploy to Github Pages. This is necessarily temporary during development and will not be where our site lives. Notably, we should be able to reuse this infrastructure in the future regardless of:
1. Whether we continue with gatsby or another static site framework, or
2. Whether we post to gh pages

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,52 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: GH Pages deploy
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16.x]
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,52 @@
...
+    needs: build
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+      - name: Download Artifacts
+        uses: actions/download-artifact@master
+        with:
+          name: J40Static
+          path: public
+      - name: Deploy 🚀
```
TomNUSDS, line 47: Is unicode safe for the name? I guess it is if it works.

TomNUSDS (dismissed):
LGTM

esfoobar-usds (dismissed):
🍏

NatHillardUSDS:
Unfortunately this is blocked for now -- `gatsby build` is failing. 

This is due to two factors, as far as I can tell so far:
1. The previously-mentioned Trussworks SSR bug (1250)
2. `gatsby-plugin-intl` appears to have introduced some dynamism which needs to be tracked down - we're seeing `error Building static HTML failed for path "/es/"` during the [latest build](https://github.com/usds/justice40-tool/runs/2656713952?check_suite_focus=true)

I will track these down today and remove the `draft` keyword

.github/workflows/node.js.yml
NatHillardUSDS: Good callout - It does seem acceptable to have them - this was taken from a tutorial but I've seen emoji in a few other projects also.

TomNUSDS (approved):
LGTM. 
We need to remember to fix the trussworks dependency, but the fix reference is such an ugly string, I'm sure it won't be a problem.

PR #74, opened 2021-05-24
Footer template basically working

TomNUSDS:
Part of #16

Still struggling to get trussworks to render the `<Address>` center aligned.

Todo: Need to figure out localization.


esfoobar-usds (approved):
🍏

client/src/components/J40footer.tsx
```diff
@@ -1,17 +1,56 @@
...
+        size="big"
+        primary={<></>}
+        secondary={<FooterNav
+          aria-label="Footer navigation"
+          size="big"   // fyi you leave this off and it silently fails... lovely
+          className={"nobreak"}
+          links={[
+            [
+              "Agency Partners",
+              <a href="https://www.epa.gov/">Environmental Protection Agency</a>,
```
NatHillardUSDS, line 19: Another small point actually is link underlines - I just left a comment about this in the spec as well, but by default links have underlines and the spec links currently do not. 

Again I think the right thing is to leave as-is here, get default link styling behavior, and change our spec, but just wanted to mention this

client/src/styles/global.scss
```diff
@@ -1,2 +1,31 @@
...
 @import "~@trussworks/react-uswds/lib/index.css";
+@import "~@trussworks/react-uswds/lib/uswds.css";
+
+.j40FloatLogoLeft {
+  width: 100%;
+  padding-left: 50%;
+}
+
+.j40Footer {
+  font-family: "sans-serif";
```
NatHillardUSDS, line 10: I wonder if we can keep this as default and set this at the [theme](https://designsystem.digital.gov/documentation/developers/#sass-and-theme-settings) level if need-be?

client/src/styles/global.scss
```diff
@@ -1,2 +1,31 @@
...
+  font-size: 12px;
+  line-height: 24px;
```
NatHillardUSDS, line 12: We should avoid manual font-sizes and use [font size classes](https://designsystem.digital.gov/utilities/font-size-and-family/) instead - these will work in accordance with the overall site relative font expectations

client/src/styles/global.scss
```diff
@@ -1,2 +1,31 @@
...
+}
+
+.j40FooterAddressTitle {
+  line-height: 2.5;
```
NatHillardUSDS, line 16: Likewise we should assume default line-height I think if possible, as this will vary based on font, language script, and device font settings

client/src/styles/global.scss
```diff
@@ -1,2 +1,31 @@
...
+  font-weight: bold;
```
NatHillardUSDS, line 17: I think we should be able to get this with the right header class

client/src/styles/global.scss
```diff
@@ -1,2 +1,31 @@
...
+  white-space: nowrap;
```
NatHillardUSDS, line 18: Can you explain why this change was necessary? It's possible this is workaround-able with classes also

client/src/styles/global.scss
```diff
@@ -1,2 +1,31 @@
...
+}
+
+.j40FooterLogo {
+  max-width: 120px;
```
NatHillardUSDS, line 22: I think a [width class](https://designsystem.digital.gov/utilities/height-and-width/) might be more resilient to resizing

client/src/styles/global.scss
```diff
@@ -1,2 +1,31 @@
...
+}
+
+// no way to control style of line address line items via react+trussworks so we're
+// overriding the style used.
+.usa-footer__contact-info {
```
NatHillardUSDS, line 27: Is it possible to start with the default options here and check with design?

NatHillardUSDS (changes requested):
I'd say the page structure looks good, but where possible we should be using size classes and out-of-the-box styling classes in order to be ready for responsive text, relative sizing, and page resizes down the line. In practice this means removing almost all manual sizing and replacing with classes wherever possible

client/src/components/J40footer.tsx
```diff
@@ -1,17 +1,56 @@
 import React from 'react';
-import { Footer } from '@trussworks/react-uswds';
-import {Link} from 'gatsby-plugin-intl';
-
-const footerLinks = [
-    <Link to="/">Home</Link>
-];
+import {Footer, Logo, FooterNav, Address} from '@trussworks/react-uswds';
+// @ts-ignore
+import whitehouseIcon from '../images/eop-seal.svg';
```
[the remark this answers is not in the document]
TomNUSDS: Agreed. It's pulled from an archived version of the whitehouse.gov site.

client/src/styles/global.scss
TomNUSDS: Good call. I was experimenting with setting the font and this should go away.

client/src/styles/global.scss
TomNUSDS: The default size is 1.0 line height which isn't 508 compliant. We should investigate changing it globally?

client/src/styles/global.scss
TomNUSDS: Sure. All the style tweaks were mostly to see how much work it was to match the design as closely as possible with minimal css adjustments.

client/src/styles/global.scss
NatHillardUSDS: Apparently `.usa-footer__logo-img` has `max-width: 5rem;` already actually [has](https://github.com/uswds/uswds/blob/develop/src/stylesheets/components/_footer.scss) `max-width: units(10);`

client/src/styles/global.scss
```diff
@@ -1,2 +1,24 @@
-@import "~@trussworks/react-uswds/lib/uswds.css";
 @import "~@trussworks/react-uswds/lib/index.css";
```
NatHillardUSDS, line 1: Ah, actually just noticed this, but this should be in a css module scoped to the header - you can copy what `layout.module.scss` has done here

client/src/styles/global.scss
NatHillardUSDS: Yeah I think for now we can start with what comes out of the box, then tweak post-convo with design ?

client/src/styles/global.scss
```diff
@@ -1,2 +1,24 @@
...
+@import "~@trussworks/react-uswds/lib/uswds.css";
+
+.j40FloatLogoLeft {
```
NatHillardUSDS, line 4: can we make this semantic? If we decide later to switch sides we'd need to update the class name

NatHillardUSDS (changes requested):
A few small points! overall removed code is good code!

client/src/pages/index.tsx
```diff
@@ -1,12 +1,180 @@
 import * as React from "react"
 import Layout from '../components/layout';
+// import {GridContainer, Grid} from '@trussworks/react-uswds';
+
+// this section seems verbose and silly
+// this inlines the svg as data:image/svg+xml ... this isnt' ideal.
```
NatHillardUSDS, line 6: Can you explain this a bit? i'm not familiar with the tradeoffs!

client/src/pages/index.tsx
```diff
@@ -1,12 +1,180 @@
...
+                <br/>
+                Send Feedback
+              </h3>
+              <p>Have ideas about how to
+                acknowledge the on-the-ground
+                experiences of your community?
+              </p>
+              Email: <a href="mailto:{{ justice40open@usds.gov }}">justice40open@usds.gov</a>
+
+              <p>&nbsp;</p>
```
NatHillardUSDS, line 161: Is there a way to accomplish this with styling / remove the nbsp?

client/src/styles/global.scss
NatHillardUSDS: we discussed offline - will do this once we have a fix for gatsby-plugin-scss-typescript we can use modules directly, will keep in global for now

NatHillardUSDS (approved):
There's a lot here, this is great! Kudos for wrestling USWDS like this.
LGTM for now. We talked about a few incremental improvements:
1.  localization - possible via  `intl.formatMessage`, but for longer blocks, we may want cms, which leads to
2. A cms / markdown for rendering content?
3. Breaking this out incrementally into smaller components
4. Breaking styles into separate css modules 
but we can handle these separately provided there's some place for them in a ticket

client/src/pages/index.tsx
TomNUSDS: updated comment

TomNUSDS:
Had merge-with-main issues, so I started a new branch and copied over my changes.
See https://github.com/usds/justice40-tool/compare/tomn-usds/template3

client/src/styles/global.scss
TomNUSDS: removed

PR #75, opened 2021-05-24
As an open source community member, I want to know what Justice40-specific terms and acronyms mean.

switzersc-usds:
**Description**
We had a glossary in our open source kickoff and someone suggested putting it in our Github repo so that it lives on and is discoverable. This PR adds the initial version of the glossary including ways to update or submit new terms to the glossary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@ Contributions are always welcome!
 
 See `CONTRIBUTING.md` for ways to get started.
 
+## Glossary
+
+Confused about a term? Heard an acronym and have no idea what it stands for? Check out [our glossary](docs/glossary.md)!
+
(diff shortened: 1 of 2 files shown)


NatHillardUSDS (approved):
Woo! Looks good to me. I could see this expanding over time to have different sections? Maybe a tech, a government, and a product component? In particular I think the acronym that initially tripped folks up was "EVCM", but this feels more like a product term and felt a bit odd to add alongside these others. Great this is set up in general as a living doc, we can add to it over time

switzersc-usds:
Oh, good points! I will add EVCM to this too.

NatHillardUSDS (approved):

PR #81, opened 2021-05-28
Fix misc stuff from lighthouse

TomNUSDS:
- Icons in list were resizing to be too small.
- Added robots.txt (full exclude for now)
- Add favicon.ico placeholder
- Added page title using helmet (react-helmet + gatsby-plugin-react-helmet)
Note: not sure if react-helmet can be in dev section of package.json?


client/static/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /
```
NatHillardUSDS, line 2: Makes sense for now; we'll need to readdress closer to launch -- might be worth a ticket?

client/package.json
```diff
@@ -50,9 +51,11 @@
   },
   "dependencies": {
     "@trussworks/react-uswds": "github:nathillardusds/react-uswds#nathillardusds/ssr",
+    "@types/react-helmet": "^6.1.1",
     "gatsby": "^3.4.1",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
+    "react-helmet": "^6.1.0",
```
NatHillardUSDS, line 58: as part of the security audit PR we're going to move these around; we can leave as-is for now

NatHillardUSDS (approved):
LGTM!

client/static/robots.txt
TomNUSDS: https://github.com/usds/justice40-tool/issues/84

client/package.json
TomNUSDS: Yeah, wasn't 100% sure where it needed to be. I'll move it into dev section and test.

Issue #11, opened 2021-04-26
As a repo contributor, I want to make sure my frontend code passes checks before/after it is merged, so that we maintain a high standard of quality and consistency


NatHillardUSDS:
**Description**
We should implement the best practices of continuous integration of source code, meaning performing a series of checks for PRs

**Solution**
Set up minimal pipelines with a chosen CI tool such as Github Actions/Jenkins/TravisCI that implement the above checks

**Describe alternatives you've considered**
The alternative is to not perform these checks, which can lead to ruin.

**Links to user research or other resources**
N/A

**Tech Tasks**
- [x] Linting/formatting (eg eslnt, prettier)
- [x] Security auditing (eg `npm audit`)
- [x] License check (eg `license-checker`)
- [x] Dependency updating (eg dependabot etc)
- [x] Container-based unit testing / reporting (eg jest, junit)
- [x] Integration testing (e.g. puppeteer)
- [x] Snapshot testing (jest-snapshot, etc)

**Definition of "Done"**
- [ ] Checks run on every PR
- [ ] We have identified a basic set of tests that run on every PR / Commit and these are running regularly and producing results

switzersc-usds:
Status update: In progress, some of these may be higher priority than others.

PR #85, opened 2021-05-28
Adding Cypress for e2e testing

NatHillardUSDS:
Addresses issue #11 item "Integration testing", using Cypress as described [here](https://www.gatsbyjs.com/docs/how-to/testing/end-to-end-testing/). 

Adds two tests out of the gate:
* Basic accessibility (catches the issues Tom identified earlier!), and
* Language switching

Also creates a daily 4am e2e test run


TomNUSDS (approved):
Looks Great To Me

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,13 @@
+name: example-cron
+on:
+  schedule:
+    # runs tests every day at 4am
+    - cron: '0 4 * * *'
```
TomNUSDS, line 5: Is there a timezone? Is it 4am UTC?

client/gatsby-config.js
```diff
@@ -53,5 +53,6 @@ module.exports = {
         },
       },
     },
+    'gatsby-plugin-react-helmet',
```
TomNUSDS, line 56: Oh good catch.

client/src/styles/global.scss
```diff
@@ -17,4 +17,11 @@
 
 .j40-two-column-icons-spacing {
   padding-right: 1em;
+  width: 2.3em;
```
TomNUSDS, line 20: I will need to merge with my pending PR. (That will teach me to sit on a merge) 😆

client/static/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
```
TomNUSDS, line 1: `robots.txt` and `favicon.ico` is already in `/static/`... is the plan to remove those in another PR?

TomNUSDS (approved):
Minor questions, but nothing blocking. Approved.

client/static/robots.txt
NatHillardUSDS: ah, this is actually because I had merged `main` into this and thus your most recent change came with

client/src/styles/global.scss
NatHillardUSDS: ah, this originally came from a merge with `main` actually so it should merge cleanly

.github/workflows/e2e.yml
NatHillardUSDS: Ah true, this is UTC - ends up being midnight; the comment came from a copied comment elsewhere,

client/gatsby-config.js
NatHillardUSDS: Yeah I believe this was an import error

TomNUSDS (approved):
Thanks for fixing up those minor issues. Approved.

PR #87, opened 2021-06-02
More tweaks from today's meeting

TomNUSDS:
Just playing with layout.
Fixed some minor items noted from today's meeting.


client/src/styles/global.scss
```diff
@@ -21,7 +21,17 @@
...
 }
 
+.j40-footer-logo {
+  display: inline-block;
+  text-align: center;
+}
+
+.j40-address-readability {
+  display: inline-block;
+  line-height: 1.5 !important;
```
NatHillardUSDS, line 31: Generally I'm iffy about `!important` - I'd wonder if there were another way to achieve this?

NatHillardUSDS (changes requested):
Great you could get these in! A small comment, otherwise looks good  - Looks like you need to update the footer snapshot though for tests to pass

client/src/styles/global.scss
TomNUSDS: Agreed that it's iffy. This is a Trussworks issue IMO. There's no easy way to control the layout of the `<Address>` and the lines are not spaced at the readability advised 1.5lines. If I replace the react component with standard html, I could remove the `!important`.

TomNUSDS: Should I make issues on the trusswork fork? 

While it's important, I don't have a ton of time to invest in driving improvements in their library.

NatHillardUSDS: We can leave as `important` for now - it might be worth bringing up in a design/tech sync that if we did the address their way it would be a bit easier for us. We should assume no changes to our trussworks fork -- meanwhile I'll file the bug I found against them directly and see about incorporating directly so we can move off of it

NatHillardUSDS (approved):
LGTM!

NatHillardUSDS (approved):
LGTM! Next step might be abstracting the `noreferrer` bit into a component but this can come later

PR #94, opened 2021-06-04
More layout fixes

TomNUSDS:
- Get aside flush with top/bottom
- Media change iconlist to one column on mobile
- Misc other fixes
TODO: body font need to be addressed at the theme level?

### DIFF
--- a/client/src/pages/index.tsx
+++ b/client/src/pages/index.tsx
@@ -38,13 +38,13 @@ const IndexPage = () => {
   return (
     <Layout>
       <main id="main-content" role="main">
-        <div className="grid-row grid-gap-2">
-          <section className="grid-container usa-section">
-            <div className="desktop:grid-col">
+        <div className={'grid-row grid-gap-2'}>
+          <section className={'grid-container usa-section'}>
+            <div className={'desktop:grid-col'}>
 
-              <div className="grid-row grid-gap">
+              <div className={'grid-row grid-gap'}>
                 <div
-                  className="usa-prose text-asset-container">
+                  className={'usa-prose text-asset-container'}>
                   <h2>About Justice40</h2>
                   <p>
                     In an effort to address historical environmental injustices,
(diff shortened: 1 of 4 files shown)


client/src/pages/index.tsx
```diff
@@ -153,32 +151,35 @@ const IndexPage = () => {
...
+
+          {/* aside */}
+          <aside
+            aria-labelledby="left-sidebar"
+            className={'j40-aside desktop:grid-col-3'}
+            id="left-sidebar">
+            <section
+              className={'usa-prose grid-gap grid-container usa-section'}>
+              <h3 className={'j40-aside-title'}>Get Involved</h3>
+              <h5>
```
NatHillardUSDS, line 164: Based on how [this page](https://designsystem.digital.gov/components/prose/) uses this, I'm thinking the analog is to make `Get Involved` an `<h1>` and then `Send Feedback` an `<h2>`?

These may not match the style spec but they seem like semantically the correct components to use -- if we're using them to get the right font-size it seems we should actually change the spec instead?

NatHillardUSDS (changes requested):
Overall looks good! Left a comment about semantics of heading items based on the `usa-prose` example though - overall I think if we get the semantics right the styles should follow , if we're matching examples

client/src/pages/index.tsx
TomNUSDS: Yeah, I was visually just trying to match the prototype sizes and style.

Think we need to fix the page font to be `Source Sans Pro` instead of the current `Merriweather` and then maybe the `<h#>` tags will be closer? (An easy way to spot the font difference is by looking at the lower-case `i`).  

All the headers tags on https://designsystem.digital.gov/components/prose/ are using `Merriweather` (as is our page right now), but mocks are using `Source Sans Pro` for the header tags.

NatHillardUSDS (approved):
OK as an incremental update - I would say the current heading values are a bit of a hack; we should revisit after changing font (maybe capture this in a ticket so we don't lose it?)

TomNUSDS:
See #95

PR #105, opened 2021-06-05
Add timeline page

TomNUSDS:
- Add timeline page
- Factor out common aside into component.
TODO: Get nav menu working.

### DIFF
--- a/client/src/styles/global.scss
+++ b/client/src/styles/global.scss
@@ -39,7 +39,6 @@
 
 .j40-title {
   font-size: xx-large;
-  font-family: "Merriweather Web";
 }
 
 .j40-aside {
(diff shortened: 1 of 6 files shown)


client/src/styles/global.scss
```diff
@@ -60,6 +59,17 @@
...
+
+// This should really be part of uswds and use $theme-step-indicator-segment-color-complete
+.j40-usa-process-list__item--complete {
+  &::before {
+    color: white;
+    background-color: #00a91c;
+    content: '✓'
+  }
+
+  border-left-color: #005ea2 !important;  // todo: fix
```
NatHillardUSDS, line 74: Can you make a ticket for this so we can track, and ref the ticket in the comment?

client/src/styles/global.scss
```diff
@@ -60,6 +59,17 @@
 
 .j40-address-readability {
   display: inline-block;
-  line-height: 1.5 !important;
+  line-height: 1.5 !important;  // trussworks issue
```
NatHillardUSDS, line 62: Can you elaborate a bit in the comment for context?

NatHillardUSDS (approved):
Cool! awesome to have this - approved with some comment comments

client/src/styles/global.scss
TomNUSDS: This is required because the Progress element doesn't really support color changing the line connecting progress (like the Steps element does). So, I it took a bigger hammer to make it work.

PR #117, opened 2021-06-09
Adding Simple URL-based feature flags

### LINKED ISSUE #66: As a developer, I want to limit the audience that sees new features, so that we can control the message and positioning of our tool
**Description**
We should add some means of limiting the audience for new features, so as to control the visibility of in-progress work while also getting real users to use our software

**Solution**
Use feature flagging on in-progress work, control through an allow list

* LaunchDarkly - a very full-featured version of this from the private sector; is there an OS equivalent?
* Security through obscurity - only sharing the link to certain people is not wise, as the link will get out. 

**Tasks**
- [ ] Add a feature flagging mechanism
- [ ] Use this feature flag for a particular feature
- [ ] Setup an initial audience for the feature

**Definition of "Done"**
- [ ] We have released a feature to only a subset of our total audience

**Open Questions**
* What tool should we use, in keeping with our OSS bent?
* How should we control the allow list, given we don't have user logins?

NatHillardUSDS:
Fixes #66: As a developer, I want to limit the audience that sees new features, so that we can control the message and positioning of our tool.

Implements simple feature flagging via URL parameters. Provide "?flags=x,y,z" to enable flags x, y, and z.

Also adds simple flagging for the "Timeline" feature: http://localhost:8000/?flags=timeline now shows the timeline.

### DIFF
--- a/client/src/styles/global.scss
+++ b/client/src/styles/global.scss
@@ -59,7 +59,7 @@
 
 .j40-address-readability {
   display: inline-block;
-  line-height: 1.5 !important;  // trussworks issue
+  line-height: 1.5 !important; // trussworks issue
   text-align: center;
 }
 
@@ -68,8 +68,8 @@
   &::before {
     color: white;
     background-color: #00a91c;
-    content: '✓'
+    content: "✓";
   }
 
-  border-left-color: #005ea2 !important;  // todo: fix
+  border-left-color: #005ea2 !important; // todo: fix
 }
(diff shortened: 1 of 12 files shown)


TomNUSDS (dismissed):
lgtm;  I'm going to have to hand merge in some of these changes with my current working branch, so when you get that PR, we'll need to double-check I didn't drop/mess anything up.

client/src/styles/global.scss
```diff
@@ -68,8 +68,8 @@
   &::before {
     color: white;
     background-color: #00a91c;
-    content: '✓'
+    content: "✓";
```
TomNUSDS, line 71: Think this is an indication that prettier isn't running correctly for my ide setup?

NatHillardUSDS: Ah yeah, that appears to have snuck in - this was formatted automatically when I ran `gatsby develop`

TomNUSDS (approved):

PR #118, opened 2021-06-10
Tomn usds/template8

TomNUSDS:
* Added initial navbar
* Pages are the about/home and the timeline.
* Made layouts consistent across browsers: Chrome, Firebox, Safari
* Simplified html, used more trussworks.
* Attempted to control more with uswds themes, but not 100% successful (navbar menu mostly). Now that it's part of the build, the build is slower. Not really sure it's worth it since what we want to theme is an inverted navbar, but we can't fully do that. 

Todo: 
* Subtle things with inverted navbar UI aren't actually inverted. No easy control to customize with trussworks.
* Kind of used a hammer in globals.scss (aka `important!`)
* Will probably undo the uswds scss "compiling" and just use the uswds css with style overrides.


TomNUSDS:
Converted to draft pending initial review and discussion with Nat.

client/src/components/J40Aside.tsx
```diff
@@ -4,50 +4,46 @@ import * as React from 'react';
...
-              target={'_blank'}
-              rel={'noreferrer'}
-              key={'github3'}>Check it out on GitHub</a>
-          </p>
-        </section>
+        aria-labelledby="right-sidebar"
+        className={'j40-aside desktop:grid-col-3 usa-prose usa-section'}
+        id="right-sidebar">
+        <h1>Get Involved</h1>
+        <h2><img
```
NatHillardUSDS, line 17: 👍 !

client/src/components/J40Header2.tsx
```diff
@@ -0,0 +1,102 @@
...
+            className={'j40-header'}
+          >
+          </PrimaryNav>
+        </div>
+      </Header>
+    </>
+  );
+};
+
+export default J40Header2;
```
NatHillardUSDS, line 102: Is this one you'd want to A/B test? Can you write a comment explaining the differences ?

client/src/components/layout.tsx
```diff
@@ -1,7 +1,9 @@
 import React, {ReactNode} from 'react';
-import * as styles from './layout.module.scss';
-import J40Header from './J40Header';
+import {GridContainer, Grid} from '@trussworks/react-uswds';
+// import * as styles from './layout.module.scss';
```
NatHillardUSDS, line 3: Can we remove comments?

client/src/components/J40Header2.tsx
NatHillardUSDS: You can also add a flag for this using our new flagging mechanism!:
```js
const FlaggedHeader = () => {
  const flags = useFlags();
  return (
    <>
      {flags.includes('header2') ? <J40Header2/> : <J40Header/>}
    </>
  );
};
```

Then later:
```jsx
<URLFlagProvider location={location}>
      <FlaggedHeader />
...
```

NatHillardUSDS (changes requested):
Great changes!
I would see if we need to re-compile the scss and/or if there is another way; also provide a comment maybe about the differences with header2

client/src/components/J40Header2.tsx
TomNUSDS: Oh, Header2 was an attempt as a "placemarker" while I did radical changes. I should roll these into Header.tsx.

client/package.json
```diff
@@ -25,7 +25,8 @@
     "lint": "eslint './src/**/*.{ts,tsx}' --ignore-pattern node_modules/ --ignore-pattern public --ignore-pattern *scss.d.ts",
     "lint:fix": "npm run lint -- --quiet --fix",
     "intl:extract": "formatjs extract 'src/(pages|components)/*.tsx' --out-file src/intl/en.json",
-    "intl:compile": "formatjs compile src/intl/en.json --ast --out-file compiled-lang/en.json"
+    "intl:compile": "formatjs compile src/intl/en.json --ast --out-file compiled-lang/en.json",
+    "compile-scss": "node-sass --include-path ./src/scss --include-path ./node_modules/uswds/dist/scss --output-style compressed -o ./src/styles/ ./src/scss/App.scss"
```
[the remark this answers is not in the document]
TomNUSDS: You are correct. I was using this check just the scss compile while I was working on getting it integrated. I removed it.

NatHillardUSDS:
LGTM!

NatHillardUSDS (approved):

PR #127, opened 2021-06-14
Update global.scss

TomNUSDS:
Fix some low-contrast accessibility issues for mobile menu.


NatHillardUSDS (approved):
LGTM - would overall wonder whether we could automate / may already be automating contrast checking via cypress - can we create a cypress test to check this?

TomNUSDS:
Oh, good thinking. Created ticket #128

Issue #128, opened 2021-06-14
Use Cypress to test accessibility contrast checks


TomNUSDS:
https://github.com/usds/justice40-tool/pull/127#pullrequestreview-683314044

Would be nice to have Cypress to automatically tests check for low-contrast (accessibility) issues.

PR #129, opened 2021-06-14
Make the accessibility checkers happy

TomNUSDS:
"Fixes" #128 - cypress accessibility check was already there, this change just gets all the tests to pass.
Checkers (cypress and lighthouse) are at 100% for accessibility.

* Include a lang specification in the `<html>` tag. This is based on the `useIntl()` value.
* reorder `<h#>` header tags.
The `<Aside` to from `<h1>`, `<h2>` to be `<h2>`, `<h3>`. This is because trusswork's `<Footer>` uses and `<h4>` and the accessibility checkers complain it was skipping from `<h2>` -> `<h4>`
* Change `<aria-details>` to `<aria-label>` for milestone list.


NatHillardUSDS (approved):
Nice, lgtm! great to get these fixed

Issue #136, opened 2021-06-15
As a developer, I want to have a clear spec of the comparison tool, so that I can begin my work.


lucasmbrown-usds:
**Description**
We need a clear spec for the comparison tool. Probably using Excel or a Python notebook.

**Solution**
There's a spec for the comparison tool that unblocks work on https://github.com/usds/justice40-tool/issues/135. 

**Tasks**
Write the spec. 

**Definition of "Done"**
* The developer working on https://github.com/usds/justice40-tool/issues/135 feels like they have enough information to begin their work.

lucasmbrown-usds:
Spec completed in python notebook at https://github.com/usds/justice40-tool/pull/140!

Issue #135, opened 2021-06-15
(Comparison tool 0.1) As a stakeholder interested in the CEJST score, I want to see a report comparing the CEJST score to one other index, so that I can understand where the CEJST score generates results that are very different than another index.


lucasmbrown-usds:
**Description**

This is essentially the "MVP" of the comparison tool. We want to see some basic analytics comparing one version of the CEJST score to one known index. 

In a future ticket, we will add support for multiple draft versions of the CEJST score for comparison.  We will also add support for more comparison indices, and more detailed analytics on the comparison. These other indices have been peer-reviewed and validated  through years of academic research; comparing different versions of the CEJST score to them provides a quick way to flag potential outliers in the CEJST score where it may need to be tweaked to provide a more reliable designation of overburdened and underserved communities.

Note that most of these indices are created at a higher geographic unit than census block group, such as census tract. This report will require a methodology for comparing indices at the unit of census tract (or county, etc) against a CEJST score at the unit of census block group.

**Solution**
A Python script or notebook that loads (1) a comparison index and (2) a current CEJST score and produces a report based on the differences between the two scores. 

**Describe alternatives you've considered**
Not having a quantitative comparison method and relying on qualitative feedback about whether the score looks representative of the most overburdened and underserved communities. Qualitative feedback is more piecemeal and moves slower, but is an important piece of feedback in combination with this work. 

**Tasks**
* Identify the top index as prioritized by https://github.com/usds/justice40-tool/issues/133. 
* Load that index.
* Load the initial CEJST score as created by https://github.com/usds/justice40-tool/issues/48
* Calculate comparison metrics between the scores
* Produce a report (such as a CSV, or a map with differences visualized, etc) that summarizes the difference between these scores

**Definition of "Done"**
* Report is ready for review with internal stakeholders

switzersc-usds:
Waiting on scoring code.

PR #140, opened 2021-06-16
Score comparison tool, first draft 

lucasmbrown-usds:
I was starting to write up a specification for how to compare different scoring techniques, and found it harder to explain in words and certainly harder to calculate in Excel than it would be in Python. 

~~This notebook provides an example of an early version of the comparison tool. Therefore, it's closest to matching the ticket #136. Once updated to use real data, it would probably meet the definition of done for #135.~~

This completes #136 and #135. 

~~It uses EJSCREEN's % low-income (<=200% of the federal poverty line) percentile as if that is the score for the Census Block Groups. This is obviously a dummy score that needs to be replaced with a draft, placeholder score.~~

This now uses the score generated by `score_calc_0.1.ipynb`.

For comparison, it uses CalEnviroScreen 4.0 data. 

Because CalEnviroScreen is calculated at the census tract, which is larger than the census block group, this comparison accounts for that using joins and some statistics. 

I also setup a Jupyter extension that allows you to render variable values in Markdown, so that there's a block of text that easily updates based on the calculated metrics. 

Here's the current version of that text: 

Screen Shot 2021-06-15 at 11 28 31 PM

Once we have a placeholder score, it should be easy to update this to use those values.


docs/decisions/0002-files/Highways.csv
```diff
@@ -1,5 +1,5 @@
-,OpenLayers,OpenLayers+MB,OpenLayers+React,Leaflet,Mapbox,Mapbox React
```
lucasmbrown-usds, line 1: My local git converted the line-endings on all of these; I think perhaps because they were otherwise Windows-based line endings? I'm not sure. This issue seems related based on the console messages I was getting. https://stackoverflow.com/questions/5834014/lf-will-be-replaced-by-crlf-in-git-what-is-that-and-is-it-important

esfoobar-usds: Here's[ a good solution for this issue](https://stackoverflow.com/a/5834094).

score/ipython/score_calc_0.1.ipynb
```diff
@@ -70,7 +72,7 @@
    "outputs": [],
    "source": [
     "# strip calculations\n",
-    "df = df[[\"ID\", \"score_a_percentile\", \"score_b_percentile\",\"score_a_top_percentile_25\",\"score_b_top_percentile_25\"]]"
+    "df = df[[\"ID\", \"ACSTOTPOP\", \"score_a\",\"score_b\", \"score_a_percentile\", \"score_b_percentile\",\"score_a_top_percentile_25\",\"score_b_top_percentile_25\"]]"
```
ginabeena, line 75: what does this value mean: `ACSTOTPOP`?

ginabeena (dismissed):
LGTM :shipit:

score/ipython/scoring_comparison.ipynb
```diff
@@ -0,0 +1,289 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "29c14b29",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Load comparison index (CalEnviroScreen 4)\n",
+    "# TODO: Instead of loading a manually-created CSV from the repo, actually download the data and convert it using \n",
```
esfoobar-usds, line 97: Cell 3 of the [ejscreen_etl notebook](https://github.com/usds/justice40-tool/blob/main/score/ipython/ejscreen_etl.ipynb) has the download functionality.

esfoobar-usds (approved):
An small comment about temporary data sources or outputs:

For transient CSVs or zip files that we get from the web, we should use the `/data/tmp` folder and at the end of the notebook delete its contents.

For example this part of the notebook writes on the main data folder:
```
# Write comparison to CSV.
comparison_df.to_csv(
    path_or_buf=DATA_DIR / "Comparison Output.csv", 
    na_rep="", 
    index=False
)
```
This should be written in `/data/tmp/` instead -- generate the markdown and then delete it.

Otherwise amazing Pandas-fu in display!!!

score/ipython/score_calc_0.1.ipynb
lucasmbrown-usds: census block group total population - we should do a quick style guide around naming practices for variables to all get on the same page. (E.g., I've often worked on projects where we renamed all input data variables into our more readable style patterns.)

Creating this now! https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/189

score/ipython/scoring_comparison.ipynb
lucasmbrown-usds: Good point! I'm gonna leave this for now since I ran into some bot protection services on their site 🤦 : see https://github.com/usds/justice40-tool/issues/194

esfoobar-usds (changes requested):
When I tried running cell 3 of the comparison notebook, I got the following error:

```
KeyError: "['ACSTOTPOP', 'score_a_percentile'] not found in axis"
```

The other change I suggest is adding instructions to download the `CalEnviroScreen_4.0_2021.csv` file into the `data/tmp` folder. I don't really like to add it to the Github repo as it is now.

Let me your thoughts.

esfoobar-usds:
@lucasmbrown-usds I see what the issue is -- there's only 3 columns on the `data/score/csv/usa.csv` file:

```
ID,score_a,score_b
010010201001,0.7091538716397453,0.49581991740654485
```

And you are trying to rename a field that doesn't exist:
```
# Rename unclear name "id" to "census_block_group_id", as well as other renamings.
cejst_df.rename(
    columns={
        "ID": CENSUS_BLOCK_GROUP_ID_FIELD,
        "ACSTOTPOP": CENSUS_BLOCK_GROUP_POPULATION_FIELD,
        "score_a": CEJST_SCORE_FIELD,
        "score_a_percentile": CEJST_PERCENTILE_FIELD,
    },
    inplace=True,
    errors="raise",
```

lucasmbrown-usds:
> @lucasmbrown-usds I see what the issue is -- there's only 3 columns on the `data/score/csv/usa.csv` file:

Ah. @esfoobar-usds try running the `score_calc_0.1.ipynb` one more time -- I updated it to add some new fields, and I'm guessing you have not run it using the new code?

esfoobar-usds:
Pero convo with @lucasmbrown-usds, I will upload the CalEnviroScreen.csv file to our S3 bucket and rewrite that cell.

score/README.md
```diff
@@ -18,6 +18,16 @@
...
+## Activating variable-enabled Markdown for Jupyter notebooks
+- Change to this directory (i.e. `cd score`)
+- Run `jupyter contrib nbextension install --user`
+- Run `jupyter nbextension enable python-markdown/main`
+- Make sure you've loaded the Jupyter notebook in a "Trusted" state. (See button near
+ top right of Notebook screen.)
+
+For more information, see [nbextensions docs](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/install.html) and
+see [python-markdown docs](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/python-markdown).
+
```
esfoobar-usds, line 30: I would move this section to the score comparison notebook itself, since this pertains specifically to generate the markdown cell, which only pertains to people who runs this, which I believe are going to be a small amount of people. @lucasmbrown-usds thoughts?

lucasmbrown-usds: Hmm. I mean, we also have the jupyter notebook setup instructions in that same place, so it feels like grouping like with like, and I generally prefer to have setup instructions in just one place. 

I could add a sentence that says "currently only used by the Notebook `scoring_comparison.ipynb`?

esfoobar-usds: Yeah, that works.

esfoobar-usds:
@lucasmbrown-usds updated the zip file location and flow, please review and if it all looks well we can merge

Issue #100, opened 2021-06-04
As a developer, I want to visualize Census TIGER vector data in my national map from my own pipeline. 


switzersc-usds:
**Description**
This is primarily centered around getting this base data into geoplatform in a usable way, not user experience for the map.

**Solution**
Use the geoplatform.gov platform for processing this straightforward dataset for our base of geographic boundaries in the national map. 

Note: We want to be able to run the data pipeline end to end locally / without hosting dependencies. We should talk to geoplatform about process for adding the TIGER data and how we might be able to host or store in version control the processing code. 

Lots of dependencies on geoplatform and unknownss.

**Describe alternatives you've considered**
We could do this without geoplatform processing but we need to move forward on setting up the infrastructure. 

**Links to user research or other resources**

**Tasks**
- [x] Geoplatform.gov will get back to us by COB 6/4 with timeline for processing the TIGER data.
- [x] Talk to geoplatform about making this code available open source and available in a docker container
- [ ] Ensure the data is available on the geoplatform.gov pipeline and tile server in the format we need
- [ ] Point the map in the tool to visualize this data. (In meantime, we can use census tile server until we set up our pipeline)

**Definition of "Done"**
- [ ] Visit the tool map page and see the census block group borders rendered on the map above zoom level 9 (locally or production)

esfoobar-usds:
https://www2.census.gov/geo/tiger/TIGER2010/BG/2010/?C=S;O=D is where you can get 2010 basic blockgroups.

`tl_2010_01_bg10.zip` would represent a state.

`tl_2010_XX_bg10.zip` `01` to `56` and `72` if you want Puerto Rico.

https://www.census.gov/geographies/reference-files/time-series/geo/tallies.html has the state tables for reference.

switzersc-usds:
Draft PR in progress: https://github.com/usds/justice40-tool/pull/122

switzersc-usds:
@esfoobar-usds Note that starting with Minnesota would be best if possible

esfoobar-usds:
This task is awaiting @NatHillardUSDS's confirmation that he can read the mvt folder on s3, once it's completely uploaded, otherwise it's complete and can see the CBGs on my local machine.

PR #148, opened 2021-06-17
Adding ol-mapbox-style for easier styling

NatHillardUSDS:
For #100 - adding `ol-mapbox-style` to make styling a bit more standardized/straightforward. Note that this modifies our `tsconfig` to support JSDoc, which both `ol` and `ol-mapbox-styles` uses


TomNUSDS (approved):
Nice simplification. LGTM

PR #149, opened 2021-06-17
Minimal refactor to make maps wider

TomNUSDS:
Not particularly pretty, but it works.


NatHillardUSDS (approved):
This works for now, thanks for tackling!

Issue #210, opened 2021-06-23
Aside returns when you click links


NatHillardUSDS:
**Describe the bug**
After the fix for #149 , the map page is full-width, but only until they reload or navigate directly to the page. This is because it brings back a trailing slash which breaks the current page check. We should update the check to fix this.

**To Reproduce**
* Load the latest page link built from main: https://usds.github.io/justice40-tool/e7ccd3/en/cejst/

**Expected behavior**
No sidebar

**Screenshots**
![Screen Shot 2021-06-23 at 12.36.41 PM.png](https://images.zenhubusercontent.com/6091a54c180d4758f3eedc40/14cb8bac-4625-4820-8371-c6f2ea6a2b55)

PR #211, opened 2021-06-23
Addresses #210, aside should not appear on map page

NatHillardUSDS:
Small fix for #210 - work for trailing slash also


TomNUSDS (approved):
Yup. Good fix. Thanks

Issue #199, opened 2021-06-21
Visual fixes needed


switzersc-usds:
List of visual fixes as discussed in a previous dev+design sync. See sheet for more details, but I've captured the short summaries as a checklist here:

- [ ] ~Remove layer toggle~
- [x] Update content in warning
- [x] Update Title
- [x] Add back warning icon in Alert
- [x] "Clickhere."
- [ ] ~Language Toggle~
- [x] Switch to light mode

NatHillardUSDS:
In the dev-design sync earlier, we also discussed:
- About page updates
  - Grid updated on About page
  - What happens when you go full width? For now, let it be vis-a-vis column width, it's a beta, though it's a bit funky -- readdress closer to ship 
  - Remove language selector for now

PR #212, opened 2021-06-23
Addresses #199 - removing non-functioning layer switcher for now

NatHillardUSDS:
Previously part of demo branch, now breaking out as part of its own fix for clarity. There's definitely more to do for #199, but needed this for map display for now

### DIFF
--- a/client/src/pages/cejst.tsx
+++ b/client/src/pages/cejst.tsx
@@ -1,7 +1,6 @@
 import React, {useState} from 'react';
 import Layout from '../components/layout';
 import MapWrapper from '../components/map';
-import MapControls from '../components/mapControls';
 import HowYouCanHelp from '../components/HowYouCanHelp';
 import Feature from 'ol/Feature';
 import Geometry from 'ol/geom/Geometry';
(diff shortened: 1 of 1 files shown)


client/src/pages/cejst.tsx
```diff
@@ -14,7 +13,7 @@ interface IMapPageProps {
 }
 
 const CEJSTPage = ({location}: IMapPageProps) => {
-  const [features, setFeatures] = useState<Feature<Geometry>[]>([]);
+  const [features] = useState<Feature<Geometry>[]>([]);
```
TomNUSDS, line 16: Humm.. `useState` without a setter? I think that just makes it a `const`?
Is this just a temporary thing so you can easily remove the `<MapControls>`

NatHillardUSDS: This is temporary for now, yes - we might need it later

TomNUSDS: OK. Maybe a comment explaining that and then the PR LGTM.

TomNUSDS (dismissed):
LGTM

client/src/pages/cejst.tsx
NatHillardUSDS: Good call - added comment just now

TomNUSDS (approved):
Reapproved

PR #213, opened 2021-06-23
Modularization + Poetry + Counties

esfoobar-usds:


score/application.py
```diff
@@ -0,0 +1,56 @@
...
+from etl.sources.census.etl_utils import reset_data_directories as census_reset
+from etl.sources.ejscreen.etl_utils import reset_data_directories as ejscreen_reset
+from utils import remove_files_from_dir
+from etl.sources.census.etl import download_census_csvs
+
+settings.APP_ROOT = Path.cwd()
+
+
+@click.group()
+def cli():
```
lucasmbrown-usds, line 15: Could we add docstrings to all our functions? It'll help out a lot in the long run!

score/application.py
```diff
@@ -0,0 +1,56 @@
+from config import settings
+import click
+from pathlib import Path
+import logging
+
+from etl.sources.census.etl_utils import reset_data_directories as census_reset
```
lucasmbrown-usds, line 6: We should have a conversation about this in the styleguide, I generally prefer using the full path to modules rather than importing them. 

E.g., `import etl` and then 

```
etl.sources.census.etl_utils.reset_data_directories(...)
```

It's more verbose, but more explicit and makes it more obvious to the reader where everything's coming from. Plus you don't have to do the renamings as you've done here. :)

score/etl/sources/census/etl.py
```diff
@@ -0,0 +1,115 @@
...
+import os
+import json
+from pathlib import Path
+import logging
+
+from .etl_utils import get_state_fips_codes
+from utils import unzip_file_from_url
+
+
+def download_census_csvs(data_path: Path) -> None:
```
lucasmbrown-usds, line 11: love the types!!

score/etl/sources/census/etl.py
```diff
@@ -0,0 +1,115 @@
...
+    # the fips_states_2010.csv is generated from data here
+    # https://www.census.gov/geographies/reference-files/time-series/geo/tallies.html
+    state_fips_codes = get_state_fips_codes(data_path)
+    for fips in state_fips_codes:
+        # check if file exists
+        shp_file_path = data_path.joinpath(
```
lucasmbrown-usds, line 17: Any reason we've switched to joinpath over just using slashes (e.g., `data_path / "census" / "shp" / ...`)? 

I don't have a strong preference, in all things I just prioritize consistency.

score/etl/sources/census/etl.py
```diff
@@ -0,0 +1,115 @@
...
+            "geojson",
+        )
+
+        if not os.path.isfile(geojson_dir_path.joinpath(fips + ".json")):
+            # ogr2ogr
+            logging.info(f"Encoding GeoJSON for {fips}")
+
+            # if it's running inside Docker container
+            if os.environ.get("DOCKER_CONTAINER"):
+                # TODO
```
lucasmbrown-usds, line 41: Could we expand on what this TODO is? Is it to figure out how to run the encoding from a docker container? 

Maybe we log an error or something if this is not implemented; otherwise I could see a future developer scratching their heads about why it's not working.

score/etl/sources/census/etl.py
```diff
@@ -0,0 +1,115 @@
...
+                    + "/tl_2010_"
+                    + fips
+                    + "_bg10.shp"
+                )
+                os.system(cmd)
+
+        # generate CBG CSV table for pandas
+        ## load in memory
+        cbg_national_list = []  # in-memory global list
+        cbg_per_state_list: dict = {}  # in-memory dict per state
```
lucasmbrown-usds, line 66: nit: why is this variable named `_list` if it's a `dict`?

score/etl/sources/census/etl.py
```diff
@@ -0,0 +1,115 @@
...
+                    geojson = json.load(f)
+                    for feature in geojson["features"]:
+                        geoid10 = feature["properties"]["GEOID10"]
+                        cbg_national_list.append(str(geoid10))
+                        geoid10_state_id = geoid10[:2]
+                        if not cbg_per_state_list.get(geoid10_state_id):
+                            cbg_per_state_list[geoid10_state_id] = []
+                        cbg_per_state_list[geoid10_state_id].append(geoid10)
+
+        csv_dir_path = data_path / "census" / "csv"
```
lucasmbrown-usds, line 80: nit: inconsistent with use of `joinpath` above.

score/etl/sources/census/etl.py
```diff
@@ -0,0 +1,115 @@
...
+        ## write to individual state csv
+        for state_id in cbg_per_state_list:
```
lucasmbrown-usds, line 82: I'm having a little trouble following - what is being written here? Is the only data in this file going to be the CBG GEOID? 

(Either way, do you mind adding some more comments explaining it? Thanks!)

score/requirements.txt
```diff
@@ -1,6 +1,76 @@
-ipython
-jupyter
-jupyter_contrib_nbextensions
-numpy
-pandas
-requests
+appnope==0.1.2; sys_platform == "darwin" and python_version >= "3.7" and platform_system == "Darwin"
```
lucasmbrown-usds, line 1: Hmm, I'm guessing there's actually all the "downstream" dependencies here. (E.g., this file could also just have 10 dependencies in it that would install their own pre-requisites.) It might be a little cleaner to only list the "upstream" dependencies. 

But! I don't know `Poetry` at all, so if this is standard and the easiest way to generate it, that's fine.

score/settings.toml
```diff
@@ -0,0 +1,8 @@
+[default]
```
lucasmbrown-usds, line 1: Nice, love this file.

score/tile/generate.py
```diff
@@ -0,0 +1,91 @@
...
+from pathlib import Path
+import shutil
+
+from etl.sources.census.etl_utils import get_state_fips_codes
+
+
+def generate_tiles(data_path: Path):
+
+    # remove existing mbtiles file
+    mb_tiles_path = data_path / "tiles" / "block2010.mbtiles"
```
lucasmbrown-usds, line 11: Nit: with all the ETL functions, we separated out the cleanup script from the generation script. Is it worth doing that here for consistency? We could always create a single function afterwards that calls both (e.g., `cleanup_and_generate_tiles`.

score/etl/sources/census/etl_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    # geojson
+    geojson_path = census_data_path / "geojson"
+    remove_files_from_dir(geojson_path, ".json")
+
+    # shp
+    shp_path = census_data_path / "shp"
+    remove_all_dirs_from_dir(shp_path)
+
+
+def get_state_fips_codes(data_path: Path) -> list:
```
lucasmbrown-usds, line 25: Hmm, every time we run this we're going to have to download a file from remote? I'm wondering if we just want to store the list as a list rather than invoking an outbound request every time.

score/tile/generate.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # join the state shape sqllite with the score csv
+    state_fips_codes = get_state_fips_codes()
+    for fips in state_fips_codes:
+        cmd = (
+            'docker run --rm -v "'
+            + pwd
+            + '"/:/home '
+            + "osgeo/gdal:alpine-small-latest ogr2ogr -f GeoJSON "
+            + f"-sql \"SELECT * FROM tl_2010_{fips}_bg10 LEFT JOIN '/home/data/score/csv/data{fips}.csv'.data{fips} ON tl_2010_{fips}_bg10.GEOID10 = data{fips}.ID\" "
```
lucasmbrown-usds, line 45: `'/home/data/score/csv/data{fips}.csv'` gets set elsewhere, right? Should we pass it in as a parameter to make sure if we change it elsewhere it's updated here?

score/tile/generate.py
```diff
@@ -0,0 +1,91 @@
...
+        print(cmd)
+        os.system(cmd)
+
+    # get a list of all json files to plug in the docker commands below
+    # (workaround since *.json doesn't seem to work)
+    geojson_list = ""
+    geojson_path = data_path / "score" / "geojson"
+    for file in os.listdir(geojson_path):
+        if file.endswith(".json"):
+            geojson_list += f"/home/data/score/geojson/{file} "
```
lucasmbrown-usds, line 57: Nit: above we used `data_path`, here we assume that `data_path` is a specific path (i.e., `/data`). Can we use a parameter here too? Otherwise if we change `data_path` everything will break.

score/tile/generate.py
```diff
@@ -0,0 +1,91 @@
...
+
+    if geojson_list == "":
+        print("No GeoJson files found. Please run scripts/download_cbg.py first")
+
+    # generate mbtiles file
+    # PWD is different for Windows
```
lucasmbrown-usds, line 63: nit: this gets used so often, can we make it a util function that returns `pwd`?

score/tile/generate.py
```diff
@@ -0,0 +1,91 @@
...
+    # get a list of all json files to plug in the docker commands below
+    # (workaround since *.json doesn't seem to work)
+    geojson_list = ""
+    geojson_path = data_path / "score" / "geojson"
+    for file in os.listdir(geojson_path):
+        if file.endswith(".json"):
+            geojson_list += f"/home/data/score/geojson/{file} "
+
+    if geojson_list == "":
+        print("No GeoJson files found. Please run scripts/download_cbg.py first")
```
lucasmbrown-usds, line 60: Nit: it seems like mostly we've switched to logging rather than printing, so I'd suggest we be consistent.

score/tile/generate.py
```diff
@@ -0,0 +1,91 @@
...
+    # remove existing score json files
+    score_geojson_dir = data_path / "score" / "geojson"
+    files_in_directory = os.listdir(score_geojson_dir)
+    filtered_files = [file for file in files_in_directory if file.endswith(".json")]
+    for file in filtered_files:
+        path_to_file = os.path.join(score_geojson_dir, file)
+        os.remove(path_to_file)
+
+    # join the state shape sqllite with the score csv
+    state_fips_codes = get_state_fips_codes()
```
lucasmbrown-usds, line 38: Nit: should we add more logging statements so it's easy to tell where we are in all these various stages?

score/tile/generate.py
```diff
@@ -0,0 +1,91 @@
...
+    cmd = (
+        'docker run --rm -it -v "'
+        + pwd
+        + '"/:/home klokantech/tippecanoe tippecanoe --drop-densest-as-needed -zg -o /home/data/tiles/block2010.mbtiles --extend-zooms-if-still-dropping -l cbg2010 -s_srs EPSG:4269 -t_srs EPSG:4326 '
+        + geojson_list
+    )
+    print(cmd)
+    os.system(cmd)
+
+    # if AWS creds are present, generate uncompressed toles
```
lucasmbrown-usds, line 77: nit: `tiles` not `toles`. 

Also, it doesn't seem like we're checking for AWS creds?

score/utils.py
```diff
@@ -0,0 +1,47 @@
...
+import zipfile
+
+
+def remove_files_from_dir(files_path: Path, extension: str = None) -> None:
+    for file in os.listdir(files_path):
+        if extension:
+            if not file.endswith(extension):
+                continue
+        else:
+            # don't rempove __init__ files as they conserve dir structure
```
lucasmbrown-usds, line 16: Nice

score/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+def remove_all_dirs_from_dir(dir_path: Path) -> None:
+    for filename in os.listdir(dir_path):
+        file_path = os.path.join(dir_path, filename)
+        if os.path.isdir(file_path):
+            shutil.rmtree(file_path)
+            logging.info(f"Removing directory {file_path}")
+
+
+def unzip_file_from_url(
```
lucasmbrown-usds, line 31: Very helpful!

score/utils.py
```diff
@@ -0,0 +1,47 @@
...
+    zip_file = open(zip_file_path, "wb")
+    zip_file.write(file_contents)
+    zip_file.close()
+
+    logging.info(f"Extracting {zip_file_path}")
+    with zipfile.ZipFile(zip_file_path, "r") as zip_ref:
+        zip_ref.extractall(zip_file_directory)
+
+    # cleanup temporary file
+    os.remove(zip_file_path)
```
lucasmbrown-usds, line 47: Should this method return a list of the extracted files, or the file contents, or something else? I suppose the person invoking the function knows they all landed at `zip_file_directory`, but it still might be helpful.

lucasmbrown-usds (changes requested):
Overall looking very solid, thanks for putting this together! I left lots of comments, mostly small nits, which will get reduced as we figure out our code style and conventions for the codebase. 

One big thing: we haven't written any tests. I know our tests here are likely to be not super high value because there's so much happening externally, but we all know that if you start not writing tests in a codebase you never really catch up, because then there's a backlog of untested code nobody wants to bother cleaning up and testing. 

@switzersc-usds what are your thoughts on our testing strategy here?

esfoobar-usds:
Thanks! I will check out the comments and address or punt. I've been thinking a lot about testing for this part of the application and to be honest the only thing we could test is the score output itself using dummy data/test fixtures. 

My philosophy with testing is to test what is in your control and that's not tested already by the library itself. I.e. we don't want to bother testing if zipped files are being unzipped (a bit of exaggeration to make my point clear). 

There is also quite a bit to figure out on the data set definition side which would touch on the YAML side of things and maybe there's some testing opportunities there.

lucasmbrown-usds:
> to be honest the only thing we could test is the score output itself using dummy data/test fixtures

Yeah, that's not a bad idea. We can also test just whether we're making the outbound calls, making calls to write to disk, etc. It's not super informative but having no tests at all could potentially set a bad precedent.

score/application.py
esfoobar-usds: There is no style recommendations on imports that I could find, [only this recommendation about breaking long ones](https://www.python.org/dev/peps/pep-0328/#rationale-for-parentheses).

However, readability is one of the most important recommendations, I would find difficult to read long module names within the code -- you would hit the 80 character line length easily and would need to use `\` characters across the code, so I am not in favor, but curious to hear what others think.

score/etl/sources/census/etl.py
esfoobar-usds: Fixed.

score/etl/sources/census/etl.py
esfoobar-usds: The TODO here is to execute the command directly on the operating system (most surely Ubuntu) versus spawning a Docker container. But I am struggling to think a scenario where you *wouldn't* run this as Docker commands out right. Maybe the answer is that if you want to run this locally outside of Docker you need to install GDAL locally?

score/tile/generate.py
esfoobar-usds: There's never a case where fewer logging statements make sense. 😉

score/tile/generate.py
esfoobar-usds: This was more of thinking out loud, removed.

score/utils.py
esfoobar-usds: That command only erases the `download.zip` file, so it wouldn't make too much sense here.

score/etl/sources/census/etl.py
esfoobar-usds: Good catch. Fixed.

score/etl/sources/census/etl.py
esfoobar-usds: Fixed.

score/requirements.txt
esfoobar-usds: This is handled by `Poetry` and not meant for human consumption, just for CI/CD consumption. Which is the advantage of using `Poetry` in the first place.

score/tile/generate.py
esfoobar-usds: For the generate tiles, in my opinion, we do want to tie the clear up and generation on the same script, as the moment you're invoking this command you want to nuke whatever is there totally, versus checking if some tiles exist and others not.

score/etl/sources/census/etl_utils.py
esfoobar-usds: Nope, this is stored on disk the first time it's run (see line 29 below, so it would be downloaded just once per container generation).

score/tile/generate.py
esfoobar-usds: Fixed.

score/tile/generate.py
esfoobar-usds: No, this is the path in the context of the mounted volume on the Docker container, which will stay the same always, so not applicable here. Good callout though.

switzersc-usds:
FYI I am reviewing this now and trying out the setup locally; hoping to leave a full review by COB or firs thing Monday. To weigh in briefly on the testing convo -- yes to adding a test for the score at minimum, as well as testing critical fail paths for things we can (eg what if a data source returns 404). 

Regarding styles, we should have a conversation about engineering styles and conventions and get those documented. This isn't super urgent to me but we should do this sooner rather than later otherwise we will proliferate styles across the code base and potentially get bogged down in nitpick conversations. Probably a good convo for right after we're back from the July 4 holiday.

score/README.md
```diff
@@ -1,49 +1,65 @@
...
 
 ## Downloading Census Block Groups GeoJSON and Generating CBG CSVs
 
 - Make sure you have Docker running in your machine
 - Start a terminal
 - Change to this directory (i.e. `cd score`)
-- Activate your virtualenv (see above)
-- Run `python scripts/download_cbg.py`
+- If you want to clear out all data and tiles from all directories, you can run: `poetry run python application.py data-cleanup`.
+- Then run `poetry run python application.py census-data-download`
```
switzersc-usds, line 33: When I try to run this, I get an error: `ModuleNotFoundError: No module named 'dynaconf'` -- I think we need to add `poetry install` to the steps, either here or in the section above.

score/README.md
```diff
@@ -1,49 +1,65 @@
...
-see [python-markdown docs](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/python-markdown).
+## Local development
+
+You can run the Python code locally to develop, using Poetry. However, to generate the census data you will need the [GDAL library](https://github.com/OSGeo/gdal) installed locally. Also to generate tiles for a local map, you will need [Mapbox tippeanoe](https://github.com/mapbox/tippecanoe)
+
+- Start a terminal
+- Make sure you have Python 3.9 installed: `python -V` or `python3 -V`
+- We use [Poetry](https://python-poetry.org/). Please follow the instructions to download to your system
+- Change to this directory (i.e. `cd score`)
+- Start a Poetry shell with `poetry shell`
```
switzersc-usds, line 25: Add `poetry install` right before this?

switzersc-usds: Also, do we need to run `poetry shell` in this list of steps? Is this necessary before going to the next section? Note that the next section uses `poetry` in the commands so seems like it assumes you're not in a poetry shell.

switzersc-usds (commented):
Leaving these comments for now if they're useful. About to pull the latest and start an updated review :)

score/README.md
esfoobar-usds: This is fixed i a subsequent commit, hadn't updated `reuirements.txt` from Poetry.

score/README.md
esfoobar-usds: You're absolutely right -- once in a shell you can run the commands without the `poetry` in front. You can also do `poetry run [cmd]` without the shell. I'll update for folks to use the latter.

score/etl/sources/census/etl.py
esfoobar-usds: New commit outdates this.

score/tile/generate.py
esfoobar-usds: This path is in the context of the container, so it will always stay the same.

esfoobar-usds: Also outdated by last commit.

score/tile/generate.py
esfoobar-usds: Also outdated by last commit.

score/tile/generate.py
esfoobar-usds: Also outdated by last commit.

score/README.md
```diff
@@ -4,18 +4,33 @@
 
 - Start a terminal
 - Make sure you have Python 3.9 installed: `python -V` or `python3 -V`
-- Create a `virtualenv` in this folder: `python -m venv venv`
-- Activate the virtualenv
-  - Windows: `./venv/Scripts/activate`
-  - Mac/Linux: `source venv/bin/activate`
-- Install packages: `pip install -r requirements.txt`
-  - If you are a Windows user, you might need to install Build Tools for Visual Studio. [Instructions here](https://stackoverflow.com/a/54136652)
+- We use [Poetry](https://python-poetry.org/). Please follow the instructions to download
```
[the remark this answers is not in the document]
esfoobar-usds: Updated.

score/README.md
```diff
@@ -4,18 +4,33 @@
...
+- Change to this directory (i.e. `cd score`)
```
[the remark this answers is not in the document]
esfoobar-usds: Resolved.

score/README.md
esfoobar-usds: Updated.

score/application.py
esfoobar-usds: Both added to tracking ticket https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/255

score/application.py
esfoobar-usds: Tracking on this ticket: https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/255

esfoobar-usds:
Merging this now, per our Zoom convo. Testing will be part of the ETL PR [here](https://github.com/usds/justice40-tool/pull/250).

Issue #206, opened 2021-06-22
As a developer, I want to ingest the unemployment data set from ACS, so that it can be tested as an input to the cumulative impact score.


BethMattern:
**Description**
The unemployment data set from ACS needs to be available in the screening tool so that it can be used as an input into the cumulative impact score. 

This data set will eventually be displayed in the map for selected census block groups and exported in the list of communities.

The system needs to classify this data set as a "Socioeconomic factors" indicator. A future version of the cumulative impact score will need this classification as part of the formula.

Both the raw value and the value converted to percentiles for the entire country must be available in the tool.

This data is not stored at the census block group level - it's at the census tract level. There is a separate card representing the work it will take to develop the methodology for combining data from census tracts into census block groups. 

**Solution**
A clear and concise description of the solution.

**Describe alternatives you've considered**

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**
- [ ] Unemployment at the census tract level is available to be used as an input into the score
- [ ] Unemployment is classified as a "Socioeconomic factors" indicator
- [ ] Unemployment is normalized into a percentile for the entire country
- [ ] Unemployment is available to be displayed in the map UI
- [ ] Unemployment is available to be downloaded in the CSV

Issue #177, opened 2021-06-17
Research source(s) for housing costs and transportation costs as possible indicators to use as in input to the score


BethMattern:
We need to find a source at the census block level for housing cost burden and transportation costs as possible indicators to use as an input into the score.

Results of research by Beth:
The Housing and Transportation (H+T®) Affordability Index provides a comprehensive view of affordability that includes both the cost of housing and the cost of transportation at the neighborhood level. https://htaindex.cnt.org

This data set is represented in **row 92** of the Justice 40 Data Set Submissions worksheet.

This data is at the census block group level and must be downloaded by state - so there would be 50 .csv files.
There is no data available for Puerto Rico or territories.

The data includes the following metrics that would be good candidates for representing affordable housing and affordable transit availability in the impact score:

Housing Costs % Income for the National Typical Household (h_nmi)
Transportation Costs % Income for the National Typical Household (t_nmi)
Housing + Transportation Costs % Income for the National Typical Household (ht_nmi)

[Data dictionary](https://htaindex.cnt.org/download/data-dictionary.php) 

**Information about this data:**
The Index shows that transportation costs vary between and within regions depending on neighborhood characteristics:

People who live in location-efficient neighborhoods—compact, mixed-use, and with convenient access to jobs, services, transit and amenities—tend to have lower transportation costs.
People who live in location-inefficient places—less dense areas that require automobiles for most trips—are more likely to have higher transportation costs.
The traditional measure of affordability recommends that housing cost no more than 30% of household income. Under this view, a little over half (55%) of US neighborhoods are considered “affordable” for the typical household. However, that benchmark fails to take into account transportation costs, which are typically a household’s second-largest expenditure. The H+T Index offers an expanded view of affordability, one that combines housing and transportation costs and sets the benchmark at no more than 45% of household income.

When transportation costs are factored into the equation, the number of affordable neighborhoods drops to 26%, resulting in a net loss of 59,768 neighborhoods that Americans can truly afford. The key finding from the H+T Index is that household transportation costs are highly correlated with urban environment characteristics, when controlling for household characteristics.

**Last Updated**
2017 Update: H+T Index Release Notes

Uses 2015 American Community Survey (ACS) data as input
Uses 2014 Longitudinal Employer-Household Dynamics data
Updated AllTransitTM data
Expand geographic coverage to metropolitan, micropolitan regions and rural counties
Enhanced transportation model including interaction variables

BethMattern:
Here's the .csv for New Mexico
[...]

PR #214, opened 2021-06-23
Ingest census data directly, add unemployment to the score

lucasmbrown-usds:
For experimentation purposes, this PR adds direct access to 2019 ACS data (for variables not included on EJSCREEN). I also added the housing and transportation index.

This accomplishes #206 and integrates the data source from #177. 

I've also gone ahead and joined this in the `score_calc` file, and added it into the socioeconomic benefits bucket for the score calculation. 

I updated the `scoring_comparison` to point to these files appropriately. This doesn't change the resulting scores very significantly:

Screen Shot 2021-06-23 at 11 01 58 AM


esfoobar-usds (approved):
Overall looks good, fixed some issues regarding directory creation and a zip file name var issue.

One small concern I have is the [censusdata](https://jtleider.github.io/censusdata/index.html) library. It's maintained by one user and has few stars and commits are sparsed, so monitoring that the etl processes don't break will be important.

lucasmbrown-usds:
> Overall looks good, fixed some issues regarding directory creation and a zip file name var issue.

Thanks for fixing that! 

> One small concern I have is the [censusdata](https://jtleider.github.io/censusdata/index.html) library. It's maintained by one user and has few stars and commits are sparsed, so monitoring that the etl processes don't break will be important.

Yeah, that's a good flag. It wouldn't be hard to replace with direct API access. I've created #229 to track this!

Issue #191, opened 2021-06-18
As a stakeholder interested in the cumulative impact score, I want to see more states in the map, so that I can further analyze the score results.


BethMattern:
**Description**
In order to get more feedback about the score formula, we need to add a few more states to the map:

A mix of:
- plains (rural) state - how about Montana
- southern state - how about Alabama
- NY
- NJ

**Solution**
A clear and concise description of the solution.

**Describe alternatives you've considered**

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**
- [ ] USDS demo can show CEQ stakeholders these additional states on a map

switzersc-usds:
Also Puerto Rico if possible.

esfoobar-usds:
@NatHillardUSDS let me know if you need anything from me. Like I mentioned in a past meeting, you can restrict the states to generate locally using slicing in this [line](https://github.com/usds/justice40-tool/blob/main/score/scripts/generate_mbtiles.py#L37). Just pass the GEOID10 for the 4 states, i.e. `state_fips_codes = ['01', '05', '13', '14']` (note: not real codes)

PR #226, opened 2021-06-24
Issue 191 - Multi-state visualization

NatHillardUSDS:
Addresses #191 - "As a stakeholder interested in the cumulative impact score, I want to see more states in the map, so that I can further analyze the score results."

Will need further visual tweaks but this works with current data

### DIFF
--- a/client/src/components/map.tsx
+++ b/client/src/components/map.tsx
@@ -92,8 +92,8 @@ const MapWrapper = ({features}: IMapWrapperProps) => {
   const popupContent = React.useRef<HTMLDivElement>(null);
 
   const overlayRef = useRef() as React.MutableRefObject<Overlay>;
-  mapRef.current = map;
-  overlayRef.current = currentOverlay;
+  mapRef.current = map!;
+  overlayRef.current = currentOverlay!;
 
   useEffect( () => {
     // create and add initial vector source layer, to be replaced layer
@@ -101,14 +101,14 @@ const MapWrapper = ({features}: IMapWrapperProps) => {
       source: new VectorSource(),
     });
 
-    popupCloser.current.onclick = function() {
+    popupCloser.current!.onclick = function() {
       overlay.setPosition(undefined);
-      popupCloser.current.blur();
+      popupCloser.current!.blur();
       return false;
     };
 
     const overlay = new Overlay({
-      element: popupContainer.current,
+      element: popupContainer.current!,
       autoPan: true,
       autoPanAnimation: {
         duration: 250,
@@ -171,7 +171,7 @@ const MapWrapper = ({features}: IMapWrapperProps) => {
     properties = selectedFeature.getProperties();
   }
 
-  const getTitleContent = (properties) => {
+  const getTitleContent = (properties: { [key: string]: any; }) => {
     return (
       <table className={styles.popupHeaderTable}>
         <tbody>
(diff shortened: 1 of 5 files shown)


client/src/components/map.tsx
```diff
@@ -20,17 +21,25 @@ const mapConfig = {
...
+        'https://c.basemaps.cartocdn.com/light_nolabels/{z}/{x}/{y}.png',
+        'https://d.basemaps.cartocdn.com/light_nolabels/{z}/{x}/{y}.png',
       ],
     },
     'custom': {
-      'projection': 'EPSG:3857',
       'type': 'vector',
       'tiles': [
-        'https://gis.data.census.gov/arcgis/rest/services/Hosted/VT_2019_150_00_PY_D1/VectorTileServer/tile/{z}/{y}/{x}.mvt',
+        'http://localhost:8080/data/tl_2010_bg_with_data/{z}/{x}/{y}.pbf',
```
TomNUSDS, line 33: `localhost` url. Maybe add a `todo` to fix it?

client/src/components/map.tsx
```diff
@@ -44,50 +53,100 @@ const mapConfig = {
...
+    React.MutableRefObject<HTMLInputElement>;
+  const mapRef = useRef() as React.MutableRefObject<Map>;
+
+
+  const popupContainer = React.useRef<HTMLDivElement>(null);
+  const popupCloser = React.useRef<HTMLAnchorElement>(null);
+  const popupContent = React.useRef<HTMLDivElement>(null);
+
+  const overlayRef = useRef() as React.MutableRefObject<Overlay>;
+  mapRef.current = map!;
```
TomNUSDS, line 114: I'm not familiar with `map!` in this context. What is it doing?

client/src/components/map.tsx
```diff
@@ -112,8 +171,105 @@ const MapWrapper = ({features}: IMapWrapperProps) => {
...
+  let properties;
+  if (selectedFeature) {
+    properties = selectedFeature.getProperties();
+  }
+
+  const getCategorization = (percentile: number) => {
+    let categorization = '';
+    if (percentile > 0.75 ) {
+      categorization = 'Prioritized';
+    } else if (0.60 < percentile && percentile < 0.75) {
```
TomNUSDS, line 205: nice choice for formatting. 
Should one of these choices be >= ? (e.g. what if percentile is exactly 0.75)

TomNUSDS (commented):
Looks great. Think I spotted on possible boundary condition?

client/src/components/map.tsx
NatHillardUSDS: Good catch! will update

client/src/components/map.tsx
NatHillardUSDS: Good call - in the latest push I use a remote url and add the localhost as a comment

client/src/components/map.tsx
NatHillardUSDS: this is the non-null assertion operator ([ref](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator)) - you occasionally have to use this in Swift and I was surprised to find it in Typescript also - basically it's assuring the compiler that you know better and that this thing actually does exist.

That said, while this particular usage had the effect of temporarily silencing the compiler, in this case it actually might not exist, so this should be handled differently. In general we should move away from using `ref`s in this way - it works but there are better ways to do this.

NatHillardUSDS: I can look into other solutions and push a spot fix for this

TomNUSDS: I've never seen it used to stop linter/compiler complaints this way. I'm fine with it with maybe a comment or `todo`.

TomNUSDS (approved):
LGTM!

PR #231, opened 2021-06-25
Fix Github Actions deploy for merges to main

NatHillardUSDS:
Up to now, only deploying to PR-based hash directories has worked; this attempts to get things working for "main" as well, which will always point to the latest merged commit to main

Paradoxically, we won't know if this truly works until we merge this PR...


TomNUSDS (approved):
Seems good to me.

PR #250, opened 2021-06-28
Initial ETL architecture 

esfoobar-usds:


score/etl/base.py
```diff
@@ -0,0 +1,27 @@
+from pathlib import Path
+
+from config import settings
+from utils import unzip_file_from_url
+
+
+class Etl:
```
lucasmbrown-usds, line 7: nitty nit: I'm really averse to acronyms for a few different reasons, haha. How about `ExtractTransformLoad`?

score/etl/base.py
```diff
@@ -0,0 +1,27 @@
...
+    def get_yaml_config(self):
+        pass
+
+    @staticmethod
+    def extract(source_url: str = None, extract_path: Path = None):
+        if source_url and extract_path:
+            unzip_file_from_url(source_url, TMP_PATH, extract_path)
+
+    @staticmethod
+    def translate():
```
lucasmbrown-usds, line 21: Nit: I think this is usually `transform` (e.g., https://en.wikipedia.org/wiki/Extract,_transform,_load)

esfoobar-usds: Oops, that was a typo. good catch!

Issue #201, opened 2021-06-21
As an EVCM, I want the map to make sense at varying zoom levels so that I'm not confused.


switzersc-usds:
**Description**
This card is about implementing different zoom-level based modifications to improve the UX of the map. It is dependent on the zoom spike. 

**Solution**
A clear and concise description of the solution.

**Describe alternatives you've considered**

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**

NatHillardUSDS:
Spike is now completed, but to implement these changes The Right Way we will need the data-side ticket, #209 . Setting as blocked on that now

Issue #196, opened 2021-06-21
[SPIKE] As a developer, I want to understand the options for implementing various zoom levels for national data at the census block group level.


switzersc-usds:
**Description**
We have a bunch of data to display at the census block group level that is easy to display at a relatively zoomed in level (eg zoom level 9); however, we want to display this as more zoomed out levels to give users an understanding of the data at a higher level. We'd love to have this conversation with the community to discover and understand options for implementing various zoom levels in terms of data, tiling, and UX design. 

**Solution**
TBD

**Describe alternatives you've considered**
TBD

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**

mikelmaron:
@switzersc-usds this is challenge I've been experimenting on within the [earthrise collaboration on Justice 40](https://earthrise-media.github.io/justice-map/). 

Because lower zoom tiles (between 1-9) will each cover more territory, by default the tiles will contain more features than tiles at higher zooms (9+), and then hit limits on MVT tile size (500kb), suffer performance hits, or have visible artifacts from dropping or over simplifying features. My approach has been to have different tiling strategies for low zoom vs high zoom tiles, with low zoom tiles having somewhat less detail but still giving a visually accurate understanding of the data. And then, in the style, gradually blend the low zoom and high zoom tiles together to smooth out the transition.

For the data preprocessing, the approach I took was to assign each block group's indicator value into 10 buckets, then dissolve neighbors by buckets. The result has significantly fewer features, that have no problem rendering at low zoom. The GeoJSON output is then processed into tiles. See [this notebook](https://github.com/earthrise-media/justice-map/blob/main/notebooks/bucket_d_pm25_2.ipynb) for details. 

For the front end styling, each layer's style is dynamically built from a configuration. A single indicator will have two fill layers -- one for high and one for low. The alpha of the fill layers are interpolated between zoom 9 and 11, with low zoom layers fading out, and high zoom layers fading in. Check out [the configuration](https://github.com/earthrise-media/justice-map/blob/main/lib/pmap.ts#L65-L194), and how the [style is built](https://github.com/earthrise-media/justice-map/blob/main/lib/pmap.ts#L22-L60).

If you find this approach helpful, let me know if you have any questions.

NatHillardUSDS:
[...]

PR #265, opened 2021-06-29
Zoom fade for higher zoom levels

NatHillardUSDS:
Addresses #196 and #201 - adds z
* Creates a unified style file modeled after zoom behavior of justice-map
* Uses this style file with ol-mapbox-style 
* Factors out popup logic into dedicated popup component (currently only works for openlayers)
* Factors out zoom logic into dedicated component
* Creates alternative implementation for comparison purposes accessible via flag


client/src/components/mapPopup.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }, []);
+
+  useEffect( () => {
+    if (position && currentOverlay) { // may be empty on first render
+      currentOverlay.setPosition(position);
+    }
+  }, [position]);
+
+  const getCategorization = (percentile: number) => {
+    let categorization = '';
```
TomNUSDS, line 51: Doesn't need to be initialized since it's always set below. (I don't care, but my linter noticed it.)

client/src/components/mapboxMap.tsx
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable no-unused-vars */
+import React, {useRef, useEffect, useState} from 'react';
+import {Map, NavigationControl, Popup, MapLayerMouseEvent, Coordinate} from 'mapbox-gl';
```
TomNUSDS, line 3: Unused imports `Popup, MapLayerMouseEvent, Coordinate`?

client/src/components/mapboxMap.tsx
```diff
@@ -0,0 +1,50 @@
...
+import * as styles from './mapboxMap.module.scss';
+import mapStyle from '../data/mapStyle';
+import ZoomWarning from './zoomWarning';
+
+const MapboxMap = () => {
+  const mapContainer = React.useRef<HTMLDivElement>(null);
+  const map = useRef<Map>() as React.MutableRefObject<Map>; ;
```
TomNUSDS, line 10: Bonus `; ;` at end.

client/src/components/openlayersMap.tsx
```diff
@@ -0,0 +1,115 @@
...
+import {Coordinate} from 'ol/coordinate';
+import olms from 'ol-mapbox-style';
+import mapStyle from '../data/mapStyle';
+import ZoomWarning from './zoomWarning';
+import MapPopup from './mapPopup';
+import * as styles from './openlayersMap.module.scss';
+
+interface IMapWrapperProps {
+  features: Feature<Geometry>[],
+};
```
TomNUSDS, line 18: `;` not needed.

client/src/components/openlayersMap.tsx
```diff
@@ -0,0 +1,115 @@
...
+  // create state ref that can be accessed in OpenLayers onclick callback function
+  //  https://stackoverflow.com/a/60643670
+  const mapRef = useRef() as React.MutableRefObject<Map>;
+  if (map) {
+    mapRef.current = map;
+  }
+
+  useEffect( () => {
+    const view = new View({
+      center: fromLonLat([-86.502136, 32.4687126]),
```
TomNUSDS, line 41: Maybe move to named const at some point so we know where/why this is lat/lng magic number?

client/src/components/openlayersMap.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const initialFeaturesLayer = new VectorLayer({
+      source: new VectorSource(),
+    });
+
+    const initialMap = new Map({
+      target: mapElement.current,
+      view: view,
+      controls: [],
+    });
+    const currentZoom = Math.floor(initialMap.getView().getZoom()!);
```
TomNUSDS, line 55: maybe instead of trailing `!` you can do a `|| 4` for a default zoom. I know it's just to make the typechecker happy, but in theory, `getZoom()` could return `undefined`?

client/src/components/openlayersMap.tsx
```diff
@@ -0,0 +1,115 @@
...
+    mapRef.current.forEachFeatureAtPixel(event.pixel, (feature) => {
+      setSelectedFeature(feature);
+      return true;
+    });
+
+    setCurrentOverlayPosition(clickedCoord);
+  };
+
+  const handleMoveEnd = () => {
+    const newZoom = Math.floor(mapRef.current.getView().getZoom()!);
```
TomNUSDS, line 97: Same feedback as above.

client/src/components/openlayersMap.tsx
```diff
@@ -0,0 +1,115 @@
...
+      if (extent != null) {
+        // fit map to feature extent (with 100px of padding)
+        map?.getView().fit(extent, {
+          padding: [100, 100, 100, 100],
+        });
+      }
+    }
+  }, [features]);
+
+  const handleMapClick = (event: { pixel: any; }) => {
```
TomNUSDS, line 85: Style suggestion: `pixel: any;` Remove bonus `;`?

client/src/components/openlayersMap.tsx
```diff
@@ -0,0 +1,115 @@
...
+    if (currentZoom != newZoom) {
+      setCurrentZoom(newZoom);
+    }
+  };
+
+  return (
+    <>
+      <div ref={mapElement} className={styles.mapContainer}/>
+      {map?
+        <MapPopup selectedFeature={selectedFeature!} map={map!} position={currentOverlayPosition} /> :
```
TomNUSDS, line 107: OK. I like the use of `!` here as long as React doesn't care about `undefined` vs `null`.

client/src/pages/cejst.tsx
```diff
@@ -50,7 +48,7 @@ const CEJSTPage = ({location}: IMapPageProps) => {
           </p>
         </Alert>
         <h2>Explore the Tool</h2>
-        <MapWrapper features={features} />
+        <MapWrapper/>
```
TomNUSDS, line 51: So, you're not passing in any `features` prop, but the MapWrapper checks the length of this prop. Does typescript/react just assume a default, empty `Feature<Geometry>[]`?
(I don't thing this is not a bug, I'm just doing a visual code inspection and my in-brain-compiler doesn't know what's happening here.)
Feel free to just chat with me in person instead of answering this in the PR.

TomNUSDS (dismissed):
Minor things, but nothing major. Once you've looked them over, it's good to go. (assuming the git automated check failing isn't serious)

client/src/components/zoomWarning.module.scss.d.ts
```diff
@@ -0,0 +1,13 @@
+declare namespace ZoomWarningModuleScssNamespace {
+    export interface IZoomWarningModuleScss {
+        zoomWarning: string;
+    }
+  }
+  
+  declare const ZoomWarningModuleScssModule: ZoomWarningModuleScssNamespace.IZoomWarningModuleScss & {
```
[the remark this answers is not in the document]
NatHillardUSDS: Yeah! apologies, I switched to a new partition and lost my VSCode settings. latest fix has linter fixes for all files

client/src/components/openlayersMap.tsx
NatHillardUSDS: Yeah, there is probably a better way to do this - incidentally this results in a bug, which I had filed [here](https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/263) -  in hte popup itself, we need the `ref` for the html element, but if we conditionally remove it we lose that ref, so we need some other way to conditionally not render something while keeping a ref to it

client/src/pages/cejst.tsx
NatHillardUSDS: Yeah, this originally came from [here](https://taylor.callsen.me/using-openlayers-with-react-functional-components/) - the idea is that you can assign a `features` prop with an array of features, in this case from a geoJSON file, which will be added to the map - this could be useful say if we had a dropdown with the ability to select a layer, the output of which would be a geoJSON representation; we don't yet have this use case, but if/when we do this will support it. Notably it may make sense to remove this until that is the case (and that may never be able to be the case given our features are not represented at the moment with geoJSON)

client/src/data/mapStyle.tsx
```diff
@@ -0,0 +1,179 @@
...
+    'geo': {
+      'type': 'raster',
+      'tiles': [
+        'http://mt0.google.com/vt/lyrs=p&hl=en&x={x}&y={y}&z={z}',
+      ],
+    },
+    'custom': {
+      'type': 'vector',
+      'tiles': [
+        'http://usds-geoplatform-justice40-website.s3-website-us-east-1.amazonaws.com/0624_demo/{z}/{x}/{y}.pbf',
```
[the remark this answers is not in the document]
NatHillardUSDS: 👍 good call! we actually already do have our https cloudfront url, which will also incidentally make things faster - I'll switch to using that instead

client/src/components/openlayersMap.tsx
TomNUSDS: Maybe just hiding/showing the element using css? In Vue, I'd just add a css class that had `display:none;` then control showing it by making the class reactive (add/remove the classname from the element). Not sure if that would work here.

client/src/components/mapPopup.module.scss
```diff
@@ -15,6 +8,7 @@
   bottom: 12px;
   left: -50px;
   min-width: 280px;
+  overflow: scroll;
```
ginabeena, line 11: so curious how this looks on different screen sizes haha

client/src/components/mapPopup.tsx
```diff
@@ -0,0 +1,130 @@
...
+    selectedFeature: FeatureLike;
+    position: Coordinate;
+}
+
+const MapPopup = ({map, selectedFeature, position}: IMapPopupProps) => {
+  const popupContainerElement = useRef<HTMLDivElement>(null);
+  const popupCloserElement = useRef<HTMLAnchorElement>(null);
+  const popupContentElement = useRef<HTMLDivElement>(null);
+  const [currentOverlay, setCurrentOverlay] = useState<Overlay>();
+
```
ginabeena, line 20: This looks like some witchcraft 👀  What's happening with `useRef<HTMLDivElement>(null)` ? Just for educational purposes.

client/src/components/mapboxMap.module.scss
```diff
@@ -0,0 +1,19 @@
+.mapContainer {
+  height: 676px;
+  margin-bottom: 29px;
+  max-width: revert;
+  margin-top: 50px;
+}
+
+.sidebar {
+  background-color: rgba(35, 55, 75, 0.9);
```
ginabeena, line 9: Might want to use color variables if we're going to be tweaking a lot. Not sure it's worth that though.

client/src/components/mapboxMap.tsx
```diff
@@ -0,0 +1,50 @@
...
+const MapboxMap = () => {
+  const mapContainer = React.useRef<HTMLDivElement>(null);
+  const map = useRef<Map>() as React.MutableRefObject<Map>;
+  const [lng, setLng] = useState(-86.502136);
+  const [lat, setLat] = useState(32.4687126);
+  const [zoom, setZoom] = useState(3);
+
+  useEffect(() => {
+    // Only initialize once
+    if (map.current && mapContainer.current) return;
```
ginabeena, line 17: Gotta love stateful DOMs in javascript/ts

client/src/components/openlayersMap.tsx
```diff
@@ -0,0 +1,118 @@
...
+import MapPopup from './mapPopup';
+import * as styles from './openlayersMap.module.scss';
+
+const DEFAULT_ZOOM = 4;
+const DEFAULT_US_CENTER = [-86.502136, 32.4687126];
+
+interface IMapWrapperProps {
+  features: Feature<Geometry>[],
+}
+
```
ginabeena, line 22: This is very readable stuff @NatHillardUSDS 👍

client/src/components/openlayersMap.tsx
```diff
@@ -0,0 +1,118 @@
...
+  useEffect( () => {
+    if (features.length) { // may be empty on first render
+      // set features to map
+      featuresLayer?.setSource(
+          new VectorSource({
+            features: features,
+          }),
+      );
+      const extent = featuresLayer?.getSource().getExtent();
+      if (extent != null) {
```
ginabeena, line 79: Sometimes I'm seeing `if(foo)` and other times `if(foo != null)` any reason for that?

client/src/components/zoomWarning.module.scss.d.ts
```diff
@@ -0,0 +1,13 @@
...
+    }
+  }
+
+declare const ZoomWarningModuleScssModule: ZoomWarningModuleScssNamespace.IZoomWarningModuleScss & {
+    /** WARNING: Only available when `css-loader` is used without `style-loader` or `mini-css-extract-plugin` */
+    locals: ZoomWarningModuleScssNamespace.IZoomWarningModuleScss;
+  };
+
+export = ZoomWarningModuleScssModule;
+
```
ginabeena, line 13: Okay this is interesting. Wanna explain how this is working? Never seen an interface between typescript and scss.

ginabeena (approved):
For my money, this is pretty 🔥 nice work on readability @NatHillardUSDS. I feel like there could be gotchas in some of these if checks, but if it works locally then it's good to go for me.  Would be so cool to have a headless browser click around in the UI when we do merges for frontend. :shipit:

client/src/components/mapPopup.module.scss
NatHillardUSDS: Yes - Definitely true, we need an explicit mobile pass - I just filed #268 to make sure we do this

client/src/components/mapPopup.tsx
NatHillardUSDS: These references are effectively ivars pointing to DOM elements - after initial load, they persist between renderings so you can reference the current html element. More [here](https://reactjs.org/docs/refs-and-the-dom.html)

client/src/components/mapboxMap.module.scss
NatHillardUSDS: Good callout!

client/src/components/mapboxMap.tsx
NatHillardUSDS: yeah, this is a side effect of using `ref`s for sure - there might bet a better way to do this but I haven't come across this yet

client/src/components/openlayersMap.tsx
NatHillardUSDS: True, fixed for consistency -- `!= null` is actually incorrect here come to think of it as `null` and `undefined` would behave differently but we don't really want that - for now moved to simple `if(extent)`

client/src/components/zoomWarning.module.scss.d.ts
NatHillardUSDS: Yeah this is from `gatsby-plugin-scss-typescript` initially - we abandoned this in favor of `gatsby-plugin-sass` due to some bugs we came across, but one beneficial thing it does is generates these type files for us - so I temporarily installed the plugin, ran `gatsby develop` to generate the interfaces, then deleted the plugin again. We should work through these problems at some point to make it easier to make modules, but for now you can also optionally copy an existing `.d.ts` file

client/src/components/openlayersMap.tsx
NatHillardUSDS: Ah, good call - that would definitely work -- in my latest push just now though I check instead in the `position` function for the existence of `selectedFeature` and key off of that instead, this seems to work!

score/ipython/hud_housing_etl.ipynb
```diff
@@ -0,0 +1,274 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c21b63a3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
```
TomNUSDS, line 10: Nothing just a comment, but all this source-as-array seems less than ideal, but I guess that's .ipynb?

TomNUSDS (approved):
LGTM. Thanks for spending the extra time address comments from earlier.

score/ipython/hud_housing_etl.ipynb
NatHillardUSDS: Yeah- this bit was from having merged with Lucas's changes while developing this branch. Agreed that notebooks outside of being rendered look messy in source; notably with Jorge's latest changes we've got more pure python fns now

Issue #245, opened 2021-06-28
What percent of our census block groups are not in a census tract of the comparison indicator?


BethMattern:
**Problem statement / question**
The comparison tool already answers the question, what % of census tracts from the comparison indicator (e.g., CalEnviroScreen) include one or more census block groups from the current version of the CEJST score. 

This asks an inverse question: what percent of CBGs from the current version of the CEJST score exist outside the areas prioritized by a comparison indicator? For instance, we might find that 10% of CBGs from the CEJST score are in census tracts that CalEnviroScreen does not consider disadvantaged communities.

lucasmbrown-usds:
In review in https://github.com/usds/justice40-tool/pull/266

PR #266, opened 2021-06-29
Scores D & E

lucasmbrown-usds:
Updated score draft. 

I made various updates to the ETL and comparison tool scripts as well.

This includes this ticket: #245. 

# Score D
Poverty (Less than 200% of federal poverty line)
Less than high-school education
Linguistic Isolation
Unemployment
Housing burdened low-income
*  Housing-Burdened Low-Income Households. Percent of households in a census tract that are both low income (making less than 80% of the HUD Area Median Family Income) and severely burdened by housing costs (paying greater than 30% of their income to housing costs). (5-year estimates, 2012-2016).

Normalization: min-max

# Score E 

The same, except normalization with percentiles.


ginabeena (approved):
Looks good to me @lucasmbrown-usds :shipit: 

As an aside, the sooner we can have integration tests around this data the better.

esfoobar-usds (approved):
This looks really, good to me 🍏 

One thing I was thinking that we haven't expressed anywhere is that the data set generation of output can't depend on each other (they don't at the moment), since they will be run out of order, depending on their TTLs. IF this is something we require, we might need to express that dependency somewhere on the YAML file spec.

Issue #271, opened 2021-06-29
As a contributor, I want to refactor the comparison tool notebooks so that they are DRY and easy to reuse and add questions/comparing data sets.


switzersc-usds:
**Description**
The notebook for the scoring comparison will get long and unwieldy as we add data sets for comparison and questions to ask. We want to refactor it (and leave as notebooks or convert to a python app) so that it is easier to add to and more modular.

**Solution**
A clear and concise description of the solution.

**Describe alternatives you've considered**

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**

switzersc-usds:
Some ideas from office hours:
- Don't refactor until we've added one more comparison data set so we can start to see patterns
- Writing out the steps (or making a mermaid diagram??) of the comparison code workflow would be helpful for guiding a refactor.
- Probably should refactor sooner rather than later so that we don't have tons of work to do to modify parallel data PRs.

switzersc-usds:
Can we have a fun colorful comparison matrix to show a comparison score between a J40 score and all the datasets?

switzersc-usds:
@lucasmbrown-usds can you please add an estimate to this?

Issue #234, opened 2021-06-28
As a contributor to J40, I want to compare the J40 score results to HUD RECAP, so that I can understand how the J40 score compares.


BethMattern:
**Description**
The J40 team is working to develop the best possible score for identifying the most underserved and overburdened communities. In order to assess the quality of the score and build credibility in the EJ community, the J40 team build a tool to compare the output of the most recent version of the J40 score with other data sources that generate scores.

We need help adding more data sources for comparison to our tool. We would like to compare to the HUD RECAP score, that can be found here: https://hudgis-hud.opendata.arcgis.com/datasets/56de4edea8264fe5a344da9811ef5d6e_0/explore

Note: This is not about bringing in RECAP to be included in the J40 score itself; the goal is to collect datasets that we compare our score output to so that we understand how our score compares to existing indices or indicators. 

**Tasks**
- [ ] ETL RECAP score
- [ ] Modify comparison tool script to load RECAP data and compare it against current score
- [ ] Modify comparison tool script to add stats about the comparison with RECAP with the current score to the report generated at the end of the script

**Definition of "Done"**
- [ ] The J40 comparison tool can be run using the most recent version of our score
- [ ] The comparison tool can access the HUD RECAP data
- [ ] A report / data export can be generated that outputs the results of the comparison on a census tract level

Issue #198, opened 2021-06-21
As a developer or data scientist, I want to read documentation on how to get started with python notebooks in this repo so that I can play with the data and related code myself.


switzersc-usds:
**Description**
We need documentation on our python notebooks and data pipeline so that data scientists can read, use, and/or integrate with our notebooks and data.

**Solution**
Beginner-friendly documentation in this github repo, geared towards data scientists and developers.

**Describe alternatives you've considered**
No documentation, which would be a poor decision for empowering the community.

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**
- [ ] Documentation PR submitted, reviewed, and merged

switzersc-usds:
Please add documentation to the `score` directory, either in the README.md or in a new `docs` subdirectory.

switzersc-usds:
How can we make this more usable by and incorporate feedback from EJ data experts?

switzersc-usds:
Public usability and moving from data to conclusion as quickly as possible. Having point and click methods will make this more accessible than Jupyter notebooks. Usability vs usefulness of the data. The client will be very interactive but how can we improve this open source data analysis stage? What's the middle ground that makes this as usable as possible without getting into client roadmap development?

Ideas for usable components:
- R/Shiny dashboard with point and click tooling
- Simple mapbox/leaflet integration to see things and drag them around
- Add some of this to the Jupyter notebooks themselves, maybe some way to have a GUI on a notebook
- Make sure even the jupyter notebook portion is dockerized, because dealing with dependencies can cause headaches; clear documentation to get the notebook running

Ideas for use:
- Use some of these components in user feedback sessions with EJ experts to discuss and get feedback on the data

widal001:
Building on the suggestion from @clayton-aldern for making jupyter notebooks interactive. I've done some work with [Binder](https://mybinder.org/) which rebuilds your repo using a docker container and makes python notebooks within it executable. It can take 5+ minutes to rebuild and could be longer since we have quite a bit in this repo in addition to the iPython notebooks, but I believe it is an open source project -- so an alternative to Google if we wanted to avoid vendor lock-in.

Here's [a tutorial](https://github.com/alan-turing-institute/the-turing-way/blob/master/workshops/boost-research-reproducibility-binder/workshop-presentations/zero-to-binder-python.md) on how to leverage Binder for Python jupyter notebooks.

clayton-aldern:
[...]

PR #272, opened 2021-06-29
(WIP) Comparison tool refactor & ETL HUD RECAP

lucasmbrown-usds:
Completes #271 and #234. 

This PR refactors the comparison tool so that it can accept an arbitrary list of draft scores (at the census block group level) and a list of comparison indices (at the census tract level), and create a comparison report for all of the combinations of those. 

In the current pull request, the draft scores are "Score A", "Score B", "Score C, "Score D", and "Score E", and the comparison indices are CalEnviroScreen and HUD RECAP. 

So it produces a CSV, markdown file, and Word doc report for each of the combinations: Score A by CalEnviroScreen, Score A by HUD RECAP, Score B by CalEnviroScreen, Score B by HUD RECAP, etc. 

This should make it easier to add new comparison indices and generate a lot of useful information about them quickly. 

However, I worry that in making this code rather DRY, I've also made it rather inscrutable. I'm not sure anyone else will be able to interpret what's happening here. Plus, there are some rather complex functions defined inside of a Python notebook, without tests, which I'm not sure is a good idea. 

But this is a useful starting point, and if we need to vastly simplify we can also do that.


widal001:
@lucasmbrown-usds I like the direction this refactor moves the code in! After spending a bit more time reading through the various python notebooks and trying to trace the flow of data through them, I have a few questions about the decisions made so far and some thoughts/suggestions on options for proceeding with the refactor and the other issues related to the comparison tool. 

## Feedback

I'll try to separate my questions and thoughts into the following categories. I also have some notes on how to tie next steps on this into work on the other issues tagged with `comparison-tool` but I'll share those in the slack channel we have. 

- **Organization:** Notes on overall file structure as well as the division of work and analysis across notebooks and scripts
- **Documentation:** Thoughts on how to provide better insight and clarity into the overall ETL, calculation, and comparison process, in addition to explanation around key steps within that process. 
- **Implementation:** Questions or suggestions about the design or implementation approach for specific steps in the ETL, calculation, or comparison process.

### Organization

- **Highlights**
  - I really like the overall organization of this section of the codebase, specifically:
    - Separating out the ETL process for each data source
    - Combining them and calculating the composite scores in one step/notebook
    - Conducting all of the comparisons in another once all of the scores are calculated
  - I also appreciate the isolation of data in a separate directory that is git ignored and the accompanying utils that take care of the population/cleaning process
- **Questions/Suggestions**
  - Is there a reason that you've chosen relegate the renaming and data-type casting of the data sources to the calculation notebook as opposed to the corresponding ETL notebook for the corresponding data source? 
    - Given that there's a fair amount of consistency in performing those steps across data sources, I can understand the instinct to make it DRY and do all of that cleaning in a single dataframe after being merged.
    - That being said, I typically think of renaming, cleaning, and potentially even normalizing the data from each source as something typically completed by the ETL process.
    - Additionally I would imagine that relegating those steps to ETL might be helpful in enforcing a kind of "contract" with the ETL scripts that ensures the resulting output satisfies some key conditions. Examples contract conditions might include:
      - Every row of the output column should be reduced/aggregated to a level that matches the other data sources
      - All output columns will be used as-is in the calculation of scores (name, datatype, etc.) 
      - The value of output columns should share a common standard for normalization, thus making relative comparisons of source data possible even before being organized into indices during the scoring process.
  - Are all of the scores (i.e. `Score A`, `Score B`, `Score C`, etc.) candidates for the Justice40 score? If so, is there (or will there be) a single composite J40 score calculated somewhere as issue #276 and others seem to suggest?
    - Either way, I would probably recommend (eventually) refactoring these scores into separate functions once we pull them into a standalone script
    - Additionally it might be helpful to identify which data sources from which inputs to those various scores are being pulled to make the data provenance of our calculations a bit clearer, but that's getting into the next section on documentation.

### Documentation

- **Highlights**
  - The inline comments are super clear and very helpful to get a sense of what's happening at each step of the calculations
  - The naming of variables and functions, as well as the general layout and approach to the cleaning, calculation, and comparison do the bulk of the work in helping someone understand what's going on.
- **Question/Suggestions**
  - It would probably be super helpful to have either a verbal write up or even a basic mermaid diagram that helps people visualize the steps performed by each of the scripts, e.g. how they populate the data directories; how they are read in, merged, and tabulated into new indices; how the resulting comparison is made and where it is stored. 
  - Additionally if it wouldn't be too much work to maintain, it could be really helpful to create or at least link to a data dictionary of each data source in the `README.md` for this section. This could also highlight any transformations and renaming that are performed against the original data source before it is factored into the score in order to enhance data provenance.
  - Likewise it would be great to have a description/explanation of the data sources that go into the newly tabulated scores. EJScreen does this well when they provide an sort of short hand "intuitive" interpretation of their indices, like the "Excess Risk" section of this [explanatory page](https://www.epa.gov/ejscreen/environmental-justice-indexes-ejscreen)
  - I know that documentation was it's own issue (#198 ) so this is likely out of scope for this particular PR and issue

### Implementation
- **Highlights**
  - The code is already fairly DRY and the way you've implemented it creates a clear path to modularizing the code into functions into separate scripts.
  - The use of named tuples to standardize operations performed against each of the data sets was clever, and it clearly reveals the parallel steps in preparing the different types of fields for calculating the various scores. 
- **Questions/Suggestions**
  - While the named tuples approach is clever, it winds up nudging the implementation to treat transformations applied to the columns individually rather than as a collective dataframe, forcing the use of `for` loops instead of other iteration strategies like `df.apply()` or even vectorization.
    - One example is in conducting min-max normalization. Here's an example of how to do it for an entire dataframe, [using `sklearn`](https://stackoverflow.com/a/26415620/7338319) 
    - Additionally it encourages the calculations applied to each column to be stored as new columns in the same dataframe which requires appending a suffix to their name, instead of storing them in a separate dataframe with the original column names, e.g. `df_normalized` and `df_percentile`
    - The dataset probably isn't large enough to suffer a meaningful performance hit since there aren't a ton of columns we're iterating through, but it does make the code a bit more verbose and less "pandorable" (admittedly, that latter point is a bit subjective, and I don't always find idiomatic `pandas` code to be the most readable for non-data scientists)
  - I haven't looked at the code closely enough to see if this would have unintended consequences elsewhere, but it might be possible to set `GEOID_FIELD_NAME` as the index once all of the data sources have been merged, so you can avoid having to repeatedly call `if data_set.renamed_field != GEOID_FIELD_NAME`

## Recommended Next Steps

- Most of the comments/suggestions above are intended as general considerations for moving forward with the buildout of this tool, and not as blockers for merging this refactoring. In total, I think this PR offers a solid move toward generalization of the code base for other Scores and data sources.
- I would **_not_** recommend refactoring any further until we begin moving this code to separate scripts that are covered by unit and (potentially) integration tests, for the following reasons:
  - iPython notebooks are difficult to meaningfully version control and often result in conflicts if multiple people are contributing to the same file.
  - Without proper testing, it would be difficult to ensure that further refactoring doesn't break existing functionality, especially if we haven't verified the accuracy or reliability of the existing code
  - Breaking this code into functions or modules that are covered with tests also creates easier refactoring "bounties" for steps that can be isolated for contributors to work on without reworking the entire data pipeline concurrently.
  - It also allows us to enforce sufficient "contracts" between segments of this pipeline, and ensures that when someone refactors a portion of the pipeline they don't violate the assumed inputs for a later stage in the pipeline, without also refactoring that portion to change those assumptions.
- Let me know if any of my own assumptions about the code or suggestions based on those assumptions are off-base, there's a good chance I misinterpreted a crucial part of the pipeline when I was reading through the code! 
- I'll also post to the slack channel with suggestions for how I can plug in to help chip away at some of the suggestions above along with the other issues tagged with `help-wanted`

score/ipython/scoring_comparison.ipynb
```diff
@@ -27,375 +46,834 @@
...
+    "calenviroscreen_df = pd.read_csv(\n",
+    "    calenviroscreen_data_path, dtype={GEOID_TRACT_FIELD_NAME: \"string\"}\n",
     ")\n",
-    "file_contents = download.content\n",
-    "zip_file_path = TEMP_DATA_DIR\n",
-    "zip_file = open(zip_file_path / \"downloaded.zip\", \"wb\")\n",
-    "zip_file.write(file_contents)\n",
-    "zip_file.close()"
+    "\n",
+    "# Convert priority community field to a bool.\n",
```
switzersc-usds, line 143: It's inconsistent that it seems we're doing this conversion here in the scoring comparison notebook for calenviroscreen but we do it in the HUD RECAP ETL for the relevant field in the recap data.

esfoobar-usds:
I'm a bit confused about this PR -- are there two new ETL notebooks here? ETLs should be Python files from now following the patterns here: https://github.com/usds/justice40-tool/pull/260

esfoobar-usds (approved):
🍏 for merge. @lucasmbrown-usds @switzersc-usds any objections in the next 30 minutes please?

lucasmbrown-usds:
Lots of fantastic comments here. Tonight I'll create a ticket tracking all of them for improvements. For now, we're going to merge so @esfoobar-usds can rebase and rewrite his ETL classes PR, which will address some of them. Thanks!

PR #277, opened 2021-06-30
Attempt no. 2 to fix auto-generated `main` link

NatHillardUSDS:
One more attempt to fix merging to main, it appears it doesn't like curly brackets

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -30,11 +30,11 @@ jobs:
       # If this is just a PR branch, put content in a hash directory. Otherwise, put it in "main"
       # This variable is used by gatsby-config.js to set a prefixPath
       - name: Set DESTINATION_FOLDER for main
-        if: ${{ github.event.pull_request.merged == true }}
+        if: github.event.pull_request.merged == true
         run: |
           echo "DESTINATION_FOLDER=main" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


TomNUSDS (approved):
Yeah, this is the syntax I see more often.
(I did a search of github for similar code when I did the review the first time. I should have called out the `${{ }}` seemed non-standard. Sorry)

Approved!

NatHillardUSDS:
Yeah, their [docs](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-expression-in-an-if-conditional) call out the below, indicating it should be working : 
```When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression, unless the expression contains any operators. If the expression contains any operators, the expression must be contained within ${{ }} to explicitly mark it for evaluation. For more information about if conditionals, see "Workflow syntax for GitHub Actions."```

but in practice it doesn't seem to, and real-world usage is low. Here's hoping this fixes this!

switzersc-usds:
@NatHillardUSDS is this related to an existing issue we can connect it to?

NatHillardUSDS:
Related to part of https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/125

Issue #200, opened 2021-06-21
"Explore the Tool" page non-map modifications


switzersc-usds:
As discussed with dev+design:

- [x] Slim Alert about public beta and limited data resources
- [ ] ~Download Packet Card (non-functional just yet)~ Add when we implement the feature
- [x] Switch to light-mode
- [ ] ~Updated "Just Progress Communities" header and paragraph~ removing because of in progress name conversation

NatHillardUSDS:
Notes from dev-design sync earlier:
* Look into subheader under "Just Progress" - "A climate and economic justice screening tool". if not possible, fall back to image. We need this text to be present, though Nat had doubts whether we could directly support a subheader at this location

PR #279, opened 2021-07-01
Site content and style update for sprint2/3

TomNUSDS:
* add legend and styling for map page
* update header to latest style/content
* add contact page
* add methodology page
#199 and #200


TomNUSDS:
Opps. I did the map legend thinking it was part of the page and I should do it. But it looks like you did it as well?

NatHillardUSDS:
Oh no! Apologies for the overlap, indeed I think we both implemented it. I like your use of `dl` for this, but it's useful to have it in its own component as in my fix in case we do need to move it around - maybe if we merged my version then converted the items to dl/dts ?

NatHillardUSDS (approved):
Overall lgtm, but see comment about legend changes - maybe actually you could merge and I could bring changes over on my branch and resolve?

Issue #37, opened 2021-05-11
As an open source community member, I want to read open source process docs in my native language, so that I can understand the processes and more effectively contribute.


switzersc-usds:
**Description**
Juan Parras raised the great point in our kickoff call that we should be translating our process docs (not just our published website) into the languages of communities impacted by this work and whom we want to ensure are included in the open source processes we are building. 

**Solution**
Provide translations of our process documentation in prioritized languages. Spanish at a minimum, and we should identify others needed.

**Describe alternatives you've considered**

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**
Translations of the following docs: 

- [x] README
- [x] CONTRIBUTING
- [x] Code of Conduct
- [ ] Data Roadmap Process
- [ ] Locations in directory structure (i.e. /EN and /ES) created to store .md files
- [ ] Issue templates (e.g. have English and Spanish in the template)

katherinedm-usds:
Update that translation has been officially kicked off with the DOS, and I asked them if they could provide by COB this Friday June 18th, at least for the .md files. 

That means we'd need an /ES directory or some place to put these, @switzersc-usds could you please advise where these should go in the repo when ready? Not super urgent but seems like it would be good to get the translations up when we receive.

One additional note, I didn't have Data Roadmap Process translated yet.

switzersc-usds:
Expecting translations 6/22

katherinedm-usds:
[...]

PR #281, opened 2021-07-01
Add translations for repo documents

switzersc-usds:
This addresses #37. Note that these translations were requested a few weeks ago so do not incorporate some of the latest content updates. We need to come up with a process for keeping both sets of docs up to date. 

In terms of organizational structure, I did not create a separate `es` directory for these because I thought they should mirror where the EN versions live (and I didn't want to move these core docs into `en` and `es` directories in the root project folder because that would make them all hard to find). Once we have ES versions of other files such as ones in directories like `docs`, I think it will make sense to create `es` and `en` as subdirectories within those directories.


ginabeena (approved):
A mí, esto parece un gran trabajo de traducción y una buena contribución. Que guay 😛

Issue #316, opened 2021-07-09
Census block groups should highlight when selected


NatHillardUSDS:
**Describe the bug**
According to the latest spec, we should have a 5px highlighted region around a census block group when a user selects it. This is not currently happening.

**To Reproduce**
1. Open the map, click a block group
2. The popup appears, but the block group is not highlighted

**Expected behavior**
The region should highlight on click

**Screenshots**
![Screen Shot 2021-07-08 at 11.19.21 PM.png](https://images.zenhubusercontent.com/6091a54c180d4758f3eedc40/273fd5b0-4e5d-4b75-95e4-2be4d0f8854f)

**Desktop (please complete the following information):**
 All

**Smartphone (please complete the following information):**
N/A

NatHillardUSDS:
Spec here:
![Screen Shot 2021-07-08 at 11.20.30 PM.png](https://images.zenhubusercontent.com/6091a54c180d4758f3eedc40/3d2c6018-0855-429f-a316-77b863e2141c)

PR #317, opened 2021-07-09
Census block groups should highlight when selected

NatHillardUSDS:
Addresses #316 - 
* Adds 5px blue border on selection of feature
* Adds tests and test helpers

Left out for now:
* zooming to block group - there were some bugs with the default `fitBounds` implementation, especially alongside the inability to zoom past 11

### DIFF
--- a/client/cypress/support/commands.js
+++ b/client/cypress/support/commands.js
@@ -21,3 +21,8 @@ Cypress.Commands.add('getMapCanvas', () => {
   return cy.get('.maplibregl-canvas');
 });
 
+Cypress.Commands.add('waitForMapIdle', (map) => {
+  return new Cypress.Promise((resolve) => {
+    map.once('idle', resolve);
+  });
+});
--- a/client/.vscode/launch.json
+++ b/client/.vscode/launch.json
@@ -42,6 +42,18 @@
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen",
       "port": 9229
-    }
+    },
+    {
+      "type": "chrome",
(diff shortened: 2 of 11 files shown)


client/cypress/support/commands.js
```diff
@@ -21,3 +21,8 @@ Cypress.Commands.add('getMapCanvas', () => {
   return cy.get('.maplibregl-canvas');
 });
 
+Cypress.Commands.add('waitForMapIdle', (map) => {
+  return new Cypress.Promise((resolve) => {
```
TomNUSDS, line 25: It seems odd to return a Promise here?
Maybe it should be "wrapped".
` return cy.wrap(new Cypress.Promise((resolve) ...`

I'm not super familiar with Cypress.

If it compiles and works, probably not super important.

TomNUSDS (dismissed):
lgtm

client/cypress/support/commands.js
NatHillardUSDS: Cypress requires Promise-like objects as returns, as it waits for the results to be available - I believe `wrap` could be used by its parent if we needed it, but at this level we need to indicate that its async

TomNUSDS (approved):
Reapproved after merge

client/cypress/support/commands.js
TomNUSDS: I missed replying to this item, but it's more a curiosity versus and an actual issue. I was just expecting something like this if a promise was returned (note the `async`).

`Cypress.Commands.add('waitForMapIdle', async (map) => {`

PR #318, opened 2021-07-09
Allows the user to download CBG list

### LINKED ISSUE #111: As a user, I want to download the current version of the CBG list, so that I can see how all communities including my own stack up. (Front-end only)
**Description**
We need a way to share the full list of census block groups so that we can begin getting user feedback. 

Excel file should include at minimum: CBG ID, impact score, each indicator score, county, state, and YES/NO in the prioritization list. 

This refers to the text/visual list of identified communities (currently one and the same as CBG) at the census block group level. Community does not equal census block group, but we are using census block group as the smallest unit of identification in the first iteration of this tool because of the data available to us.

Open questions: 
- [x] What endpoint provides this list? Is this on geoplatform? -- S3
- [x] How are we versioning the list? Punting this to another convo, other than figuring out how to pull latest CSV

**Solution**
A version of the "Download packet" (packet = .zip containing an explainer PDF and the list file) design where there's a button for an Excel version of the packet, and a button for the .csv version. @KameronKerger to confirm and update design.

See [how_to_download.pdf](https://app.zenhub.com/files/358242767/6741be00-9cca-4343-a683-bbcb422fd80c/download) Kameron created to illustrate how downloading and unzipping the packet would work - it's straightforward, but still including to ensure it's clear.

**Tasks**
- [x] This data available as zipped CSV on our S3 bucket (ideally our prod resources are up, if not we can send the CSV manually to S3)
- [ ] Make sure code pulls most recent CSV - ID or hash? Use S3 Versioning?
- [ ] Create the button
- [ ] Dependent on decision for threshold (as a default we will do 10% pending decision)
- [ ] Tests!!!

**Definition of "Done"**
- [ ] User visits tool and downloads a zip file with a compressed CSV of identified communities based on the current iteration of the scoring methodology.

vim-usds:
closes #111 

Currently downloading this [csv](https://usds-geoplatform-justice40-website.s3.amazonaws.com/data/score/csv/usa.csv) as a placeholder for when the actual URL is determined.

### DIFF
--- a/client/src/components/downloadPacket.module.scss
+++ b/client/src/components/downloadPacket.module.scss
@@ -31,6 +31,16 @@ $primary-color: #112f4e;
         .downloadBoxButtonContainer {
           margin-top: 20px 0 0 0;
           align-self: center;
+
+          a {
+            text-decoration: none;
+          }
+
+          img {
+            // This should be a global css-filter value as it's generated
+            // from the primary color.
+            filter: invert(15%) sepia(9%) saturate(5666%) hue-rotate(175deg) brightness(96%) contrast(95%);
+          }
         }
 
         .downloadBoxButton{
--- a/client/src/components/downloadPacket.module.scss
+++ b/client/src/components/downloadPacket.module.scss
@@ -1,0 +1,49 @@
+$primary-color: #112f4e;
+
+.downloadBoxContainer {
+    flex: 0 0 350px;
+    // flex: 1;
+    // border: 1px solid red;
+
+    color: whitesmoke;
+    margin: auto;
+
+    .downloadBox {
+      background-color: $primary-color;
+      min-width: 400px;
+      border-radius: 6px 6px;
+
+      .downloadBoxTextBox {
+        // border: 2px solid green;
+        padding: 25px 25px;
+        display: flex;
+        flex-direction: column;
+        
+        .downloadBoxTitle {
+          font-weight: bold;
+          margin-bottom: 10px;
+        }
+
+        .downloadBoxText {
+          margin-bottom: 20px;
+        }
+
+        .downloadBoxButtonContainer {
+          margin-top: 20px 0 0 0;
+          align-self: center;
+        }
+
+        .downloadBoxButton{
+          background-color: white;
+          color: $primary-color;
+          display: flex;
+
+          .downloadPacketText {
+            padding-top: 4px;
+            padding-left: 5px;
+          }
+        }
+
+      }
+    }
+  }
--- a/client/src/components/downloadPacket.module.scss.d.ts
+++ b/client/src/components/downloadPacket.module.scss.d.ts
@@ -1,0 +1,19 @@
+declare namespace DownloadPacketModuleScssNamespace {
+    export interface IDownloadPacketModuleScss {
+        downloadBoxContainer: string;
+        downloadBox: string;
+        downloadBoxTextBox: string;
+        downloadBoxTitle: string;
+        downloadBoxText: string;
+        downloadBoxButtonContainer: string;
+        downloadBoxButton: string;
+        downloadPacketText: string;
+    }
+  }
+
+declare const DownloadPacketModuleScssModule: DownloadPacketModuleScssNamespace.IDownloadPacketModuleScss & {
+    /** WARNING: Only available when `css-loader` is used without `style-loader` or `mini-css-extract-plugin` */
+    locals: DownloadPacketModuleScssNamespace.IDownloadPacketModuleScss;
+  };
+
+export = DownloadPacketModuleScssModule;
--- a/client/src/components/downloadPacket.test.tsx
+++ b/client/src/components/downloadPacket.test.tsx
@@ -1,0 +1,12 @@
+import * as React from 'react';
+import {render, screen} from '@testing-library/react';
+import DownloadPacket from './downloadPacket';
+
+test('download packet component defined', () => {
+  render(<DownloadPacket />);
+
+  const downloadButton = screen.getByRole('button', {name: /download packet/i});
+
+  expect(downloadButton).toBeDefined();
+});
+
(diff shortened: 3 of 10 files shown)


client/src/components/J40Map.module.scss
```diff
@@ -30,10 +30,16 @@ $sidebar-color: #ffffff;
   line-height: 40px;
   position: absolute;
   left: 20px;
-  top: 300px;
+  top: 500px;
   z-index: 10;
 }
 
+@media screen and (max-width: 400px){
```
NatHillardUSDS, line 37: Good catch!

NatHillardUSDS: A side note, we may need an @media query for "phablet" as well - at 500px I was seeing it jump as well. I realize this bug was actually introduced with my latest fix, however, in using absolute positioning. We can / should revisit this later

client/src/components/downloadPacket.module.scss
```diff
@@ -0,0 +1,49 @@
+$primary-color: #112f4e;
+
+.downloadBoxContainer {
+    flex: 0 0 350px;
```
NatHillardUSDS, line 4: At some point we do have to break down to pixels, but I'd wonder if we could use percentages and/or column widths from USWDS? `content` may also work for `flex-basis` .

client/src/components/downloadPacket.module.scss
```diff
@@ -0,0 +1,49 @@
...
+    // flex: 1;
+    // border: 1px solid red;
+
+    color: whitesmoke;
+    margin: auto;
+
+    .downloadBox {
+      background-color: $primary-color;
+      min-width: 400px;
+      border-radius: 6px 6px;
```
NatHillardUSDS, line 14: This goes for all pixel values in this file actually - could these be ems or percentages to ensure [responsiveness](https://dev.to/j836/font-size-em-rem-pixel-or-percent-3f3p)? I realize this holds for all of our existing css files also but we should have been more diligent!

client/src/components/downloadPacket.module.scss
```diff
@@ -0,0 +1,49 @@
...
+
+      .downloadBoxTextBox {
+        // border: 2px solid green;
+        padding: 25px 25px;
+        display: flex;
```
NatHillardUSDS, line 19: ❤️  flexbox!

client/src/components/downloadPacket.module.scss.d.ts
```diff
@@ -0,0 +1,19 @@
+declare namespace DownloadPacketModuleScssNamespace {
```
NatHillardUSDS, line 1: Woo modules!

client/src/components/downloadPacket.test.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import {render, screen} from '@testing-library/react';
+import DownloadPacket from './downloadPacket';
+
+test('download packet component defined', () => {
```
NatHillardUSDS, line 5: good call adding jest tests also

client/src/components/downloadPacket.module.scss
NatHillardUSDS: May make sense to handle this as part of the mobile audit actually - better to merge as-is and make a followup ticket to nix all `px` across the site, we've got quite a few

NatHillardUSDS (dismissed):
LGTM! 🔥  first major change - jest + cypress, clean, readable code and styles. Overall awesome! A small comment about using `px` but we do this throughout the codebase; I think this one may benefit from a more general audit across the site next sprint

client/src/components/downloadPacket.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const DownloadPacket = () => {
+  const cbgFileURL = 'https://usds-geoplatform-justice40-website.s3.amazonaws.com/data/score/csv/usa.csv';
+  return (
+    <div className={styles.downloadBoxContainer}>
+      <div className={styles.downloadBox}>
+        <div className={styles.downloadBoxTextBox}>
+          <div className={styles.downloadBoxTitle}>Just Progress Packet</div>
+          <div className={styles.downloadBoxText}>This downloadable packet includes the list of Just Progress
+          prioritized communities (30,021 census block groups) and 18 datasets.
```
NatHillardUSDS, line 15: Ah, just saw these hard-coded values - maybe we could parameterize this somehow? We could have it pull from graphql potentially and store this as we build? This could be a followup ticket also

client/src/components/J40Map.module.scss
vim-usds: agreed, there's a part of me that wants to fix most of fixed px values

client/src/components/downloadPacket.module.scss
vim-usds: good point! I will look into this

client/src/components/downloadPacket.tsx
vim-usds: I was thinking this also. If the content was coming in from another source (CMS), we could have product/design edit it as well. Might be worth looking into another ticket for this.

switzersc-usds:
+1 to the idea of a general audit of `px` next sprint, probably as part of a look at general mobile responsiveness and breakpoints. Could this be part of https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/268 or shall we make a new ticket?

vim-usds:
> +1 to the idea of a general audit of `px` next sprint, probably as part of a look at general mobile responsiveness and breakpoints. Could this be part of https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/268 or shall we make a new ticket?

Sounds good to me!

NatHillardUSDS (approved):
LGTM !

NatHillardUSDS (dismissed):
Re-approving!

NatHillardUSDS (dismissed):

vim-usds:
Final review request please... merge conflicts are now resolved!

NatHillardUSDS (approved):

NatHillardUSDS:
Ah also can you make up the ticket for auditing px/em if you haven't already?

Issue #311, opened 2021-07-08
Tiles disappear after zoom 11


switzersc-usds:
**Describe the bug**
Map data disappears as you zoom in past zoom level 11. We should be using overzoom so theoretically zoom levels 12 and above should still work, but for some reason this is not working. Overzoom should be enabled by default - maybe we are not implementing overzoom correctly.

**To Reproduce**
1. Open the map page in the app
2. Zoom in past zoom level 11.99 (you can see the zoom level in the URL)
3. Tiles disappear.

**Expected behavior**
At step 3 above, data should still be visible.

**Screenshots**
![image](https://user-images.githubusercontent.com/54290681/124984901-c2a63f00-e007-11eb-8a28-04f1b2b8bd9a.png)

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser: Firefox
 - Version: 89

**Smartphone (please complete the following information):**

**Additional context**

NatHillardUSDS:
We should note that in our [style file](https://github.com/usds/justice40-tool/blob/main/client/src/data/mapStyle.tsx) we are setting minzoom and maxzoom , but even if these are removed, or the source maxzoom is set to be less than the layer maxzoom, we're still seeing that the tiles disappear at these higher levels

NatHillardUSDS:
Our current tippecanoe command is as follows:
```docker run --rm -it -v ${PWD}:/data mikegiddens/tippecanoe tippecanoe --force -Z 5 -z 11 -l tl_2010_bg -o /data/tl_2010_bg/mbtiles/tl_2010_42_bg10.mbtiles -l blocks /data/tl_2010_bg/geojson/tl_2010_42_bg10.json```

thus the stopping after 11 (the `-z 11` arg above) - that said, with overzoom this shouldn't be a problem, so I suspect something is up in our mapbox style config

mikelmaron:
@NatHillardUSDS  try setting maxzoom on the source to the maxzoom in the layer style and see if that helps. overzoom should still happen beyond where tiles are available

NatHillardUSDS:
Check this out here:
https://d2zjid6n5ja2pt.cloudfront.net/justice40-tool/be0256/en/cejst/?flags=mb

janelien:
The overlay disappears at zoom level 22, but I don't think that's a problem as most people will not need to zoom in that far. 

Looks great!

PR #324, opened 2021-07-12
[BUG] Tiles disappear after zoom 11

NatHillardUSDS:
Addresses #311 - tiles disappear after zoom level 11. By setting a max zoom for the source which matches the max zoom of the layer, we enable "overzoom" - more on this [here](https://docs.mapbox.com/help/glossary/overzoom/). Thanks to Mikel for this tip!


TomNUSDS (approved):
LGTM

PR #339, opened 2021-07-13
Parameterize zoom experiments

NatHillardUSDS:
With this change, our `flags` parameter takes an argument , and our map style is set via a method. 

Combining these, we can do:
http://localhost:8000/en/cejst?flags=low_tiles=tract_low,high_tiles=0629_demo to set a bucket explicitly for low tiles and high 

(this change looks bigger than it is - it converts the flags to a dict, still supporting the old `flag=1` syntax but replacing it with a dict with values like `{1:true}`, and it converts the map style to use the `low_tiles` and `high_tiles` params

### DIFF
--- a/client/src/components/J40Header.tsx
+++ b/client/src/components/J40Header.tsx
@@ -25,7 +25,7 @@ const J40Header = () => {
   const toggleMobileNav = (): void =>
     setMobileNavOpen((prevOpen) => !prevOpen);
 
-  const headerLinks = (flags: string[] | undefined) => {
+  const headerLinks = (flags: {[key: string] : any} | undefined) => {
     // static map of all possible menu items. Originally, it was all strings,
     // but we need to handle both onsite and offsite links.
     const menuData = new Map<string, JSX.Element>([
@@ -64,7 +64,7 @@ const J40Header = () => {
     // select which items from the above map to show, right now it's only two
     // possibilities so it's simple. Note: strings are used as react keys
     const menu =
-      flags?.includes('sprint3') ?
+      ('sprint3' in flags!) ?
         ['about', 'cejst', 'methodology', 'contact'] :
         ['about', 'cejst', 'methodology', 'contact'];
     // TODO: make feature flags flags work.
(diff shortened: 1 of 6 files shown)


client/src/components/J40Header.tsx
```diff
@@ -25,7 +25,7 @@ const J40Header = () => {
   const toggleMobileNav = (): void =>
     setMobileNavOpen((prevOpen) => !prevOpen);
 
-  const headerLinks = (flags: string[] | undefined) => {
+  const headerLinks = (flags: {[key: string] : any} | undefined) => {
```
TomNUSDS, line 28: So flags are now key:value pair? Seems reasonable.

client/src/contexts/FlagContext.test.tsx
```diff
@@ -6,18 +6,19 @@ describe('URL params are parsed and passed to children', () => {
   describe('when the URL has a "flags" parameter set', () => {
     // We artificially set the URL to localhost?flags=1,2,3
     beforeEach(() => {
-      window.history.pushState({}, 'Test Title', '/?flags=1,2,3');
+      window.history.pushState({}, 'Test Title', '/?flags=1,2,3,test=4');
```
TomNUSDS, line 9: nice

TomNUSDS (approved):
LGTM

client/src/components/J40Header.tsx
NatHillardUSDS: Yep - should arguably have been them all along - 
example is `?flags=low_tiles=NAME`

client/src/components/J40Header.tsx
```diff
@@ -64,7 +64,7 @@ const J40Header = () => {
     // select which items from the above map to show, right now it's only two
     // possibilities so it's simple. Note: strings are used as react keys
     const menu =
-      flags?.includes('sprint3') ?
+      ('sprint3' in flags!) ?
```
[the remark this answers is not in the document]
NatHillardUSDS: yeah, in this case we can assume they exist - in theory they can be undefined but haven't seen in practice

Issue #188, opened 2021-06-18
As a developer, I want to add state, state abbrv, and county to my score CSV to help with data analysis and so that it can be accessible via the UI.


switzersc-usds:
**Description**
In order to make the score CSV more useful during analysis and development, we want to add the state, state abbreviation, and county names. We know we also want this data in the CSV that we make available for download in the UI, which may pull from the score CSV S3 bucket directly or be generated from the geoJSON.

**Solution**
 Determine the state, state abbrv, and county from the census block group ID and add these three columns to the score CSV using pandas.

**Describe alternatives you've considered**

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**
- [ ] The score CSV that lives (or will eventually live) on S3 has state, state abbreviation, and county name for every census block group row.

esfoobar-usds:
Just to be super clear -- I would add 3 columns: state_name, state_abbrev and county_name to the score CSV

switzersc-usds:
@esfoobar-usds sounds good. Card updated.

switzersc-usds:
@esfoobar-usds is this data available in the Census TIGER files we're combing the CSV with for the geoJSON?

esfoobar-usds:
Not sure, will start the research today.

NatHillardUSDS:
As part of https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/132 , I found these:  https://www.census.gov/programs-surveys/geography/technical-documentation/records-layout/nlt-record-layouts.html  - these tables translate FIPS state/county codes to human-readable values. In particular this page from 2010 can help -- though its per-state: https://www.census.gov/geographies/reference-files/time-series/geo/name-lookup-tables.2010.html .

esfoobar-usds:
I couldn't see one that maps to county name. I downloaded Alabama.

Here's the list of files:
Geographic Area	<entity>	Field Layout
Congressional districts	CD	STATEFP|DISTRICT|NAME|NAMELSAD
State legislative districts - upper	SLDU	STATEFP|DISTRICT|NAMENAMELSAD
State legislative districts - lower	SLDL	STATEFP|DISTRICT|NAME|NAMELSAD
Voting districts	VTD	STATEFP|COUNTYFP|DISTRICT|NAME|NAMELSAD
Elementary school districts	SDELM	STATEFP|DISTRICT|NAME|NAMELSAD
[...]

PR #333, opened 2021-07-12
Add notebook to create lower zoom level data 

NatHillardUSDS:
Fixes #201 - Adding notebook representing how to dissolve values to lower zoom levels. To be productionized with #332

High-level strategy:
1. Combine state geojson files into a data frame
2. dissolve these by tract
3. dissolve these by "bucket", an arbitrarily-chosen number of collapsible delineations, with a set of MULTIPOLYGONs
4. split up MULTIPOLYGONs into separate POLYGONs

This will need to be tweaked further in order to represent things correctly at lower levels.

We would put this into the FE with #201

Issue #302, opened 2021-07-07
As a data developer, I want to limit the size of the mbtiles database, so that the frontend can have a performant experience


NatHillardUSDS:
**Description**
There are a few optimizations that have been suggested for the mbtiles database from geoplatform to improve performance. We should adopt these. 

**Solution**
- Let's do the following optimizations to shrink tile sizes.

**Describe alternatives you've considered**
- By default, map loading is slow with the full set of 95+ properties

**Links to user research or other resources**
- From [this link](https://docs.mapbox.com/help/troubleshooting/mapbox-gl-js-performance/):  
```render time = constant time + [ number of sources * per source time ] + [ number of layers * per layer time ] + [ number of vertices * per vertex time ]```
Thus if we can shrink the number of properties per tile we can overall lower load times.

**Tasks**
- [ ] Limit the mbtiles db properties to the values shown on the spreadsheet for the popup
- [ ] Limit floating point values to two significant figures (maybe using strings? we can convert types on the frontend)
- [ ] Make “official” property names shorter (we'll need an official long name to show in the UI)

**Definition of "Done"**
The mbtiles database has the above modifications in place

NatHillardUSDS:
We've discussed a bit the possibility of having multiple tilesets with different information, having one that only has the score, for instance, and another for other details, which is queried when you click "more" in the detail view, for instance. This may not be necessary if there is only a small number of properties that we need to care about in the popup (it sounds like this may be the case)

esfoobar-usds:
@NatHillardUSDS what are the columns that will be shown in the popup? Is there a document where I can check?

esfoobar-usds:
@NatHillardUSDS FYI, I created a list of the columns we want baked into the tiles [here](https://github.com/usds/justice40-tool/blob/7b37f50b2ae4dd4c22589acbd2387d449a252f57/score/etl/score/etl_score_post.py#L26-L33)

PR #347, opened 2021-07-13
County Names for Score #188

esfoobar-usds:
Work for https://github.com/usds/justice40-tool/issues/188


esfoobar-usds:
@lucasmbrown-usds ready for review.

score/etl/score/etl_county_state_score.py
```diff
@@ -0,0 +1,89 @@
...
+
+from etl.base import ExtractTransformLoad
+from utils import get_module_logger
+
+logger = get_module_logger(__name__)
+
+
+class CountyStateScoreETL(ExtractTransformLoad):
+    def __init__(self):
+        self.CENSUS_COUNTIES_ZIP_URL = "https://www2.census.gov/geo/docs/maps-data/data/gazetteer/2020_Gazetteer/2020_Gaz_counties_national.zip"
```
NatHillardUSDS, line 11: We'll want to use the 2010 version of this to match the census data we're currently pulling - that's linked to from here: https://www.census.gov/geographies/reference-files/time-series/geo/gazetteer-files.2010.html

esfoobar-usds:
TODO (after meeting with Engineering)

- [ ] - Rename ETL to Score Post Processing - take care of different score output files
- [ ] - Comb through and use 2010 info for everything
- [ ] - Add optimizations mentioned on #302 
- [ ] - After this PR tackle #333 
- [ ] - Clean up per state generation of files (we now use `tilejoin` with the whole us)
- [ ] - Create a ticket to rewrite the local version of mbtiles/mvt generate py so that the Lambda function uses it and users can generate them locally without relying on the GeoPlatform provided US mbtile

NatHillardUSDS (changes requested):
Overall looks great! Only flag would be to use the 2010 version of the data

score/etl/score/etl_county_state_score.py
esfoobar-usds: Updated

NatHillardUSDS (approved):
lgtm!
(Ah, sorry, realized just now that there's the above checklist still)

PR #354, opened 2021-07-14
Fixed es.json

TomNUSDS:
- converted remaining english to spanish in file so as to check more html locations.
- the `formatjs compile` for Spanish was failing converted back to english for now. build does the correct thing, so it's more just to keep things correct in source.


nathillard (dismissed):
Hm, this works for now; I would think it may make sense to leave this file blank for now? We risk misrepresenting things with our google translate version. 

I would also see if you could add this actually? https://formatjs.io/docs/tooling/linter/ - this way we catch missing translations with our linter itself.

TomNUSDS:
> Hm, this works for now; I would think it may make sense to leave this file blank for now? We risk misrepresenting things with our google translate version.

Yeah, hopefully we wouldn't hand them our es.json but just feed the en.json into the translation tool. 

It's likely all text on the site will be replaced in the next batch of UI changes, so I'll just empty the es.json with those.

Random thought: maybe we should have some mock language code for Ipsum Lorem? Would be fun to have some iplo.json in the language file that jest could sanity check going to /iplo/ replaces all the text correctly.

> I would also see if you could add this actually? https://formatjs.io/docs/tooling/linter/ - this way we catch missing translations with our linter itself.

Good thinking. I'll check into it.

TomNUSDS:
Added `eslint-plugin-formatjs` checking. Think typescript does most of the enforcing we need, but nice to have anyways.

nathillard (approved):
LGTM - we should make sure to not go live with the Google translate versions

NatHillardUSDS (approved):
Approving with correct account this time

Issue #366, opened 2021-07-15
Each transpiled file includes its own copy of styles instead of a static reference


TomNUSDS:
**Describe the bug**
There is something wrong with our build. CSS styles should be referenced in an external css file not embedded in each file. This is making the file very large.

**To Reproduce**
1. run `gatsby clean`
2. run `gatsby build`
3. optional run `prettier --write ./public/` to reformat "compressed" content.
4. Open any of the `index.html` files
Result: each file has **21k lines** of css in it.

**Expected behavior**
The css should be referenced via a external file.  eg. `<link type="text/css" rel="stylesheet" href="...">`

TomNUSDS:
https://github.com/gatsbyjs/gatsby/issues/1526#issuecomment-433347822

katherinedm-usds:
hi @TomNUSDS! I see this is in Design column. Not sure from this bug description and history if this specifically needs Design input right now, and if so what that input is?

TomNUSDS:
This was purely an dev issue. It was just a big inefficient in generating the html/css.

switzersc-usds:
Fixed!

PR #369, opened 2021-07-15
Fix inlined css in every single page.

TomNUSDS:
Fix for issue #366

Created a gatsby-ssr.js to manually remove the css styles embedded in the pages.

SSR means "server side rendering". This fix is from a long github discussion about how broken including all css inline in the file is. 

The patch came from here: https://github.com/gatsbyjs/gatsby/issues/1526#issuecomment-433347822 

But honestly, I can't believe there is no flag to control this behavior and the issue was marked "Closed" by the Gatsby folks the way it works now. Especially since newer [CSP guidelines](https://content-security-policy.com/unsafe-inline/) say all css should be in a separate file.

The outputted file size for main `index.html`

```
Before:   462,099 bytes
After:      3,191 bytes
```
This is for _every single statically generated page_.

I tested locally but we should watch for production build issues.

### DIFF
--- a/client/gatsby-ssr.js
+++ b/client/gatsby-ssr.js
@@ -5,13 +5,13 @@
   `ssr` means server-side rendering.
   */
 export const onPreRenderHTML = ({getHeadComponents}) => {
-  if (process.env.NODE_ENV !== 'production') {
+  if (process.env.NODE_ENV !== 'production') { // ONLY run in production
     return;
   }
 
   getHeadComponents().forEach((el) => {
     // Remove inline css. https://github.com/gatsbyjs/gatsby/issues/1526
-    if (el.type === 'style') {
+    if (el.type === 'style' && el.props['data-href']) {
       el.type = 'link';
       el.props['href'] = el.props['data-href'];
       el.props['rel'] = 'stylesheet';
--- a/client/gatsby-ssr.js
+++ b/client/gatsby-ssr.js
@@ -1,0 +1,25 @@
+
+/* this is a hack to fix `gatsby-plugin-sass` from including the
+  css in every single file (which is > 120k lines).
+  see: https://github.com/gatsbyjs/gatsby/issues/1526
+  `ssr` means server-side rendering.
+  */
+export const onPreRenderHTML = ({getHeadComponents}) => {
+  if (process.env.NODE_ENV !== 'production') {
+    return;
+  }
+
+  getHeadComponents().forEach((el) => {
+    // Remove inline css. https://github.com/gatsbyjs/gatsby/issues/1526
+    if (el.type === 'style') {
+      el.type = 'link';
+      el.props['href'] = el.props['data-href'];
+      el.props['rel'] = 'stylesheet';
+      el.props['type'] = 'text/css';
+
+      delete el.props['data-href'];
+      delete el.props['dangerouslySetInnerHTML'];
+      delete el.props['children'];
+    }
+  });
+};


client/gatsby-ssr.js
```diff
@@ -0,0 +1,25 @@
...
+  `ssr` means server-side rendering.
+  */
+export const onPreRenderHTML = ({getHeadComponents}) => {
+  if (process.env.NODE_ENV !== 'production') {
+    return;
+  }
+
+  getHeadComponents().forEach((el) => {
+    // Remove inline css. https://github.com/gatsbyjs/gatsby/issues/1526
+    if (el.type === 'style') {
```
NatHillardUSDS, line 14: This commenter needed to check on existence of `props['data-href']` - seems we might also want to do this?

client/gatsby-ssr.js
```diff
@@ -0,0 +1,25 @@
+
+/* this is a hack to fix `gatsby-plugin-sass` from including the
+  css in every single file (which is > 120k lines).
+  see: https://github.com/gatsbyjs/gatsby/issues/1526
+  `ssr` means server-side rendering.
+  */
+export const onPreRenderHTML = ({getHeadComponents}) => {
+  if (process.env.NODE_ENV !== 'production') {
```
NatHillardUSDS, line 8: so if I'm reading this correctly, we only do this on dev builds? Is the intention to speed up local development builds?  Not sure I'm following the larger reasoning

NatHillardUSDS (changes requested):
Hm, interesting catch, but if it's not running on our prod builds I'm not sure I see the reasoning, though I may be misreading this / misinterpreting the purpose. 

In any case it does seem like one of the commenters had to check for a particular property's existence so marking as request changes at least for that; otherwise I'd be curious to tie this back to real-world repercussions also if we can

vim-usds (approved):
What a creative fix! Is it safe to say that the styles were duplicated (in-line) and external and this is now just pointing to the external stylesheet?

client/gatsby-ssr.js
TomNUSDS: Excellent point. It always makes sense to do it. I could see maybe NOT doing it on dev builds?

TomNUSDS:
> What a creative fix! Is it safe to say that the styles were duplicated (in-line) and external and this is now just pointing to the external stylesheet?

Yes. The gatsby build processes the files to produce static html output for deploying web pages. This hooks into it to fix an odd behavior in the gatsby sass processing plugin.

TomNUSDS:
> Hm, interesting catch, but if it's not running on our prod builds I'm not sure I see the reasoning, though I may be misreading this / misinterpreting the purpose.
> 
> In any case it does seem like one of the commenters had to check for a particular property's existence so marking as request changes at least for that; otherwise I'd be curious to tie this back to real-world repercussions also if we can

This gatsby sass plugin hasn't been worked on in a long time and I'm confused why they closed this issue without fixing it. The _real_ fix is to add a flag to the plugin to control this behavior, not have everyone stub their toe on it and then have to add custom code.

client/gatsby-ssr.js
TomNUSDS: I think this code says if it's not production, then exit. Meaning it only runs on production. This makes sense if dev has sourcemap files issues I guess?

NatHillardUSDS: oh no, I missed the `!`! ha, good to know

NatHillardUSDS:
Apologies for my misreading, clearly I have Friday Brain. Would be good to go with that extra check I think

client/gatsby-ssr.js
TomNUSDS: Heh. When you said, "why not production?", I looked at the same code and was like "YEAH, Why is this backwards!!1!".. Then I really looked at it a second time and saw it was prod only... so, I think it's written oddly and a least deserves a comment.

vim-usds: I totally missed the `not` there

NatHillardUSDS (approved):
LGTM, and good catch!

PR #375, opened 2021-07-16
Folder re-arrangement in prep for 361

esfoobar-usds:


esfoobar-usds:
Re-starting again, need to reset data folders

Issue #383, opened 2021-07-19
As a user, I want to see the cumulative impact score and associated info for communities that are not prioritized, so that I can learn more about how communities are scored


katherinedm-usds:
**Description**
The current version of the tool does not let you select non-prioritized or threshold CBGs to see what their cumulative index score percentile, non-prioritized Categorization, indicators and associated percentiles for those, etc. 

**Solution**
For non-prioritized CBGs, enabled CBG selection and side panel information similar to prioritized and threshold communities

**Describe alternatives you've considered**

**Links to user research or other resources**
- @janelien confirm design, so it's clear what the Categorization should look like on the side panel for non-prioritized communities, etc.

**Tasks**
 - [ ] @janelien confirm design
 - [ ] Dev enable non-prioritized CBG selection
 - [ ] Dev implement valid score data to be populated in side panel matching design

**Definition of "Done"**
Valid cumulative index score percentile, categorization, CBG ID, indicators and percentiles shown for all non-threshold CBGs

NatHillardUSDS:
Pushed the first part of this change just now , which allows you to interact with non-prioritized communities:
You can test this here: https://d2zjid6n5ja2pt.cloudfront.net/justice40-tool/ccb080/en/cejst/

We still need to add the side panel as well to complete the above requirements, but this is the first part

janelien:
Design is under Sprint 4 so it's next to the other side panel designs but titled:
J40_Sprint 5_map_not prioritized

janelien:
Met with Vim and discussed a few small edits. 
- Changing from title case to sentence case (for Housing burden and Linguistic isolation)
- Unemployment to be 'Unemployment rate'
- Changing mobile breakpoint to 414px. It's currently at 400px, but 400px doesn't work for iPhone 11s. This would be a temporary fix with a more permanent fix being tackled in the mobile readiness ticket. 

We looked at breakpoints and 800px would be good for pretty much all tablets/mobile devices. Tablet view would benefit from being the same as the mobile view. 

Overall, it's looking great!

janelien:
A bit more of what we looked at for reference:

https://www.browserstack.com/guide/ideal-screen-sizes-for-responsive-design

According to the Worldwide Screen Resolution Stats (Jan 2020 – Jan 2021), the most commonly used resolutions across mobile, desktop, and tablet are:

1920×1080 (8.89%)
1366×768 (8.44%)
360×640 (7.28%)
414×896 (4.58%)
[...]

PR #386, opened 2021-07-19
Show score for non-prioritized communities 

NatHillardUSDS:
Fixes part of #383 : As a user, I want to see the cumulative impact score and associated info for communities that are not prioritized, so that I can learn more about how communities are scored . 

This fix makes non-prioritized tiles interactive by removing the filter we had previously applied against them. It also alters the highlight layer to display at layer 8 to help clarify further the boundaries of the non-prioritized regions.

Filter is still applied at zoom levels 0-7 because otherwise map would turn white at lower zooms


TomNUSDS (approved):
LGTM

PR #388, opened 2021-07-20
Fix #353 - site icon

### LINKED ISSUE #353: Change favicon / site icon

TomNUSDS:
Fixes #353 
site not being at the "topmost level" (e.g. there's a `/path`) makes the approach of having the favicon in a known, static location break (`https://[tld]/favicon.ico`).

Setting the icon via the plugin should make it work. It should be here anyways.
We'll still need to add the SEO plugin as part of the accessibility fixes (setting page meta tags in a i18n way)


NatHillardUSDS (approved):
lgtm

PR #398, opened 2021-07-22
adding tree equity score to the data pipeline 

rohitmusti:
This adds the Tree Equity Score and its underlying data to the data download process! If this PR gets pulled in, I'll add it to the scoring comparison pipeline as well.

links to issue #276 it addresses the first task, with its approval, I'll address the other tasks, I was hoping to move in small PRs to cause minimal disruption to the overall workflow


rohitmusti:
find out more about tree equity score [here](https://treeequityscore.org/map/) and explore it in our [map](https://treeequityscore.org/map/)

rohitmusti:
links to issue #276

data/data-pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -0,0 +1,42 @@
...
+        tes_state_dfs = []
+        for state in self.states:
+            tes_state_dfs.append(gpd.read_file(f"{self.TMP_PATH}/{state}/{state}.shp"))
+        self.df = gpd.GeoDataFrame(pd.concat(tes_state_dfs), crs=tes_state_dfs[0].crs)
+
+    def load(self) -> None:
+        logger.info(f"Saving Tree Equity Score GeoJSON")
+        # write nationwide csv
+        self.CSV_PATH.mkdir(parents=True, exist_ok=True)
+        self.df.to_file(self.CSV_PATH / "tes_conus.geojson", driver='GeoJSON')
```
switzersc-usds, line 42: Discussed this in and after office hours -- summary and conclusion here:

The other ETL classes' `load` functions output a CSV rather than geoJSON. @rohitmusti decided to output geoJSON here because the geometries might be useful in the current or future comparison tool, and it won't be difficult to use the geojson in the current notebook. @esfoobar-usds and I chatted about this and we think that makes sense, so let's continue with geoJSON here and handle the merge in the comparison notebook in the future PR.

switzersc-usds (changes requested):
This looks great! Thanks so much for the contribution. The only issue is that the additional `geopandas` in the `pyproject.toml` is a duplicate and therefore raises an error when the `poetry` command is run. If you could just remove that please this will be good to go! :)

rohitmusti:
@switzersc-usds just resolved the extra dependency! thank you for the great PR experience

switzersc-usds (approved):
🎉 whoo! thanks! 🎉

Issue #356, opened 2021-07-14
Accessibility report fixes


TomNUSDS:
# Errors section
- [ ] Link to github is incorrect `https://github.com/usds/justice40-tool/crisis-at-home-and-abroad/` should be `https://github.com/usds/justice40-tool/`
- [ ] report tool doesn't have the following words in it's dictionary and thinks they are typos: datasets dataset github. Probably should add to the tool?

- [ ] `/static/sourcesanspro-italic-webfont.woff2`
This is a USWDS referenced font that I don't think is used. It is referenced in the css. May indicate a build issue or may just be unused and tree shaken removed by build.
```css
@font-face {
  font-display: fallback;
  font-family: Source Sans Pro Web;
  font-style: italic;
  font-weight: 400;
```

# Accessibility Section

## Level A
- [ ] duplicate id `main-content` in same page
- [ ] duplicate title across multiple pages `Justice40` Each page should have a unique title
- [ ] `<main>` appears in the `layout.tsx` and some pages like cejst.txt and 404.tsx
- [ ] Missing `lang` attribute for English pages. Note: `Helmet` is setting but doing so in the J40Header. Maybe move this up to the `layout.tsx`?

## Level AA
- [ ] If you set any of the colors on the body or a elements you must set all of them. `background: rgb(239,246,251)` aka `#eff3fb` This is the background color of our `<aside>` Not sure th fix. Don't have background colors? There may be foreground elements that are white `rgb(255,255,255)` somewhere here?
<details><summary>See suggestion</summary>
<p>

> In HTML the color attributes are text, bgcolor, link, alink and vlink. In CSS the attributes are color and background-color. Some users have browser defaults set to white text on a black background, so setting one color without setting the others can result in black text on a black background. High contrast mode on Windows turns off background images so you must set a `background-color` in addition to a `background-image`.
</p>
</details>

- [ ] Add a sitemap (there's a gatsby plugin)

- [ ] The CSS outline or border style on this element makes it difficult or impossible to see the dotted link focus outline. Think this is just something in the USWDS css file and "how it's done"?
<details><summary>See suggestion</summary>
<p>

> Change the style to avoid obscuring the focus outline around focusable elements.
> Using a border or outline style that obscures the focus ring causes problems for keyboard-only users in:

> Chrome (obscures the focus indicator on links, buttons, dropdowns and range controls)
> Firefox (obscures the focus indicator on links)
> Internet Explorer (obscures the focus indicator on links, buttons and range controls)
[...]

PR #403, opened 2021-07-23
Cypress tests are failing locally

### LINKED ISSUE #373: Cypress tests are failing locally
**Describe the bug**
Due to recent changes, our cypress tests are failing.

**To Reproduce**
1. Checkout main at hash e066f459
2. Run `npm start` in one window, then `npm run cy:run` in another
3. See attached output

[https://app.zenhub.com/files/358242767/f424ba15-b8b2-44a6-874f-45b7a26f65ce/download](https://app.zenhub.com/files/358242767/f424ba15-b8b2-44a6-874f-45b7a26f65ce/download)

**Expected behavior**
Our tests should pass

**Additional context**
Due to #125 we had not caught this before, as cypress tests were failing

NatHillardUSDS:
Fixes #373 - 
* Adding debug instructions for cypress tests
* Updating translation strings to get translation tests to pass
* Updating class names to get lat/lng tests to pass
* Localizing focus buttons on the way to adding proper area-labels for each button
* Removing buggy highlight test for now for map tests
* Setting accessibility bar to be "critical", to be updated later as part of accessibility audit #356

### DIFF
--- a/client/cypress/e2e/language.spec.js
+++ b/client/cypress/e2e/language.spec.js
@@ -1,11 +1,14 @@
 // / <reference types="Cypress" />
 
 describe('Translation Test', () => {
-  it('Checks that locales have correct content', () => {
+  it('Sets default language to /en and redirects', () => {
     cy.visit('http://localhost:8000');
     cy.url().should('include', '/en/');
-    cy.get('header').contains('Justice40');
+    cy.get('h1').contains('About Justice40');
+  });
+
+  it('Sets page content to spanish when visiting Spanish URL', () => {
     cy.visit('http://localhost:8000/es');
-    cy.get('header').contains('Justicia40');
+    cy.get('h1').contains('Acerca de Justice40');
   });
 });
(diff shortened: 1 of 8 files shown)


client/cypress/e2e/language.spec.js
```diff
@@ -1,11 +1,14 @@
 // / <reference types="Cypress" />
 
 describe('Translation Test', () => {
-  it('Checks that locales have correct content', () => {
+  it('Sets default language to /en and redirects', () => {
```
TomNUSDS, line 4: Should we start using Cypress over Jest? I'm hitting odd behavior around Intl with Jest while trying to write tests.

TomNUSDS (approved):
LGTM!

client/cypress/e2e/language.spec.js
NatHillardUSDS: We should continue to use both, but for different purposes - Cypress for full end-to-end integration tests and feature tests, and jest for unit tests and snapshot tests (we've been working toward this based on internal docs, but we should document this formally in this repo!). You should file the jest issue as a bug and we can look into it

PR #406, opened 2021-07-24
adds map side panel

### LINKED ISSUE #288: As an EVCM, I want to easily access CBG detail information on both desktop and mobile, so that I don't have to scroll forever. (Implement sidebar)
**Description**
Our current popup-based CBG detail view takes up a lot of screen space, especially on mobile. We should consider moving to a sidebar-based map as seen in sites like [Tree Equity Score](https://treeequityscore.org/map/#10.5/33.6607/-112.1231)

**Solution**
We should move the content that is currently in a popup to a sidebar instead, and move that sidebar to below the map on mobile

Support: Desktop chrome, firefox, safari, and edge; Mobile chrome and safari. Assume 400px breakpoint.

If one of the above browser reqs becomes a blocker, let's discuss at the time. 

- The current popup is large and intrusive, and doesn't look good on mobile

- NEEDED: Design spec

**Tasks**
- [x] Work with design on a spec to show this information
- [ ] Be sure to show county!
- [x] Move popup content into the sidebar instead
- [x] Ensure sidebar is moved below the map on mobile

**Definition of "Done"**
- [x] CBG detail content is visible in a sidebar on desktop
- [x] CBG detail content is visible below the map on mobile

vim-usds:
Hello! 

Big PR here. Ended up being much more deeper than I had originally estimated. 😅 

[Deployed PR URL
](http://usds-geoplatform-justice40-website.s3-website-us-east-1.amazonaws.com/justice40-tool/b38036/en/cejst/#3/32.47/-86.5)

## What this PR should do:
1. Closes #288 
2. Render the side panel in both desktop and mobile
3. Full-screen flag should also work and replaces popUp contents and styling with the contents and styling of side panel. It's a little buggy when closing from full-screen
4. The data for the side panel is being filtered on the client side as we wait for the server data to be ready
5. Unit tests test basic rendering
6. Cypress tests test 3 devices (desktop, iPhone6 and SGS10)
7. Added a `*Utils.module.scss` to share styles among sass modules to keep it DRY

### Bugs
"Locally everything works" 🤣   but....
1. (pre-existing) When viewing the app on mobile (via dev console) on the deployed PR URL:
- The map isn't resizing correctly (see comment)
- When clicking on a region on the map, the app throws an an `error`:

```
assert.js?aa7a:3 Uncaught Error: @math.gl/web-mercator: assertion failed.
```
Some more info [here](https://github.com/visgl/react-map-gl/issues/1237) needs more investigating.

I went back to a deployment prior to this feature and it seems to be happening before this PR. 🤔  This error does NOT happen locally. Makes me think it's the build output?

2. (pre-existing?) The console shows a `warning` when the app first starts up

```
Warning: Cannot update a component (`J40Map`) while rendering a different component (`ForwardRef`). To locate the bad setState() call inside `ForwardRef`
```
If we're all seeing it, then maybe we can let it go for now? I think it's due to the pop-up logic (maybe) 🤔 
This warning IS happening locally as well. 

### Little things
1. Modify cypress yaml cross browser testing
2. One unit test is weirdly not rendering (not a show stopper, see comment below)
3. One JSdoc requirement is still hanging me up (not a show stopper, see comment below)

## Questions
1. Should the string literals be moved out to `en.json` and `es.json`? 🤔 

## Final Thoughts
Sorry for the super long PR. I want to get better at breaking these up.

Looking forward to any and all feedback! Have a great weekend!


client/cypress/e2e/mapSidePanel.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    cy.get('div[class*="score"]').should('be.visible');
+    cy.get('div[class*="topRowSubTitle"]').should('be.visible');
+    cy.get('li[class*="indicatorBox"]').should('be.visible');
+    cy.get('li[class*="indicatorBox"]')
+    // currently the padding-top = .5rem => 8px
+        .each((indicator) => cy.wrap(indicator).should('have.css', 'padding-top', '8px'));
+  });
+});
+
+// Todo VS: Figure out why this forEach block doesn't work:
```
vim-usds, line 83: This is basically just trying to wrap 3 individual tests into a `.foreach`? Need another pair of eyes on it. The non-DRY way works

client/src/components/mapSidePanel.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  };
+  const selectedFeatureId = 345;
+
+  render(<MapSidePanel
+    className={'J40Map-module--mapSidePanel--8Ap7p'}
+    detailViewData={detailViewData}
+    selectedFeatureId={selectedFeatureId}
+  />);
+
+  // Todo VS: figure out why this component doesn't render?
```
vim-usds, line 44: This makes me question whether or not we always need unit tests. I'm not sure, because the cypress tests covers this case. 

I'm sure the "mock "data is in error somehow which may not be allowing the component to render.

client/src/components/mapSidePanel.tsx
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable require-jsdoc */
+// Todo VS: what is a better way around this jsdoc disabling?
```
vim-usds, line 2: Would like to review how to remove this `es-lint-disable`. Sat down with Nat on this earlier, but still scratching my head...

client/src/components/J40Map.tsx
```diff
@@ -173,7 +184,7 @@ const J40Map = () => {
         maxZoom={constants.GLOBAL_MAX_ZOOM}
         mapOptions={{hash: true}}
         width="100%"
-        height="52vw"
+        height={constants.isMobile && '90vw' || '44.8vw'}
```
vim-usds, line 187: This might be causing the error of where it's not resizing on mobile. One idea is to create an entirely new instance of the ReactMapGL component and return if mobile is true.

TomNUSDS (dismissed):
Looks great. Love the tests. Some minor comments.

client/src/components/areaDetail.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+  render(<AreaDetail properties={properties}/>);
+
+  it('checks if various text fields are visible', () => {
+    expect(screen.getByText('Cumulative Index Score')).toBeVisible();
+    expect(screen.getByText('Categorization')).toBeVisible();
+    expect(screen.getByText('Poverty')).toBeVisible();
+    expect(screen.getByText('Education')).toBeVisible();
+    expect(screen.getByText('Linguistic Isolation')).toBeVisible();
+    expect(screen.getByText('Housing Burden')).toBeVisible();
+  });
```
NatHillardUSDS, line 27: Nice adding tests here! Maybe also add a snapshot to ensure right sizing

client/src/components/areaDetail.tsx
```diff
@@ -2,91 +2,126 @@ import * as React from 'react';
...
 
+export const readablePercent = (percent: number) => {
+  return `${(percent * 100).toFixed(1)}`;
+};
+
+export const getCategorization = (percentile: number) => {
+  let categorization;
+  let categoryCircleStyle;
+
+  if (percentile >= 0.75 ) {
```
NatHillardUSDS, line 13: I realized actually we now have these in variables in `constants` (`SCORE_BOUNDARY_LOW`, `SCORE_BOUNDARY_THRESHOLD`, and `SCORE_BOUNDARY_PRIORITIZED`) - can you reference those here?

client/src/components/areaDetail.tsx
```diff
@@ -2,91 +2,126 @@ import * as React from 'react';
...
-          <span className={styles.titleIndicatorName}>Just Progress Categorization: </span>
-          <span>{getCategorization(score)}</span>
-        </div>
-        <div>
-          <span className={styles.titleIndicatorName}>Cumulative Index Score: </span>
-          <span>{readablePercent(score)}</span>
-        </div>
-      </div>
-    );
+  const povertyInfo:indicatorInfo = {
```
NatHillardUSDS, line 41: Maybe a comment or TODO here referencing https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/367 ?

client/src/components/areaDetail.tsx
```diff
@@ -2,91 +2,126 @@ import * as React from 'react';
...
+          <div className={styles.priority}>
+            <div className={categoryCircleStyle} />
+            <div className={styles.prioritization}>{categorization}</div>
+          </div>
         </div>
-      </div> :
-    '' }
-    </>
+      </header>
+      <ul className={styles.censusRow}>
```
NatHillardUSDS, line 89: Maybe log a ticket here for localizing this content? this may be a bigger discussion as we may want to have this on the backend; we'll likely get into this more as part of https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/367

client/src/components/mapIntroduction.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+// @ts-ignore
+import lightbulbIcon from '/node_modules/uswds/dist/img/usa-icons/lightbulb_outline.svg';
+import * as styles from './mapIntroduction.module.scss';
+
+const MapIntroduction = () => {
+  return (
+    <aside className={styles.mapIntroContainer}>
+      <header className={styles.mapIntroHeader}>Zoom and select a census block group to view data</header>
```
NatHillardUSDS, line 9: We also want to localize this text - you can see examples here: https://github.com/usds/justice40-tool/blob/7b241795fa9c7f70370b6913cc7dbc6cf57209c4/client/src/components/territoryFocusControl.tsx and here: https://github.com/usds/justice40-tool/blob/cfce0dc8269be08c277edbda339d66e65b703cdf/client/src/components/J40Header.tsx

client/src/components/mapSidePanel.test.tsx
NatHillardUSDS: I would say this is more snapshot test territory - what's important is that the layout matches when elements are composed

client/src/components/mapSidePanel.tsx
NatHillardUSDS: looks like [tsdoc](https://github.com/microsoft/tsdoc) is a pretty good option here - maybe you can make up a ticket to convert to using this and leave a TODO?

client/src/components/mapSidePanel.tsx
```diff
@@ -0,0 +1,50 @@
...
+// See FlagContext.tsx for how we're doing JSDOCs
```
NatHillardUSDS, line 3: You can remove this comment

client/src/components/mapSidePanel.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const tempData = {
+    [indicators[0]]: detailViewData?.properties[indicators[0]],
+    [indicators[1]]: detailViewData?.properties[indicators[1]],
+    [indicators[2]]: detailViewData?.properties[indicators[2]],
+    [indicators[3]]: detailViewData?.properties[indicators[3]],
+    [indicators[4]]: detailViewData?.properties[indicators[4]],
+    [indicators[5]]: detailViewData?.properties[indicators[5]],
+    [indicators[6]]: detailViewData?.properties[indicators[6]],
+    [indicators[7]]: detailViewData?.properties[indicators[7]],
+  };
```
NatHillardUSDS, line 38: You should just be able to pass `detailViewData?.properties` direct to `AreaDetail`, no?  This would save the above two maps

client/tsconfig.json
```diff
@@ -15,6 +15,8 @@
     "allowJs": true,
     "resolveJsonModule": true,
     "baseUrl": "./",
+    // Todo VS: is this really needed?
+    "suppressImplicitAnyIndexErrors": true,
```
NatHillardUSDS: We actually don't want this - more [here](https://www.typescriptlang.org/tsconfig#suppressImplicitAnyIndexErrors) but this can indicate places where we need to be more explicit about typing / can cast

NatHillardUSDS (changes requested):
Overall great fix! 
Awesome that you prioritized testing, device compatibility, and mobile. 💯 !

Had a few comments about structure, testing, and styling - we should talk later about how to break down and prioritize future tasks.

In general, as you note, this PR is a bit big, and it took some time to look over/review and introduced a good handful of changes - we can start to tackle this with followup PRs being smaller and more scoped. I'd imagine a number of the suggested fixes, those larger than cosmetic, can be made as followups, we can talk about this later.

I did notice two call-out-able issues just now in testing this, which I do think we should address if possible as part of this fix:

- When you click on a focus region after clicking a BG, it actually goes to its neighbor:
  - click on a CBG
  - click on another CBG
  - click on "48"
  - It goes to an adjacent CBG, rather than the region that was called out 
(This does not appear to be happening on main)
- When you click on a CBG, it shrinks the map size -- this one in particular should be addressed as part of the first PR if possible.

client/cypress/e2e/mapSidePanel.spec.js
```diff
@@ -0,0 +1,108 @@
...
+const devices = [
+  [1024, 720],
+  ['iphone-6', 'portrait'],
+  ['samsung-s10', 'portrait'],
+];
+
+describe('tests that the map side panel renders MapIntroduction component', () => {
+  devices.forEach((device) => {
+    it(`should render MapIntroduction component on ${device[0]} x ${device[1]}`, () => {
+      cy.visit('http://localhost:8000/en/cejst');
```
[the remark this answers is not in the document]
vim-usds: Awesome! They are all passing now!

client/cypress/e2e/mapSidePanel.spec.js
```diff
@@ -0,0 +1,108 @@
...
+//         cy.get('li[class*="indicatorBox"]').should('be.visible');
+//         cy.get('li[class*="indicatorBox"]')
+//         // currently the padding-top on desktop = 1.5rem => 24px
+//         // currently the padding-top on mobile = .8 rem => 8px
+//             .each((indicator) => cy.wrap(indicator)
+//                 .should('have.css', 'padding-top', `${device[0].isInteger ? '24px' : '8px'}`));
+//       });
+//     });
+//   });
+// });
```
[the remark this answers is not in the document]
vim-usds: yeah so this dynamic test (which basically wraps the other 3 in a for loop), seems to fail. It fails specifically when checking for the padding-top value. This is CSS property changes when in mobile via a media query so using that to test the mobile view. It seems that the padding top doesn't change for mobile when placed in the loop. 🤔

NatHillardUSDS:
Here are some videos of the bugs I described:
1. Here's the first one, what I previously called "the shrinking map" - it seems to happen at half-width in particular. Start with an unselected region, zoom in, and click a CBG, then the side panel grows in height but the map stays the same size. I initially thought the map was shrinking, but instead it was that the side panel was growing. Instead of this, I think we need to set an explicitly pegged height and `overflow-y` to `true`, or similar.
https://user-images.githubusercontent.com/72811320/127033944-fdb86e46-4378-46ca-8e8e-0ef315cb281e.mov

2. The second one is here: 
https://user-images.githubusercontent.com/72811320/127034542-d850b1cc-3665-43a7-ab78-dcd8f78d0252.mov

client/src/components/areaDetail.tsx
vim-usds: added intl here!

vim-usds:
- [ ] Bug on overflow
- [ ] Mobile not resizing on PR
- [ ] Utils scss file instead of module
- [ ] forEach Cypress tests 
- [ ] URL to constants file
- [ ] intl all strings
- [ ] mapInfo to Intl
- [ ] rename to infoPanel from sidePanel
- [ ] add snapshot testing

NatHillardUSDS (commented):
Really Nice! 

Changes look good - only thought would be to make sure that we've got tickets for all the issues you've identified, and in the places in the code that need updating, we reference these tickets with comments to make sure we get back to them -- approved once we've got a paper trail for these other fixes in some form!

NatHillardUSDS:
A related thing I realized : we have another `isMobile` check in constants.tsx ([ref](https://github.com/usds/justice40-tool/blob/main/client/src/data/constants.tsx#L103)), can you move that over to use the package version as well?

client/src/components/mapIntroduction.tsx
vim-usds: added intl!

client/cypress/e2e/mapInfoPanel.spec.js
```diff
@@ -0,0 +1,63 @@
...
+    it(`should render MapIntroduction component on ${device[0]} x ${device[1]}`, () => {
+      cy.visit(ENDPOINTS.EXPLORE_THE_TOOL);
+      cy.viewport(device[0], device[1]);
+      cy.get('aside').should('be.visible');
+    });
+  });
+});
+
+/**
+ * Todo VS:
```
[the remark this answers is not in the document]
vim-usds: added a reference!

NatHillardUSDS (approved):
Sweet! Merge away!

Issue #410, opened 2021-07-26
Territory button does not work if it's located on top of a prioritized region


NatHillardUSDS:
**Describe the bug**
On main (67b39475f), when you click on a focus region (e.g. 48, AK, HI, PR) after clicking a BG, it actually goes to a random neighbor.

**To Reproduce**
1. Go to http://localhost:8000/en/cejst , zoom in
2. click on a CBG
3. click on another CBG
4. click on "48"

**Expected behavior**
It is expected that this would take you to the boundaries of the 48 states. Instead, it goes to an adjacent CBG, rather than the region that was called out.

**Screenshots**
See this video:
https://images.zenhubusercontent.com/358242767/45feb72a-a83d-483f-9d7b-a7e9574adafa/region.mov

**Desktop (please complete the following information):**
All

**Smartphone (please complete the following information):**
All

**Additional context**
This appears related to how we're storing the `selectedRegion` state - we should find a way to have a single source of truth on the selected region state, and if possible, get out of the business of having the map store this state.

PR #412, opened 2021-07-26
Clicking territory focus button sometimes focuses on a different CBG, not the actual territory

NatHillardUSDS:
Fixes issue #410 : Clicking a CBG region, then another, then hitting territory button results in random CBG being selected

What was happening here: a click on the territory focus buttons was bleeding through to the block group behind the button. The main fix was to call `stopPropagation()` on the event callback. Also though, as part of this change, I removed the logic associated with `setFeatureState` , which in practice resulted in the map storing the selected feature as opposed to the component. 

With this change, which uses `react-map-gl`'s [Layer](https://visgl.github.io/react-map-gl/docs/api-reference/layer) , the map component itself owns the selected feature and the `filter` on the Layer component does the work for us. This overall simplifies logic and centralizes data ownership


TomNUSDS (approved):
LGTM

PR #414, opened 2021-07-27
Loading a URL with lat/lng/zoom specified does not work

### LINKED ISSUE #409: Loading a URL with lat/lng/zoom specified occasionally does not work
**Describe the bug**
On latest main (67b39475f704e46942e062e2f1cfec83ea8807c5), if you load a URL with a different lat/lng/zoom than the default, it will occasionally fail to load. I suspect this is a race condition between processing the URL and loading the initial state of the map. 

**To Reproduce**
1. Go to http://localhost:8000/en/cejst#12.44/41.40965/-75.65978 , which should load the Dunder Mifflin factory in Scranton, PA. 
2. But it doesn't! the URL changes to the default, http://localhost:8000/en/cejst/#3/32.47/-86.5

**Expected behavior**
It should load a viewport centered around the Dunder Mifflin factory in Scranton, PA. Instead it loads the default map viewport with the whole US visible

 - OS: MacOS 11.4
 - Browser Chrome
 - Version 92.0.4515.107 

Also happens on all smartphone versions

**Additional context**
In the ideal flow, we initialize the `ReactGLMap` with its own viewport, and later, the URL is processed. Instead sometimes, due to a race condition, I believe, the processing happens out of order, and the URL is processed before the default viewport is loaded, which then overrides these.

NatHillardUSDS:
Fixes #409 - Loading a URL with lat/lng/zoom specified does not work
* We now parse the URL on page before initializing the map
* Removes now-unnecessary `MapWrapper` component (previously used for different map impl comparisons)
* Also adds basic test for this (may make sense as unit test / we can assess when Vim's changes go in)


TomNUSDS (dismissed):
LGTM

vim-usds:
@NatHillardUSDS 

A quick heads up, I actually ended up using the <MapWrapper> component. I placed it all in a section tag and placed all map related markup in that component, ie, header (explore the tool), map, mapInfoPanel and legend.

NatHillardUSDS:
Added back the map wrapper - had removed with a merge conflict but now it's back - thanks for catching!

vim-usds (approved):
Looks great!

Issue #362, opened 2021-07-15
It should be clearer how branches are deployed


NatHillardUSDS:
**Describe the bug**
Right now it's necessary to know where to look to know where and how a ticket is deployed on a branch, and this can be a bit cryptic. We should make it clearer via our github actions where a given branch is going.

**To Reproduce**
1. Push a change to a branch
2. Wait for "build & deploy" action to finish
3. Try to find where it has been posted
4. If you know where to look at present, you can find it:
![image.png](https://images.zenhubusercontent.com/6091a54c180d4758f3eedc40/955d63df-3c5d-476a-9007-bace43a75427)

**Expected behavior**
It should be clear where every branch goes, namely:
1. Github pages: https://usds.github.io/justice40-tool/{HASH}/en
2. Standard S3 bucket version (http only) : http://usds-geoplatform-justice40-website.s3-website-us-east-1.amazonaws.com/justice40-tool/{HASH}/en
3. CloudFront https version: https://d2zjid6n5ja2pt.cloudfront.net/justice40-tool/{HASH}/en/cejst/
This should at very least be output directly at the end of each build and/or posted to the PR that initiates the build. As a stretch goal, we can post to slack.

PR #415, opened 2021-07-27
Merges to main don't get posted to main && it should be clearer how branches are deployed

### LINKED ISSUE #319: merges to main still don't get posted to /main
**Describe the bug**
We have a github action [here](https://github.com/usds/justice40-tool/blob/main/.github/workflows/build_deploy.yml) that is supposed to conditionally set a variable based on whether a particular build was generated as the result of a merge to main or not, but currently it is not funcitoning:

```
- name: Set DESTINATION_FOLDER for main
        if: github.event.pull_request.merged == true
        run: |
          echo "DESTINATION_FOLDER=main" >> $GITHUB_ENV
      - name: Set DESTINATION_FOLDER for branch
        if: github.event.pull_request.merged != true
        run: |
          echo "DESTINATION_FOLDER=$(echo $GITHUB_SHA | cut -c 1-6)" >> $GITHUB_ENV
```

**To Reproduce**
1. Create a PR in `justice40-tool`, then merge it
2. Check out the build page, for example [this one](https://github.com/usds/justice40-tool/runs/3030161022?check_suite_focus=true)
3. The step labeled `Set DESTINATION_FOLDER for main` is not firing, despite this line: `github.event.pull_request.merged == true` 

**Expected behavior**
If the merge is a merge to main, it is supposed to be pushed to `/main`. If it is a PR branch, it is supposed to be pushed to `/{HASH}`

NatHillardUSDS:
fixes #319 - merges to main still don't get posted to /main , and #362 It should be clearer how branches are deployed.

Specifically:
* Moves to using `pull_request` instead of `push` -- this is actually the correct event, as we want these flows to fire when opening and closing pull requests, not simply pushing. We can't push to master directly anyhow.
* Also imposes the restriction that if the PR is closed, and merged, then we should update the `main` URL; otherwise, we should use the hash. Both of these need to be true; we could theoretically close without merging or merge without closing. 
* Only comment with the deployment URL when the PR is opened - when it has been closed, all PRs will go to `/main`. 
* Bot now adds comment to the PR itself with the URL of the preferred deployment
* Job now echoes the full set of URLs to the console when running the flow. 

tl;dr - we were using the wrong event and trigger; fixing this made it so we're setting the right variables to the right values and running at the right time.

(Interestingly because this fix targets `./client`, it won't fire for this change itself, but I have confirmed this fix does indeed work with a local repository.

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -6,12 +6,11 @@
 
 name: GH Pages deploy
 on:
-  push:
+  pull_request:
+    types: [ closed, opened ]
     branches: [main]
     paths:
       - 'client/**/*'
-  pull_request:
-    branches: [main]
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -30,7 +29,7 @@ jobs:
       # If this is just a PR branch, put content in a hash directory. Otherwise, put it in "main"
       # This variable is used by gatsby-config.js to set a prefixPath
       - name: Set DESTINATION_FOLDER for main
-        if: github.event.pull_request.merged == true
+        if: github.event_name == 'pull_request' && github.event.action == 'closed' && github.event.pull_request.merged == true
         run: |
           echo "DESTINATION_FOLDER=main" >> $GITHUB_ENV
       - name: Set DESTINATION_FOLDER for branch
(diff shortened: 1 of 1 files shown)


TomNUSDS (dismissed):
LGTM. Super excited for this change! 

Interesting thing I found when researching this change: many project use labels like feature-flags to control the ci process. `contains(github.event.pull_request.labels.*.name, 's:benchmark-required')` type stuff.

NatHillardUSDS:
> LGTM. Super excited for this change!
> 
> Interesting thing I found when researching this change: many project use labels like feature-flags to control the ci process. `contains(github.event.pull_request.labels.*.name, 's:benchmark-required')` type stuff.

Yeah! This is definitely an interesting opportunity; we used similar branch prefix-based jobs with jenkins at my last job as part of the `git-flow` model (more [here](https://nvie.com/posts/a-successful-git-branching-model/), interesting side note that the creator now advocates for feature branches instead!) - that model has a lot more baggage (and was actually quite helpful for mobile development), but it does seem we could adopt some general branch prefixes for this project to do some cool stuff like pre-releases etc.

TomNUSDS (approved):
Re-approving. Thanks!

Issue #420, opened 2021-07-27
Turn Lighthouse on for all pages and fix failures.


switzersc-usds:
Our current lighthouse test ([here](https://github.com/usds/justice40-tool/blob/main/client/cypress/e2e/accessibility.test.js)) , only runs against the main page. 

The `nathillardusds/failing_lighthouse` branch fixes tests to run instead on all pages. When you do this, you get the below output:
[acccessibility_failures.log](https://github.com/usds/justice40-tool/files/6887301/acccessibility_failures.log)

We should look through these and fix them

switzersc-usds:
We turned it on for all pages and have lots of errors we need to fix!

vim-usds:
All pages are failing the following:

1. document-title
2. html-has-lang
3. landmork-one-main
4. page-has-heading-one

Other than item 2, the pages do have all the others. Needs more investigating why it's failing

vim-usds:
Seeing some light on this.

While the cypress.axe alerts on a11y issues are great for CI/CD, it doesn't show where the error is happening. 
The Axe chrome [plugin](https://www.deque.com/axe/) is showing promise. On localhost it seems to not only identify a11y issues, but also point out where it's occurring. One this is addressed via the plugin,  then one can check the cypress.axe tests to see if alerts are passing. More to come on this.

vim-usds:
All automated a11y testing (aka cypress-axe) are now passing! 🥳 

For both mobile and desktop, all pages on the staging URL has an A11y lighthouse score of 100% ! 🥇 

About = 100% ✅
Explore = 100% ✅
Methodology = 100% ✅
Contact = 100% ✅

The automated a11y testing (via cypress-axe) is more stringent than Lighthouse (it seems). The only _visible_ change to the site that needs review was the addition of an `<h1>` tag to the About page:

- http://usds-geoplatform-justice40-website.s3-website-us-east-1.amazonaws.com/justice40-tool/c40867/en/

Looking forward to your review!

PR #425, opened 2021-07-28
fix #249 header 2

### LINKED ISSUE #249: As an EVCM, I want to see clearly at the top of the site the key things I can navigate to (About, Explore the tool, Data & Methodology, Contact in header)
**Description**
Header should be updated with the links to the pages within the site that we know we'll support for July 27th. 

**Solution**
Implement links to About, Explore the tool, Data & Methodology, Contact in the header

Tom will also add a few extra things here: accessibility support and page meta tags. Probably will also implement localization (breaking out the strings and putting into localization strings).

Related to https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/233 and https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/113

**Tasks**
 - [ ] Update header links, remove "Timeline" and "CEJST"
 - [ ] Link Methodology to new Methodology page https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/233
 - [ ] Link Contact to new Contact page https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/113
 - [ ] Update logo

Already done (at least until content is finalized) is "About" which links to the informational page.

**Definition of "Done"**
 - [ ] Links are navigable on http://usds-geoplatform-justice40-website.s3-website-us-east-1.amazonaws.com/justice40-tool/f12ab4/en/ or similar dev/demo URL

TomNUSDS:
- Fixes As an EVCM, I want to see clearly at the top of the site the key things I can navigate to (About, Explore the tool, Data & Methodology, Contact in header) #249
- Bring the header inline with latest design spec.
- Add unit tests
- New icon
- Remove byline
- Update title
- Move Meta tags out of header and into layout (<Helmet>)
- Remove unused feature flags

I rolled back a bunch of changes on https://github.com/usds/justice40-tool/tree/tomn-usds/fix-249-header and it was easier to start a new branch off main and merge over just what was needed. (Process note: I left that outdated branch/PR in place so it can be referenced to make sure I addressed all issues from PR review.)


TomNUSDS (commented):
@NatHillardUSDS @vim-usds 
package-lock.json got bumped to `"lockfileVersion": 2,` on main? 
We really need to figure out how to prevent the bouncing between 1 and 2

client/src/components/J40Header.tsx
```diff
@@ -89,67 +93,50 @@ const J40Header = () => {
...
       </Header>
-      <Alert
-        className={'j40-sitealert'}
-        type="info">
-        <b>Public beta — </b>
-        Welcome to the public beta of the Just Progress Map, a climate and
-        economic justice screening tool. The tool will be continuously updated.
-        Please submit feedback.
+      <Alert className={'j40-sitealert'} type="info">
+        <b><FormattedMessage
```
ginabeena, line 130: bold of you to use a bold tag 😺

client/src/components/layout.tsx
```diff
@@ -2,33 +2,33 @@ import React, {ReactNode} from 'react';
...
+        <meta charSet="utf-8"/>
+        <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+      </Helmet>
       <J40Header/>
       <GridContainer containerSize={'desktop-lg'}
         className={'j40-grid-container'}>
         <Grid row>
           <main id={'main-content'}
-            className={'usa-layout-docs j40-main-content ' + gridCssClass}>
+            className={'usa-layout-docs j40-main-content desktop:grid-col-12'}>
```
ginabeena, line 29: you could always extend the class `desktop:grid-col-12` into the `usa-layout-docs` class name in scss

client/src/pages/cejst.tsx
```diff
@@ -15,45 +15,51 @@ const CEJSTPage = ({location}: IMapPageProps) => {
...
+          <DownloadPacket/>
+        </div>
+      </section>
+
+      <section>
+        <h2>Explore the Tool</h2>
         <MapWrapper location={location} />
-        <HowYouCanHelp />
-      </main>
+        <HowYouCanHelp/>
```
ginabeena, line 61: how does this work?

client/src/pages/contact.tsx
```diff
@@ -1,16 +1,47 @@
...
 const ContactPage = ({location}: ContactPageProps) => {
+  const generalemail = 'screeningtool.feedback@usds.gov';
+  const techemail = 'screeningtool.support@usds.gov';
+
   return (<Layout location={location}>
     <section className={'usa-prose'}>
-      <h1>Contact</h1>
+      <h2><FormattedMessage
+        id={'contact.pageheader'}
+        description={'H2 header for contact page'}
```
ginabeena, line 17: who sees this description? you can probably remove the html comments in there and just say `Contact Page Header` or something

client/src/styles/global.scss
```diff
@@ -84,19 +84,61 @@ $primary-color: #112f4e;
...
 
   // this is the title
-  .usa-logo__text {
-    font-size: 1.8em;
-    font-family: "serif";
-  }
+  @include at-media("desktop") {
+    .j40-title {
+      font-size: 0.9em;
+      padding-top: 25px; // allows vertical centering on logo and text
```
ginabeena, line 90: if the font size is responsive depending on root document size, but the padding is specific to pixels, this will break on some views. not worth fixing, but probably code smell

vim-usds:
> @NatHillardUSDS @vim-usds
> package-lock.json got bumped to `"lockfileVersion": 2,` on main?
> We really need to figure out how to prevent the bouncing between 1 and 2

Looking into this, its seems that this is related to the npm version we might have:
https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion

What version do we have?
VS: 6.14.13
TN: ?
NH: ?

I'm guessing we don't all the same version which might be causing the jump between `lockfileVersion`s?

client/src/styles/global.scss
```diff
@@ -185,12 +227,25 @@ $primary-color: #112f4e;
...
 
-  padding-top: 7px;
-  padding-bottom: 7px;
-  min-height: 4em;
+  min-height: 2.5em;
+  margin: 0 !important;
 
-  * + .usa-alert {
+  .usa-alert {
     margin: 0 !important;
```
ginabeena, line 234: Kind of worried that we need to override everything but that isn't really in your PR, more a comment that we will probably need to refactor custom styles at some point.

ginabeena (approved):
If it looks good on your machine :shipit:

client/src/components/J40Header.tsx
NatHillardUSDS: Yeah we should probably use CSS here

client/src/pages/cejst.tsx
```diff
@@ -15,45 +15,51 @@ const CEJSTPage = ({location}: IMapPageProps) => {
...
-                page.
-              </p>
-            </div>
-            <DownloadPacket />
+      <section>
+        <h2>Just Progress communities</h2>
+        <div className={styles.disclaimer}>
+          <div className={styles.textBox}>
+            <p>
+              Just Progress helps identify and prioritize communities across the
```
NatHillardUSDS, line 23: I know this was here before, but we should make sure this is wrapped with an `intl.formatMessage` - maybe a task ticket for @vim-usds ?

client/src/pages/cejst.tsx
```diff
@@ -15,45 +15,51 @@ const CEJSTPage = ({location}: IMapPageProps) => {
   //   We will bring back later when we have interactive controls.
   return (
     <Layout location={location}>
-      <main id="main-content" role="main">
```
NatHillardUSDS, line 18: We should still have a main, right? This is one of the warnings we've been getting from Lighthouse - "Ensures all page content is contained by landmarks". I was able to fix this locally for a different page by adopting the guidance here: https://www.w3.org/WAI/tutorials/page-structure/regions/

client/src/pages/contact.tsx
NatHillardUSDS: (side note, this is for translators, who may or may not have the full context of the page and may be seeing this string in isolation  - better to leave a semantic description of how it's used in a way that differentiates between the multiple senses of "contact" -- more on this here: https://www.transifex.com/blog/2020/tips-providing-context-translation/ , but something like "Heading for page to allow users to contact project maintainers" is a typical comment / description for this kind of string. A small point, and not something to change now necessarily, but we can talk with the translators and ask what they would find useful as well when it comes to this)

client/src/styles/global.scss
```diff
@@ -84,19 +84,61 @@ $primary-color: #112f4e;
...
+    .usa-navbar {
+      width: 100%;
+    }
+
+    // this actual site logo
+    .sitelogo {
+      float: left; // allows vertical centering on logo and text
+      margin-right: 0.5em; // space between logo and text
+      width: 4em;
+      padding: 0.5em; // this will change the size of the logo too
```
NatHillardUSDS, line 104: appreciate these comments!

NatHillardUSDS (approved):
Approved overall, some small things; the accessibility-related issue with `main` is something we should tackle site-wide -- as you can see in https://github.com/usds/justice40-tool/issues/420 this is on every page -- so doesn't necessarily need to be tackled here but we should fix

client/src/components/J40Header.tsx
TomNUSDS: Yeah. Agreed. The title and message are now in spans with css classnames that can control them.

client/src/components/layout.tsx
TomNUSDS: Yeah, my scss foo is weak. I'd love to do a separate PR to just clean up this stuff and pair work on it. @NatHillardUSDS and @vim-usds might like be interested as well.

Is it ok if I leave it messy for this PR?

client/src/pages/cejst.tsx
TomNUSDS: Yeah, I started to localize the cejst file, but then thought better of mixing too many things into the "fix header" changelist.

Can I get the PR for the header and then do the cleanup/localization on a different PR?

client/src/pages/cejst.tsx
TomNUSDS: @NatHillardUSDS We should do a code in zoom walkthrough!

client/src/pages/contact.tsx
TomNUSDS: Changed to `"Heading for page to allow users to contact project maintainers"`. 👍

client/src/styles/global.scss
TomNUSDS: Good call. Thanks!

TomNUSDS: Ends up this was not needed.

client/src/styles/global.scss
TomNUSDS: 100% agree! I think I managed to remove it in this case, but there are other places we should try to fix.

client/src/pages/cejst.tsx
vim-usds: I agree, I think let's make this part of the other ticket on the methodology page

client/src/pages/cejst.tsx
vim-usds: I want to say this might be ok to delete, since the layout component already has a `main`. I think this `main` under question, is the 2nd nested `main` that is being flagged by the a11y tool that Glennette is using

vim-usds (approved):
LGTM

client/src/pages/cejst.tsx
TomNUSDS: The issue is that you shouldn't have more than one `<main>` per page (I think the accessibility report touched on this). 

The `layout.tsx` has this element, so it shouldn't be in any per-page templates.

NatHillardUSDS: ah ok this makes sense

Issue #262, opened 2021-06-28
As a developer, I want to have a YAML file describing the metadata for our existing datasets.


switzersc-usds:
**Description**
We will be using the YAML file to be the source of truth for datasets used in the tool, and the frontend (for now) will be calling on the YAML files to display information on the website. We need to create these YAML files and flesh it out with info for our existing data sets.

**Solution**
Create one YAML file per dataset and update with necessary fields. During the data-pipeline process we also want to pull these and create a JSON file for the frontend to consume.

**Describe alternatives you've considered**

**Links to user research or other resources**

**Tasks**
- [ ] Identify the list of fields needed by the mockup (ie what we want on the site) - Sync with Design on what's available and what we might want. These are in screenshots.
- [ ] Identify what other fields we want in the file
- [ ] Session showing product and design how to create and update these files
- [ ] Create files
- [ ] Implement JSON conversion

**Definition of "Done"**
- [ ] We have an accurate YAML file for each of the 3 data sets (ACS, EJScreen, and housing)
- [ ] Data is available from S3 or Github as JSON file (this may be generated and available from a Github action)

NatHillardUSDS:
Note from discussion just now: we should convert to json for easier consumption by the frontend

BethMattern:
Here is a sample .csv of what we were thinking of including. This isn't the final list of indicators but is an example of the data we want to display.
[Just Progress Datasets - Sheet1.csv](https://github.com/usds/justice40-tool/files/6772376/Just.Progress.Datasets.-.Sheet1.csv)

BethMattern:
Here's an excel file for the data indicators that are currently involved in the score that should be the source for this page.[https://app.zenhub.com/files/358242767/1b63fbb7-f7d5-4673-88b0-8515f9846e63/download](https://app.zenhub.com/files/358242767/1b63fbb7-f7d5-4673-88b0-8515f9846e63/download)

switzersc-usds:
@BethMattern can you please list the fields here from the spreadsheet? This should be a list of all fields the frontend needs for a given indicator or dataset.

BethMattern:
Fields needed to display in UI:
Indicator name
Long description
Short description
Data resolution
Category
Data source
Date range
Link to source

cc: @esfoobar-usds @switzersc-usds

BethMattern:
@esfoobar-usds this can be closed right? It's an old version of the YAML work I believe.

esfoobar-usds:
[...]

PR #427, opened 2021-07-28
Adding template for dataset request

NatHillardUSDS:
Part of #262 - As a developer, I want to have a YAML file describing the metadata for our existing datasets 

This is an experimental attempt to add a github form for dataset requests , which will essentially auto-validate new dataset requests at the time of input by using newly-introduced [github issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository). 

We can then setup actions to parse this content and do things with it, using something like [this](https://github.com/peter-murray/issue-forms-body-parser)

### DIFF
--- a/.github/ISSUE_TEMPLATE/dataset_request.yaml
+++ b/.github/ISSUE_TEMPLATE/dataset_request.yaml
@@ -1,0 +1,174 @@
+name: Dataset Request
+description: Request a datast for inclusion in the product
+title: "[Dataset Request]: "
+labels: [dataset, triage]
+assignees:
+  - switzersc-usds, lucasmbrown-usds
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for taking the time to fill out this dataset request!
+  - type: input
+    id: contact
+    attributes:
+      label: Contact Details
+      description: How can we get in touch with you if we need more info?
+      placeholder: ex. email@example.com
+    validations:
+      required: false
+  - type: input
+    id: dataset-name
+    attributes:
+      label: What is the name of the dataset?
+      description: Try to provide the standard name under which this dataset is known
+      placeholder: Dataset name
+    validations:
+      required: true
+  - type: input
+    id: dataset-source
+    attributes:
+      label: What is the source URL of the dataset?
+      description: Provide the canonical source URL of the dataset
+      placeholder: www.example.com
+    validations:
+      required: true
+  - type: textarea
+    id: relevance-to-ej
+    attributes:
+      label: What is the relevance of the dataset to environmental justice?
+      description: Please help us to see the relevance of your suggestion in detail
+      placeholder: factor X has a significant impact on health
+    validations:
+      required: true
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/dataset_request.yaml
```diff
@@ -0,0 +1,174 @@
...
+    id: operating-systems
+    attributes:
+      label: What formats does this source support?
+      description: You may select more than one.
+      options:
+        - label: CSV
+        - label: XLS/XLSX
+        - label: geoJSON
+        - label: ArcGIS REST API
+        - label: other
```
ginabeena, line 54: other...? 😈

.github/ISSUE_TEMPLATE/dataset_request.yaml
```diff
@@ -0,0 +1,174 @@
...
+    validations:
+      required: false
+  - type: dropdown
+    id: spatial-resolution
+    attributes:
+      label: What is the spatial resolution of this dataset?
+      options:
+        - County
+        - Census tract
+        - Census block group
```
ginabeena, line 70: Wonder if people will know what these mean. Census tract vs Census block group... 🤔

.github/ISSUE_TEMPLATE/dataset_request.yaml
```diff
@@ -0,0 +1,174 @@
...
+    validations:
+      required: true
+  - type: dropdown
+    id: data-quality-rating
+    attributes:
+      label: How would you rate the quality of this data?
+      options:
+        - Low
+        - Medium
+        - High
```
ginabeena, line 97: What would be an example of low quality data and why would someone use it?

.github/ISSUE_TEMPLATE/dataset_request.yaml
```diff
@@ -0,0 +1,174 @@
...
+      description: Provide known margin of error if known
+      placeholder: 5%
+    validations:
+      required: false
+  - type: textarea
+    id: data-quality-issues
+    attributes:
+      label: Known data quality issues
+      description: Please describe any data quality issues you know about with this dataset.
+      placeholder: Limited accuracy in rural areas, etc.
```
ginabeena, line 113: oh i see...

ginabeena (approved):
Amazing idea and I love that this hooks into forms. Great work Nat.

esfoobar-usds (approved):
🍏 - great work @NatHillardUSDS.

@lucasmbrown-usds: at some point we want to discuss this new approach with you.

NatHillardUSDS:
These are all valid questions, thank you @ginabeena for weighing in, I agree with all of these. 

For now, for the purposes of discussion, our thought was to merge this as-is so we see how it looks with the github issue forms renderer, though we definitely need to tweak these going forward to made them easier to understand etc. - we can make up a followup issue for this if we decide to go forward with this setup

PR #431, opened 2021-07-29
Add dataset cards to methodology page

vim-usds:
This is just an initial draft to get some quick feedback from design.


client/src/components/datasetCard.module.scss
```diff
@@ -0,0 +1,40 @@
+.datasetCard {
+  background-color: white;
+  padding: 2.7rem 3rem 3rem 3rem;
```
NatHillardUSDS, line 3: Good call on using `rem` - I was reading about this yesterday and I didn't realize the limitations around using `em`, which is local to its context. `rem` definitely seems superior

client/src/components/datasetCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <h3 className={styles.datasetCardIndicator}>{datasetCardProps.indicator}</h3>
+      <div className={styles.datasetCardWhatIsIt}>What is it?</div>
+      <div className={styles.datasetCardDescription}>
+        {datasetCardProps.description}
+      </div>
+
+      <ul className={styles.datasetCardList}>
+        <li className={styles.datasetCardListItem}>
+          <span className={styles.datasetCardLabels}>
+            Data resolution:&nbsp;
```
NatHillardUSDS, line 22: Two questions about this one:
1. Can we move the `nbsp` to css ?
2. Can we use `formatMessage`?

NatHillardUSDS (approved):
Overall looks good, a few smaller comments that we could optionally address later as part of other tickets

client/src/components/datasetCard.module.scss
vim-usds: yeah totally still up in the air on this.. also heard that rem works well!

client/src/components/datasetCard.tsx
vim-usds: great call on the `nbsp`! and yes to `formatMessage`

vim-usds: nbsp = not being super productive (lazy) 🤣

vim-usds:
Draft PR is completed.. time for the real thing!

PR #433, opened 2021-07-29
Tile Generation Script

esfoobar-usds:
Satisfies #361


data/data-pipeline/tile/generate.py
```diff
@@ -1,63 +1,54 @@
...
+    os.system(cmd)
 
-    # remove existing mvt directory
-    mvt_tiles_path = data_path / "tiles" / "mvt"
-    if os.path.exists(mvt_tiles_path):
-        shutil.rmtree(mvt_tiles_path)
+    # generate high mvts
+    logger.info(f"Generating USA High mvt folders and files")
+    cmd = "tippecanoe "
+    cmd += " --drop-densest-as-needed --no-tile-compression  -zg "
```
NatHillardUSDS, line 35: Might be clearer to include the full version of the args rather than the abbreviated - I was initially surprised to find that `-zg` did what it does, and needed the full version to know.

Relatedly, actually, we may not want `-zg`, and/or may want to explicitly specify a low zoom. 
To generate the current high tiles I ran
```sh
docker run --rm -it -v ${PWD}:/data tippecanoe tippecanoe --no-tile-compression -Z 5 -z 11 -e /data/usa_high_prdc -l blocks /data/usa_merged_prdc.geojson
```
for instance, which makes it more explicit what the bounds are. `-zg` feels a bit automagic to me

data/data-pipeline/tile/generate.py
```diff
@@ -1,63 +1,54 @@
...
-
-    # generate mbtiles file
-    cmd = (
-        "tippecanoe --drop-densest-as-needed -zg -o /home/data/tiles/block2010.mbtiles --extend-zooms-if-still-dropping -l cbg2010 -s_srs EPSG:4269 -t_srs EPSG:4326 "
-        + geojson_list
-    )
+    # generate low mbtiles file
+    logger.info(f"Generating USA Low mbtiles file")
+    cmd = "tippecanoe "
+    cmd += "--force -Z 5 -z 11 -l usa_low "
```
NatHillardUSDS, line 43: Low should be from 0-7

NatHillardUSDS: We should also include un-abbreviated args

data/data-pipeline/tile/generate.py
```diff
@@ -1,63 +1,54 @@
...
 
-    # generate mvts
-    cmd = (
-        "tippecanoe --drop-densest-as-needed --no-tile-compression  -zg -e /home/data/tiles/mvt "
-        + geojson_list
-    )
+    # generate low mvts
+    logger.info(f"Generating USA Low mvt folders and files")
+    cmd = "tippecanoe "
+    cmd += " --drop-densest-as-needed --no-tile-compression  -zg "
```
NatHillardUSDS, line 51: Likewise with `-zg` here -- we should also generally make sure that `mbtiles` and `mvt` behaviors match

NatHillardUSDS (changes requested):
overall looks good! need logic change on low tile generation and I would recommend un-abbreviating args for clarity

esfoobar-usds:
@NatHillardUSDS feel free to check locally and make changes and push -- these flags are a mashup of things I found looking around, but feel free to add/remove flags as you see fit

esfoobar-usds:
[I was reading up on the options today](https://github.com/mapbox/tippecanoe#options), I think I have a good idea of next steps.

esfoobar-usds:
@NatHillardUSDS ready for another review

data/data-pipeline/tile/generate.py
```diff
@@ -24,31 +25,30 @@ def generate_tiles(data_path: Path) -> None:
...
     cmd += str(score_geojson_dir / "usa-low.json")
-    os.system(cmd)
+    call(cmd, shell=True)
 
     # generate low mvts
     logger.info(f"Generating USA Low mvt folders and files")
-    cmd = "tippecanoe "
-    cmd += " --drop-densest-as-needed --no-tile-compression  -zg "
-    cmd += "-e " + str(low_tile_path) + " "
+    cmd = "tippecanoe --minimum-zoom=0 --maximum-zoom=7 --no-tile-compression "
```
NatHillardUSDS, line 51: might make sense for min and max zoom for low/high to be constants, as they're repeated a few times in this file - easier to change if we need to alter in the future / easier to keep mvt and mbtiles in sync

NatHillardUSDS (commented):
suggestion to use a constant for these values, otherwise looks good

NatHillardUSDS (approved):
LGTM!

Issue #402, opened 2021-07-22
ADR: Use OSM/Nominatim for Geolocation


NatHillardUSDS:
# Use OSM/Nominatim for Geolocation

## Context and Problem Statement

We need a short-term provider of a geolocation service, that is free, open source, and usable for basic searches in our product.

## Decision Drivers

- Free
- Open Source
- Usable for basic general-purpose searches
- Capable of servicing a small number of searches

## Considered Options

- OSM/Nominatim
- Various paid alternatives: Google/Bing/Others
- Data Science Toolkit
- Census REST API

## Decision Outcome

Chosen option: OSM/Nominatim, because it is a free an open-source solution that allows for low-volume searches provided you follow straightforward restrictions.

We will monitor our traffic stats and re-evaluate if we have a large surge of people or need to otherwise account for higher traffic.

### Positive Consequences

- Free
- Open Source
- Suitable for general-purpose searches
- Able to give polygon boundaries with `polygon_geojson=1` option

### Negative Consequences

- Results are not as high quality as with some other services
- Various restrictions apply to search volume:
  - a. We must have "an absolute maximum of 1 request per second"
  - b. We must provide a valid HTTP Referer or User-Agent identifying the application
  - c. Clearly display attribution (using the AttributionControl)
  - d. We are advised to setup a proxy to enable caching of search requests.
  - e. They also have this note: "Note: periodic requests from apps are considered bulk geocoding and as such are strongly discouraged. It may be okay if your app has very few users and applies appropriate caching of results. Make sure you stay well below the API usage limits."
- Relevant result summary -- Here is the result of various kinds of relevant searches using Nominatim. About ~50% accuracy on a variety of responses:
    - St. Paul, Virginia - kind of works
    - St. Paul, VA - did not work
    - 24283 - works
    - Appalachia - did not work (not surprising)
    - St. Paul - did not work
    - Wise County - works
    - Wise County, VA - works
    - St Paul - does not work
    - Clinch river - works (surprisingly)
    - [a more rural address] 3025 4th Ave St. Paul, VA - does not work
    - 3025 4th Ave St. Paul, VA 24283 - does not work
    - 3025 4th Ave - does not work
    - Western Front Hotel - did not work
    - 3025 Fourth Avenue - does not work
    - 4th Ave and Broad St - does not work
    - Carytown - works
    - Carytown, Richmond - works
    - [a more urban address] 3109 W Cary St, Richmond, VA 23221 - works
[...]

PR #437, opened 2021-07-29
Adding ADR PR for associated ADR issue #402

NatHillardUSDS:
PR to add ADR for using Nominatim for search (associated with issue #402)


vim-usds (approved):
LGTM!

PR #445, opened 2021-08-02
adds nvmrc file

vim-usds:
Adds an nvmrc file. 

Install [NVM](https://github.com/usds/justice40-tool/blob/main/INSTALLATION.md#install-node-using-nvm) can be seen here.

Once NVM and node is installed, use the recommended version of node via:

`$ nvm use`


NatHillardUSDS (approved):
this works; looks like there's no way to explicitly pin down the version of `npm` itself, but we can generally assume that people are using the version you get by default (that is, without having done `npm install -g npm@lts`)

Issue #438, opened 2021-07-29
Links in footer aren't correct


TomNUSDS:
Verify all links in the footer are correct. Related to #220

TomNUSDS:
Dup to #220

PR #447, opened 2021-08-02
fix footer layout and links

### LINKED ISSUE #220: Update footer to match design spec
## Description
We need to update the footer for launch to match design spec. Agency partners should no longer be listed.

## Solution
Design to confirm spec

TomNUSDS:
Fix up trusswork's `<Footer>` + `<Logo>` + `<Address>` to match design
Fixes #220 and #438


vim-usds (commented):
Looks good! :)

Not sure if mobile is part of this, if not, its gtg. 

If it is, then I'm seeing an accordion effect on the footer. I think the mobile mocks may be different.

TomNUSDS:
> If it is, then I'm seeing an accordion effect on the footer. I think the mobile mocks may be different.

Good catch. Ugh. That's just how the USWDS footer behaves apparently? Seems wrong. @janelien 
https://designsystem.digital.gov/components/footer/

I'm going to go through some of these sites and see how they are doing things.
https://designsystem.digital.gov/getting-started/showcase/all/

janelien:
I looked through a few of these and they didn't seem to have an issue, though they do all seem a bit different from one another. This one is similar in design to what we're trying to do: https://www.dds.mil/

vim-usds (approved):
Maybe we can ask design and product about the trade-offs we spoke about today when using truss components? The components-approach seem to be recommended by thought leaders in the space. It also seems to be a challenge at times to modify the components to match designs. The solution is probably a mixture of (in no particular order):

1. understanding the flexibility of components (story book might help with this)
2. have designs use mostly recommended components ( I believe this might already be happening)
3. have product understand that when 1 and 2 are explored, we may have to reach for custom components that may not have the long term stability as the recommended components.

NatHillardUSDS (approved):
Re-approved after modifications

PR #454, opened 2021-08-04
[DRAFT] explore tool page sprint 4 design

vim-usds:
testing deployed URL


katherinedm-usds:
overall looks great thanks Vim! I noticed a couple little things about the color key box when loading that I'll open a separate bug for.

PR #464, opened 2021-08-05
Backend change for Zipfile

### LINKED ISSUE #303: As a J40 developer, I want to be able to generate a zip file that represents the score and upload it to a static S3 URL, so that EVCMs can download it
**Description**
We need to add a step to the ETL Score generation, so that it generates a downloadable version of all the CBGs with their indicators and scores in a static S3 URL that the frontend client can point to for EVCMs to download it locally on their computer.

**Solution**
Add a method on the Score ETL that generates a CSV locally and if there are AWS credentials present on the container, uploads it to an S3 bucket.

None

**Tasks**
- [ ] Design/product to provide column names
- [ ] Create the local file with the CSV columns based on a spec 
- [ ] Write functionality to upload the file to S3

**Definition of "Done"**
- [ ] A CSV file is generated on every score run

NatHillardUSDS:
Fixes #303 - As a J40 developer, I want to be able to generate a zip file that represents the score and upload it to a static S3 URL, so that EVCMs can download it

What's left:
- [ ] Push tests
- [ ] Push separate change using and describing VSCode debugging (separate ticket)
- [ ] Create Github Action to upload file to S3 on changes to data dir
- [ ] FE Change to use URL


vim-usds:
Looks awesome!

Just wanted to communicate that I ran into some issues downloading anything other than a zip file in Safari. I believe the code may be creating directories for other file types (csv, xls, etc). Are these directories meant to be reached by the FE? Or are they temporary placeholders in the pipeline or maybe something else?

NatHillardUSDS:
> Looks awesome!
> 
> Just wanted to communicate that I ran into some issues downloading anything other than a zip file in Safari. I believe the code may be creating directories for other file types (csv, xls, etc). Are these directories meant to be reached by the FE? Or are they temporary placeholders in the pipeline or maybe something else?

Yeah, this is important to note - we should codify and formalize which files will be accessed by the frontend in some way; since we're only laying down paths now, the idea is that the FE should select which files are relevant and access those paths. In this case, this is only the zipfile for now

NatHillardUSDS:
Closing for now, reopening rebase

PR #457, opened 2021-08-04
Data directory should adopt standard Poetry-suggested python package structure

NatHillardUSDS:
Fixes #456 - Data directory should adopt standard Poetry-suggested python package structure

PR #465, opened 2021-08-05
Hotfix for fips zip download location + added full-score-run command

esfoobar-usds:


NatHillardUSDS (dismissed):
There's some overlap with the latest fix I pushed to https://github.com/usds/justice40-tool/pull/457 (specifically [this commit](https://github.com/usds/justice40-tool/pull/457/commits/9407add9eb14a6da7a221eab2672fe878597e30d)) 

That said, I can approve this and work out the difference for now ; notably it looks like there are linter issues though so you'll need to run the commands in the tox file

Issue #466, opened 2021-08-05
Task: Check in VSCode config for easier local debug


NatHillardUSDS:
## Describe the task
We should check in a version of the `.vscode` folder with pointers for easier local development. 

## Acceptance Criteria

- [ ] launch.json with targets defined
- [ ] task.json with tasks defined (linting, formatting, etc)
- [ ] Flows are documented

## Additional context

* Because the click CLI uses `console_scripts`, commands cannot be run / debugged directly via VSCode by calling function names; it is necessary instead to call the function as a user would with the translated, hyphenated command. More on this [here](https://stackoverflow.com/questions/64556874/how-can-i-debug-python-console-script-command-line-apps-with-the-vscode-debugger)

Issue #430, opened 2021-07-28
update frontend to use actual zip file URL


vim-usds:
The current zip file only has a zipped csv. The final zip file will have csv, pdf, xlsx, etc.

1. replace URL (if needed)
2. validate unit tests still work
3. validate e2e tests still work
4. test on Safari as well.

Issue #341, opened 2021-07-13
[SPIKE] As a J40 developer, I want to write Unit Tests for etl_score_post, so that tests are run on each commit.


esfoobar-usds:
**Description**

Update: changed this scope to just be fore etl_score_post instead of all data pipeline classes. 

We want to make sure any code updates don't break the ETL processes or Score generation, so we will be writing tests that are ran via Github actions on every commit to the codebase.

This card is just for the spike to figure out the pattern (and any refactors needed) to get unit tests going, for example an initial ETL test suite (we need 11 total, one for each source we're ETLing). 

**Solution**
We will write `pytest` modules with small data fixtures and assert that the ETL processes, including the score generation, output the right values and columns.

**Describe alternatives you've considered**
None

**Links to user research or other resources**
None

**Tasks**
- [ ] Review current draft PR from Nat
- [ ] Determine direction for any needed data-pipeline refactor
- [ ] Create tickets for adding unit tests

**Definition of "Done"**
- [ ] Create new, specific testing-backend tickets for adding unit tests
- [ ] Potentially create a refactor pull request if needed

switzersc-usds:
@NatHillardUSDS has written the first one of these (still in progress). We want to break these out into smaller tickets so it's easier to tackle one at a time / simultaneously. We think each ETL test suite will be about 1 day's worth of work. We have 10 more to do.

We want to get to a good pattern that is reusable for future tests, and then de-prioritize finishing these off for now until we know which data sources are critical for the score and therefor require tests.

Issue #309, opened 2021-07-08
Excel file with styling/formatting in zip download packet


katherinedm-usds:
**Description**
We want to support federal agency program officers in being able to filter, sort, and easily digest the data shown for census block groups in the community list using a nicely styled Excel spreadsheet instead of a plain old CSV. 

**Solution**
Design for styling of the .xlsx, potentially including columns to distinguish location information (state, county, CBG ID) from score information, from "bonus" indicators

**Describe alternatives you've considered**

**Links to user research or other resources**
Related card with task for column headers https://github.com/usds/justice40-tool/issues/303 

**Tasks**
 - [ ] Create .xlsx mockup showing desired column and column header styling

**Definition of "Done"**
 - [ ] .xlsx for packet is styled consistent with design mockup

BethMattern:
Test comment

BethMattern:
test comment

katherinedm-usds:
Adding Glennette as design assignee

katherinedm-usds:
Beth has shared document plus descriptions of indicators for Glennette to mockup what Excel styling could look like

glennettec:
Here is the mockup for the .xls file.
There are three tabs:

- Dataset definitions
- Score data
- Non-score data

[Screening Tool dataset download for .xls](https://app.zenhub.com/files/358242767/526e9374-7b95-411b-a5ab-a84a9eb53c25/download)

glennettec:
Updated mockup for the .xls file. Passed Excel's accessibility checker.

Four tabs:

- About the score
- Dataset definitions
- Score data
- Non-score data

[Screening Tool Datasets download](https://app.zenhub.com/files/358242767/b44d9a55-db3e-4144-962e-4aa393d7b472/download)

esfoobar-usds:
Duplicate of #303

katherinedm-usds:
this ticket is specifically around styling, .xlsx file that is now part of the download packet has readable column names in bold, but still has more formatting and explanatory tabs to be added. Adding Blocked label as this is blocked until indicators to be included are finalized. We can format and add explanatory tabs to meet Design spec as a manual step for launch.

PR #469, opened 2021-08-05
Backend change for Zipfile pt. 2

### LINKED ISSUE #303: As a J40 developer, I want to be able to generate a zip file that represents the score and upload it to a static S3 URL, so that EVCMs can download it

NatHillardUSDS:
Fixes #303 - As a J40 developer, I want to be able to generate a zip file that represents the score and upload it to a static S3 URL, so that EVCMs can download it

What's left:

- [x] Create Github Action to upload file to S3 on changes to data dir - will update this ticket

What's covered in other tickets:
- [ ] Tests (#341 
- [ ] VSCode debugging (#466 )
- [ ] FE Change to use URL (#430)


NatHillardUSDS:
The full pipeline now runs, as you can see [here](https://github.com/usds/justice40-tool/runs/3267245733?check_suite_focus=true) . This works, but is not a particularly good setup. 

Filed #485 just now to improve things to instead initiate the job on AWS  , which is probably a more sustainable arch

esfoobar-usds (approved):
🍏 

Only question I have is about the Excel version -- I thought @lucasmbrown-usds had written some formatting thingies on an Python-generated Excel?

NatHillardUSDS:
interesting re: excel formatting, I can sync with Lucas on this - I was thinking to touch it up as part of #309 which I can take on shortly.

PR #470, opened 2021-08-05
Updating census ETL and checking for existing files

### LINKED ISSUE #467: If the census download task is run more than once, us.csv doubles in size and all data is removed from dataframe
**Describe the bug**
If you run the census ETL more than once, us.csv ends up with ~600K rows instead of ~300K. Later on, when we do a duplicate check in `etl_post`, all the data is removed

**To Reproduce**
1. Run `poetry run download_census`
2. Run it again, without running the clean task
3. Check output of `etl_score_post` - data has been removed entirely because of a duplicate removal step

**Expected behavior**
These tasks should be idempotent, able to be run multiple times without altering the results.

NatHillardUSDS:
Fixes #467: If the census download task is run more than once, us.csv doubles in size and all data is removed from dataframe

Additionally updates census to use now-standard ETL structure

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/sources/census/etl.py
+++ b/data/data-pipeline/data_pipeline/etl/sources/census/etl.py
@@ -2,6 +2,7 @@ import csv
 import json
 import os
 from pathlib import Path
+import subprocess
 
 import geopandas as gpd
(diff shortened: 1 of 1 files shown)


esfoobar-usds (approved):
🍏 -- and just in time!

PR #474, opened 2021-08-05
Census ETL should use standard ETL form

### LINKED ISSUE #473: Census ETL should use standard ETL form
**Describe the bug**
`data/data-pipeline/data_pipeline/etl/sources/census/etl.py` currently has a very different structure from other ETL files. For the sake of consistency and testability, we should convert this over to use the new format. 

**To Reproduce**
1. Compare `data/data-pipeline/data_pipeline/etl/sources/census/etl.py` to `data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py` 
2. The former is not a subclass of `ExtractTransformLoad`, but the latter is

**Expected behavior**
We should subclass `ExtractTransformLoad` and setup the file like other ETL processes

NatHillardUSDS:
Fixes #473 - Census ETL should use standard ETL form

Reformats Census file to look and behave like other ETL classes; also separates out functionality into smaller, testable units

### DIFF
--- a/data/data-pipeline/data_pipeline/application.py
+++ b/data/data-pipeline/data_pipeline/application.py
@@ -2,7 +2,6 @@ import click
 
 from .config import settings
 from .etl.runner import etl_runner, score_generate, score_geo
-from .etl.sources.census.etl import download_census_csvs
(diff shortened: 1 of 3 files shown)


esfoobar-usds (approved):
🍏

PR #475, opened 2021-08-06
[DRAFT] Github Action to upload data to S3

NatHillardUSDS:
[DRAFT for debugging purposes]
A draft of a github action to upload content to S3  - need to test remotely

### DIFF
--- a/.github/workflows/deploy_data.yml
+++ b/.github/workflows/deploy_data.yml
@@ -10,52 +10,42 @@ on: [push, pull_request]
 jobs:
   deploy_data:
     runs-on: ubuntu-latest
+    defaults:
+      run:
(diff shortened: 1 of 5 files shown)


NatHillardUSDS:
The latest draft has succeeded, data posted here: https://github.com/usds/justice40-tool/runs/3266557560?check_suite_focus=true !!

Need to update the other PR so that we're also posting the `downloadable` data, will do that shortly.

That said, this should be seen as a bridge - we should be pushing events to AWS, not running these actions on GHA. The latest run took ~50 minutes , and the data eventually ended up on AWS anyway, so the best practice is probably to push an event, after which everything takes place in AWS.

That said, for now we can use this as a crutch until the infra has been posted.

PR #476, opened 2021-08-06
Removes relative python imports

NatHillardUSDS:
Moving from relative imports to absolute to enable `poetry run python …data-pipeline/application.py [command]`. Relative imports do not work if running the file directly. Still preferred to run `poetry run download_census`, which should always work, but we should support this alternative way of calling it also


esfoobar-usds (approved):
🍏 looks good but I will try both docker and poetry runs with this change and see if there are any issues

Issue #459, opened 2021-08-04
[SPIKE] Encountered spec/tech difference on ExploreTool sprint 4: Tooltip component


vim-usds:
The design spec requires or suggests that the tool tip look like this when hovering over the info icon of the map legend:

![Screen Shot 2021-08-04 at 3.23.33 PM.png](https://images.zenhubusercontent.com/60d4d92c1bf5e498d6e808ba/a77284ff-7eea-4769-89a8-5767e96c1e67)

The Tooltip from `@trussworks/react-uswds` was investigated for its viability to adhere or meet this design spec.

[...]

PR #481, opened 2021-08-06
Modifies ExploreTool page to match sprint 4 design

### LINKED ISSUE #350: Update the Explore the tool page to reflect Sprint 4 design and content (excluding side panel).
**Description**
Updates to the text, layout (for example, color key and download link locations), and where text is shown (for example, CGG details in side panel vs. pop up) have been made to the Explore the tool page design to improve usability and understandability. Some tweaks to the content and bug fixes, etc. might still be needed later, but we should update the site to reflect this latest set of updates.

Side panel work is being done in a separate issue.

@janelien to confirm when designs for Sprint 4 are good for devs to start using.

**Tasks**
- [ ] Jane finalize for Sprint 4 Explore the tool designs and confirm 
- [ ] Update layout on Explore the tool page
- [ ] Update content on Explore the tool page

**Definition of "Done"**
Content and layout on site matches Sprint 4 designs

vim-usds:
closes #350 

This PR is using a placeholder tooltip using trussworks. There's a [spike](https://github.com/usds/justice40-tool/issues/459) to address the tooltip.


client/src/components/J40Alert/index.tsx
```diff
@@ -3,7 +3,12 @@ import {useIntl} from 'gatsby-plugin-intl';
 import {defineMessages} from 'react-intl';
 import * as styles from './j40Alert.module.scss';
 
-const J40Alert = () => {
+interface IJ40AlertProps {
+  location: Location;
+}
+
+const J40Alert = ({location}:IJ40AlertProps) => {
+  const isPadded = location?.pathname.match(/cejst\/?/);
```
NatHillardUSDS, line 11: We should avoid relying on `location` in this way - components should be independent of their context of use, and the mapping from location to page is not guaranteed to be persistent / introduces a hidden dependency. 

If a given page needs to do this just on the page, we should introduce a wrapper and include this as part of the styling of the page

client/src/components/J40Header.tsx
```diff
@@ -132,7 +133,7 @@ const J40Header = ({location}:IJ40HeaderProps) => {
         </span>
         <br/>
       </Alert>
-      {!isMethodologyPage && <Alert
+      {(!isMethodologyPage && !isExploreToolPage) && <Alert
```
NatHillardUSDS, line 136: Another place to avoid matching on `location` - we should look to other solutions here

client/src/components/MapLegend/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <div ref={ref} className={styles.infoIconWrapper} {...tooltipProps}>
+          {children}
+        </div>
+      ),
+  );
+  IconWrapper.displayName = 'custom info wrapper';
+
+  return (
+    <div className={styles.legendContainer}>
+      <h3 className={styles.legendHeader}>COLOR KEY</h3>
```
NatHillardUSDS, line 90: should localize this

client/src/components/MapLegend/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+    thresholdLabel: {
+      id: 'legend.info.threshold.label',
+      defaultMessage: constants.THRESHOLD_COMMUNITY,
+      description: 'the label of the threshold community legend',
+    },
+  });
+
+  // Type definitions required for @trussworks tooltip. This type defines the div that wraps the icon.
+  // This allows to pass children and other attributes.
+  type IconWrapperProps = React.PropsWithChildren<{
```
NatHillardUSDS, line 74: We can't just do 
```js
const text = "Communities that have cumulative 
            index score of Xth percentile
            and above"
<Tooltip position="left" label={text}>
      {text}
    </Tooltip>
``` ?

Comment (NatHillardUSDS) [reply client/src/components/MapLegend/index.tsx:74]: more examples [here](https://github.com/trussworks/react-uswds/blob/main/src/components/Tooltip/Tooltop.stories.tsx)

File: client/src/data/constants.tsx [hunk 5]
```diff
@@ -114,4 +115,9 @@ export const SCORE_BOUNDARY_LOW = 0.0;
 export const SCORE_BOUNDARY_THRESHOLD = 0.6;
 export const SCORE_BOUNDARY_PRIORITIZED = 0.75;
 
+
+// Explore the Tool page Map Legend
+export const PRIORITIZED_COMMUNITY = 'Prioritized community';
+export const THRESHOLD_COMMUNITY = 'Threshold community';
```
NatHillardUSDS, line 121: Shouldn't these be localized as well?

NatHillardUSDS (changes requested):
Overall like the tests and the new folder structure; called out a few things to be localized and a few places where we should avoid referring to `location` though; these would be good to fix so as not to set this pattern for the future

client/src/components/J40Alert/index.tsx
vim-usds: Good call! replacing this with an inversion of control pattern allowing the instance of the component determine its styling.

client/src/components/MapLegend/index.tsx
vim-usds: I don't think we can. I'm sorry Nat, I know you've brought this up a few times, but I'm not seeing how to use the default version. Thanks for being so patient! 

This code renders this:

Screen Shot 2021-08-06 at 3 07 25 PM

My current understanding leads me to believe that there are a few issues with this approach:

1. It seems to me that the target of the tooltip is the default blue button (shown above), in our case we want it to be the icon (something we set). Do you know how to change what's targeted using the default case you've mentioned? Placing the icon as a child of `<Tooltip>` still renders the button.  I believe that's why they have a custom tooltip component for this very reason. But maybe I'm missing something? If my understanding is correct, this is a deal breaker. 
2. Assuming bullet 1 is correct, having the label be a `text` variable doesn't allow us to control its width, which gets cut off on mobile. The tooltip doesn't seem to scale for mobile. Might be some lower level uswds classes that could fix this, however bullet 1 still dominates.

Looking forward to hearing your thoughts!

NatHillardUSDS: ah ok, I had misunderstood what was what here, thank you for the explanation -- I was thinking we could just treat the image as a button, which in this case it effectively is (and in fact for accessibility reasons we will need it to be), but didn't realize it came along with this other button-styling baggage. 

Looks like this way works to wrap the `img` in an `a` and do the right thing from an accessibility perspective

client/src/data/constants.tsx
vim-usds: maybe?

The thinking was that both constants, i.e.,  `*_COMMUNITY` are being called in two places:

1. Here:
https://github.com/usds/justice40-tool/blob/3e8f12ebd57f25433bb144ac24a288b5aec57989/client/src/components/MapLegend/index.tsx#L62

2: In the tests:
https://github.com/usds/justice40-tool/blob/3e8f12ebd57f25433bb144ac24a288b5aec57989/client/src/components/MapLegend/tests/mapLegend.test.tsx#L17

Localization is occurring in both places. if these constants ever change the localization will change as well because of where they are used maybe?

Also since localization is using hooks (useIntl) they may require to be wrapped in functions.

client/src/components/J40Header.tsx
vim-usds: You mentioned this above:

> We should avoid relying on location in this way - components should be independent of their context of use, and the mapping from location to page is not guaranteed to be persistent / introduces a hidden dependency.
> 
> If a given page needs to do this just on the page, we should introduce a wrapper and include this as part of the styling of the page

Agree with the sentiment and would like to validate implementation:

1. Move Alerts out of J40header and wrap into `<AlertWrapper />`
3. Call `<AlertWrapper />` on each page
4. Use page styling to display or not display each Alert via `<AlertWrapper />`'s props ?

NatHillardUSDS: The location has up to now been used to get flags, which are independent of the specific page they’re used on - this is a different use case than styling a component based on location — worth looking into whether there are other ways to pass the location for flag purposes, but the gatsby context by default only passes it to pages.

client/src/data/constants.tsx
```diff
@@ -2,6 +2,7 @@ import {LngLatBoundsLike} from 'maplibre-gl';
 import {isMobile as isMobileReactDeviceDetect} from 'react-device-detect';
 
 // URLS
+export const DOWNLOAD_ZIP_URL = 'https://justice40-data.s3.amazonaws.com/Score/usa.zip';
```
NatHillardUSDS, line 5: As of just now, you can use this URL to get the real thing: `https://justice40-data.s3.amazonaws.com/data-pipeline/data/score/downloadable/Screening+Tool+Data.zip`

client/src/data/constants.tsx
NatHillardUSDS: Hm, interesting - I wonder if we can call`defineMessages` in this context, without a hook? I'm thinking something like:
```js
 const messages = defineMessages({
    prioritizedCommunity: {
      id: '...,
      defaultMessage: '...',
      description: '...',
    },
```
then at the calling context:
```
intl.formatMessage(constants.messages.prioritizedCommunity)
```
something like this?

seems constants that are localized are a common use case, I think

vim-usds: Totally agreed!

client/cypress/e2e/downloadPacket.spec.js
```diff
@@ -2,20 +2,22 @@
...
+    //       body: 'success',
+    //       headers: {
+    //         'Content-Type': 'text/html; charset=utf-8',
+    //         'Content-Disposition': 'attachment',
+    //       },
+    //     },
+    // ).as('downloadRequest');
+    // cy.get('button[class*="downloadPacket"]').click();
+    // cy.wait('@downloadRequest');
+    // cy.readFile(`cypress/downloads/${filename}`).should('exist');
```
NatHillardUSDS, line 21: Now that we have https://justice40-data.s3.amazonaws.com/data-pipeline/data/score/downloadable/Screening+Tool+Data.zip , can we add back this test?

client/src/components/HowYouCanHelp/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+            <Link to={'/methodology'}>Data & methodology</Link>
+            {` `}
+            and send us feedback.</div>
+        </li>
+        <li className={styles.howYouCanHelpList}>
+          <div className={styles.howYouCanHelpText}>Find your community and
+            {` `}
+            <a href={'mailto:screeningtool.feedback@usds.gov'}>share your feedback</a>
+            {` `}
+            .</div>
```
NatHillardUSDS, line 41: Can we add localization ?

NatHillardUSDS (changes requested):
Nice! Overall looking good - with the remaining localization on the contact page and adding back the test should be good to go!

client/cypress/e2e/downloadPacket.spec.js
vim-usds: Not yet. The download packet component was removed and will be placed on the methodology page. There is a ticket for that. Once that ticket is started, this will code will be reinstated.

client/src/components/J40Header.tsx
vim-usds: just removed location from the two aforementioned places!

client/src/components/HowYouCanHelp/index.tsx
vim-usds: added!

NatHillardUSDS (approved):
Nice! LGTM

PR #484, opened 2021-08-06
'NoneType' object is not subscriptable because of missing census data

NatHillardUSDS:
Adding back census to list of potential datasets, but separating out from standard list

Error this addresses:
  File "/Users/lucas/Documents/usds/repos/justice40-tool/data/data-pipeline/data_pipeline/etl/runner.py", line 71, in etl_runner
    f"data_pipeline.etl.sources.{dataset['module_dir']}.etl"
TypeError: 'NoneType' object is not subscriptable


esfoobar-usds (approved):
🍏

Issue #363, opened 2021-07-15
As a FE developer, I want consistency in page layout across the app, so that its clear how to make future sizing decisions


NatHillardUSDS:
**Description**
Right now, we're using a mix of the USWDS [layout grid](https://designsystem.digital.gov/utilities/layout-grid/), [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/), and raw px/em values. 

**Solution**
We should choose a system or systems, make existing page layouts consistent, and develop guidelines for future usage. 

**Describe alternatives you've considered**
- As things stand, it's not clear how to style a new page. 

**Tasks**
- [ ] Audit existing page styling mechanisms
- [ ] Based on this, and best practices, write out guidelines for how to layout elements on the page, and store these in a readme
- [ ] Make necessary changes to bring our page layout in line with these guidelines.
- [ ] Fix header
- [ ] Fix footer (see #564)
- [ ] FIx main content
- [ ] Fix "USA header"
- [ ] Fix Alerts (see #562)

**Definition of "Done"**
We have written out guidelines for how to layout pages and made our page layouts consistent with these.

switzersc-usds:
Note that Tom has started tackling some of this. Vim and Tom will touch base on more specifics and knock out the little things that are left.

switzersc-usds:
This may be done. Any outstanding issues are probably addressed in #571. I'm closing this for now and if we find specific layout issues we will create new bug issues for those.

PR #486, opened 2021-08-07
Refactor layout

TomNUSDS:
Goal was to move `<DatasetContainer>` out of `layout.tsx` and down into `methodology.tsx` page. Required a bit of refactoring.


client/src/pages/404.tsx
```diff
@@ -1,21 +1,26 @@
...
-const pageStyles = {
-  color: '#232129',
-  padding: '96px',
-  fontFamily: '-apple-system, Roboto, sans-serif, serif',
-};
+// const pageStyles = {
+//   color: '#232129',
+//   padding: '96px',
+//   fontFamily: '-apple-system, Roboto, sans-serif, serif',
+// };
```
NatHillardUSDS, line 12: Can we remove this commented-out bit?

client/src/components/J40MainGridContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children: ReactNode,
+  fullWidth?: boolean,
+  className?: string
+}
+
+const J40MainGridContainer = ({
+  children,
+  fullWidth = false,
+  className = 'j40-grid-container'}: ILayoutProps) => {
+  return !fullWidth ? (
```
NatHillardUSDS, line 16: Had to think a bit about the `!` - Maybe flipping this would make it a bit easier to follow? 

e.g. 
```js
fullWidth ? <div>... : <GridContainer>
```

NatHillardUSDS (changes requested):
Hm, I would wonder overall what other pages do here - I'd think it should be possible in theory to have the Layout control a top-level grid, though as we've discussed this seems a bit difficult to do in practice -- though this seems to satisfy our needs for the moment - thank you for taking this on! I would see if we could revisit at some point (perhaps as part of #363 ?) 

The only thing I would say is that we should remove that commented-out bit, if we take that out then good to merge from me!

client/src/pages/404.tsx
TomNUSDS: Good catch. Done

client/src/components/J40MainGridContainer.tsx
TomNUSDS: Fair. I debated flipping it. Done.

NatHillardUSDS (approved):
LGTM!

PR #487, opened 2021-08-09
Check in VSCode config for easier local debug

### LINKED ISSUE #466: Task: Check in VSCode config for easier local debug

NatHillardUSDS:
Fixes #466 - Check in VSCode config for easier local debug
Makes it easier to run local debug through VSCode - these files aren't typically checked in on a project, but with this project things are a bit trickier given the use of click (which is a great library but makes testing a bit more difficult because of its use of `console_scripts`)

### DIFF
--- a/data/data-pipeline/.vscode/launch.json
+++ b/data/data-pipeline/.vscode/launch.json
@@ -1,0 +1,71 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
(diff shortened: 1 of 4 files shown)


esfoobar-usds (approved):
🍏

PR #489, opened 2021-08-09
Add Cumulative Score "steps" section to methodology page

TomNUSDS:
Layout was non-standard USWDS, so lots of custom scss.

This is phase one. Localization is not yet done. It's actually easier to review with English in place.


client/src/styles/global.scss
```diff
@@ -329,3 +331,89 @@ $primary-color: #112f4e;
...
+/* the > is for child only syntax in css. This is required so that lists that are nested under the
+      list are not affected (e.g. <ul><li><p><ul><li>not styled</li></ul></p></li><ul>*/
+.j40-process-list-wrapper > {
+  $j40-steps-list-color: #002D3F;
+  ul {
+    margin-top: 3rem;
+    margin-bottom: 1rem;
+    padding-inline-start: 0;
+
+    > li {
```
TomNUSDS, line 345: This `>` syntax means to only apply the style one child element deep. This is required because there are `ul` lists _inside_ the content of these items (see `Data inputs` section).

Bonus, it prevent styles from bleeding down into child elements.

client/src/styles/global.scss
```diff
@@ -329,3 +331,89 @@ $primary-color: #112f4e;
...
+    > li:last-child {
+      /* no line on last item */
+      border-left: .2rem solid transparent;
+      margin-top: -1.9rem;
+    }
+  }
+}
+
+/* The math equation on the methodology page.. which is a royal pain */
+.j40-math-division-container {
```
vim-usds, line 377: Looks great!

It seems we have two ways to style in this project:

1. Place styles of the component in the global sass file
2. Place styles for the component in the component.module.scss file

How should we determine when to use which?

I haven't been using global sass file too much but I am starting to. I am using global sass file for styles that exist on more than one page (for example the h1 styling for each page) and need to shared across thee app. I have been using components module styling for all component based styling.

client/src/styles/global.scss
```diff
@@ -135,7 +136,7 @@ $primary-color: #112f4e;
       float: left; // allows vertical centering on logo and text
       margin-right: 0.5em; // space between logo and text
```
vim-usds, line 137: When are we electing to you use em over rems? I've been using rems pretty much throughout

client/src/components/scoreStepsList.tsx
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
+import {GridContainer, Grid} from '@trussworks/react-uswds';
+
+const ScoreStepsList = () => {
```
vim-usds, line 4: Is the TimeLinePage in the `pages` directory  related to this effort or is that something separate?

client/src/styles/global.scss
TomNUSDS: Actually, it's a bug. I'll flip them all to `rem`.

client/src/styles/global.scss
TomNUSDS: Yeah. I started to go down this path, but the scss is so complicated as it is that I was saving abstractions like localizations and componentizing/typescript names (for scss) for a second pass. Probably a good idea to generate a bug so these tasks don't get lost.

My personal opinion, but I'm still on the fence about turning one file `scoreStepsList.jsx`

client/src/styles/global.scss
TomNUSDS: fixed in latest.

vim-usds:
Thanks for your comments.

Do you think adding unit and/or e2e tests for this makes sense?

client/src/styles/global.scss
vim-usds: Might be worthwhile to discuss the complexity you mention here at some point. I'm thinking that most of the component specific scss written in the `global` file could theoretically (might be wrong here) be placed in a module no?

TomNUSDS: Yeah. It's a tradeoff between the complexity of having two additional files and extra layers of indirection in the project for scoping versus using the scoping built into scss and css. It probably has more to do with me being more comfortable with scss than with generating scss typescript interfaces to replace classname strings.

I'm happy to convert it over, I'd just prefer to make sure what we have is OK with design because, when there's extra levels of indirection for styles, I find big edits are trickier. (react -> typescript interface names -> scss vs react -> scss).

TomNUSDS:
Some kind of tests wouldn't be a horrible idea. Are we moving everything from jest to cypress or is it ok to do jest tests?

client/src/styles/global.scss
vim-usds: agreed

vim-usds (approved):
Looks really nice on mobile!

vim-usds (approved):
I created an issue where we discuss what we'd like to be standardized on future components. 

https://github.com/usds/justice40-tool/issues/502

Thanks!

PR #491, opened 2021-08-09
Tile-baking columns with floating rounds completed

### LINKED ISSUE #287: As an EVCM, I want to have only relevant information shown to me in the block group detail view, so that I'm not confused by too much information (Backend only sends relevant info).
**Description**
We only want to send data elements to the frontend that the frontend design needs.

**Solution**
Update data-pipeline to generate GeoJSON and tiles that have only data elements we want in the sidebar.  Also data should include property names that are user readable; this could be a column that is distinct from the raw/machine-readable property name.

- The current full list of properties (almost 100) feels overwhelming

**Tasks**
- [x] Work with Product to define the relevant list of properties to include (and exclude) 
- [ ] Work with design to devise a good column label and formatting of data within that column 
- [ ] Create a filter on our CBG detail view that only shows these properties

**Definition of "Done"**
- [ ] The CBG detail view only shows the chosen properties

esfoobar-usds:
Fixes #287


data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -165,16 +177,23 @@ def transform(self) -> None:
...
 
         # list the null score cbgs
         null_cbg_df = merged_df[merged_df["Score E (percentile)"].isnull()]
 
         # subsctract data sets
-        removed_df = pd.concat([merged_df, null_cbg_df, null_cbg_df]).drop_duplicates(
-            keep=False
-        )
+        removed_df = pd.concat(
+            [merged_df, null_cbg_df, null_cbg_df]
```
NatHillardUSDS, line 195: Noticed this earlier today, should there by two `null_cbg_df`s?

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -188,9 +207,23 @@ def _save_full_csv(self):
...
+            "Score E (percentile)",
+            "Score E (top 25th percentile)",
+            "Poverty (Less than 200% of federal poverty line)",
+            "Percent individuals age 25 or over with less than high school degree",
+            "Linguistic isolation (percent)",
+            "Unemployed civilians (percent)",
+            "Housing burden (percent)",
+        ]
+        decimals = pd.Series(
+            [2, 2, 2, 2, 2, 2, 2], index=TILES_SCORE_FLOAT_COLUMNS
```
NatHillardUSDS, line 223: hm, this is a bit confusing - is it possible to do this as a map?

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -41,16 +43,20 @@ def __init__(self):
...
             "Score E (top 25th percentile)",
-            "GEOID",
-            "State Abbreviation",
-            "County Name",
+            "Poverty (Less than 200% of federal poverty line)",
+            "Percent individuals age 25 or over with less than high school degree",
+            "Linguistic isolation (percent)",
+            "Unemployed civilians (percent)",
+            "Housing burden (percent)",
         ]
```
NatHillardUSDS, line 56: Geoplatform had brought up that the length of the column names themselves can be expensive to write out; might be interesting to see if shrinking the names of the columns has a big impact on tile size?

NatHillardUSDS: If so, we would need some sort of legend on disk that we could consult, mapping shortened name to expanded, or something like this

NatHillardUSDS (commented):
A few requests / questions

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
esfoobar-usds: Per our conversation on zoom, we will punt this as the binaries don't have props like Geojson files do.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
esfoobar-usds: Will leave a comment on the approach in-line.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
esfoobar-usds: Per our zoom convo, will make a constant at the top of the file with the float precision and generate the list using a generator (i.e. `[e] * n`)

NatHillardUSDS (dismissed):
LGTM!

NatHillardUSDS (approved):
Nice, LGTM!

Issue #495, opened 2021-08-10
Task: Create FE precommit hook


vim-usds:
## Describe the task
A few things that should be run on each commit as a pre-commit:

1. prettier
2. tests
3. cypress
4. localizing

PR #498, opened 2021-08-10
updates snapshot

vim-usds:
A components snapshot wasn't updated. A few precommit hooks should be added. It has been captured here #495


TomNUSDS (approved):
lgtm.

PR #503, opened 2021-08-11
adding layer to mvts

esfoobar-usds:


NatHillardUSDS (approved):
Might be helpful to add a comment as part of this PR itself (ie not a code change) about why this was necessary — in the future having an associated issue ticket might also give context. Overall goal is to be able to say “why was this necessary?” When looking through git history

vim-usds:
> @vim-usds can you add the context here of why we needed this change?

sure thing @esfoobar-usds, 

We have a number of constants that correspond to various properties of the geojson file (a json file that is the result of decoding a pbf file) on the FE. The source `layer` is one of those properties. The layer constant was expecting the geojson layer value to be set as `blocks` for both the high and low layers. When low tile's geojson had the layer property set to something else (not blocks), it resulted in an error. See client/VIEW_MAP_DATA.md for more details.

There were some other issues as well around cache invalidation and permissions which have been resolved. The app is able to get the data! Thank you!

esfoobar-usds:
Thanks @vim-usds. Merging this now.

PR #509, opened 2021-08-12
Data Unit Tests

### LINKED ISSUE #341: [SPIKE] As a J40 developer, I want to write Unit Tests for etl_score_post, so that tests are run on each commit.

NatHillardUSDS:
Fixes #341 -
As a J40 developer, I want to write Unit Tests for the ETL files, so that tests are run on each commit


NatHillardUSDS:
Added Load tests just now using `tmp_path_factory` as my final push! - we may want to switch over to using [pyfakefs](https://github.com/jmcgeheeiv/pyfakefs) at some point as it seems overall easier, but I couldn't quite get this to work. 

Looks like there's a conflict with main still but this should be straightforward - happy to pair on this if/when we can find a time!

data/data-pipeline/data_pipeline/etl/score/tests/test_score_post.py
```diff
@@ -0,0 +1,113 @@
...
+    )
+    pdt.assert_frame_equal(
+        score_data_actual,
+        score_data_expected,
+    )
+
+
+def test_create_tile_data(etl, score_data_expected, tile_data_expected):
+    output_tiles_df_actual = etl._create_tile_data(score_data_expected)
+    pdt.assert_frame_equal(
```
vim-usds, line 76: is this where we are testing if CBG feature data (poverty, housing, education, etc) exists?

vim-usds (commented):
Looks great, small question on testing if features properties exist or not makes sense?

data/data-pipeline/data_pipeline/etl/score/tests/test_score_post.py
NatHillardUSDS: Right - this particular line consults `score_data_expected`, which is drawn automatically from `conftest.py` in in the `score_data_expected` fixture, which in turn gets it from `snapshots/score_data_expected.pkl` . This is a "pickled" data structure, so you can load it into a python interpreter and print it out to verify. 

We aren't yet using `pytest-snapshot`, which works like jest snapshot, as I didn't want to add another dependency just yet, and plus its a bit wonky with pandas dataframes

vim-usds: Thanks! Kudos to function naming which allows a wanna-be pythonista (me) to actually follow along!

data/data-pipeline/data_pipeline/etl/score/tests/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+    os.environ["PYTEST_DEBUG_TEMPROOT"] = str(
+        basetemp
+    )  # this sets the location of the temp directory inside the project folder
+    basetemp.mkdir(parents=True, exist_ok=True)
+    root = tmp_path_factory.mktemp("root", numbered=False)
+    return root
+
+
+@pytest.fixture(autouse=True)
+def settings_override(monkeypatch, root):
```
switzersc-usds, line 28: Rather than monkey-patching, how about setting dir/path/file names as env variables and having an `.env.test` that sets the path to the tmp path? Then we wouldn't need a lot of this code and we'd have flexibility for changing path names later or in other envs.

data/data-pipeline/data_pipeline/etl/score/tests/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+            "State Code": pd.Series(["01", "02", "04"], dtype="string"),
+            "State Name": pd.Series(["Alabama", "Alaska", "Arizona"], dtype="object"),
+            "State Abbreviation": pd.Series(["AL", "AK", "AZ"], dtype="string"),
+        },
+    )
+
+
+@pytest.fixture()
+def score_transformed_expected():
+    return pd.read_pickle(pytest.SNAPSHOT_DIR / "score_transformed_expected.pkl")
```
switzersc-usds, line 94: I'm not sure about using pickle. It doesn't seem to me that we have files/fixtures big enough to warrant translating to/from binary, and it obfuscates the expected output such that if you wanted to troubleshoot you'd have to take an extra step (or more?) to see what's going on. Do we get anything from this other than some (arguably slight) performance or memory gains?

switzersc-usds (changes requested):
Hey Nat! Sorry for the delayed review here. Thanks for getting us started on the tests and refactor for the score work! Overall this looks like a good foundation for the Score ETL unit tests and I appreciate the refactoring for the constants as well. 

There are a couple typo fixes and updates I'd like to make to the README, so I'll just add a commit for those. I'm also going to merge main and resolve the conflicts at the same time and then push to update this PR.

Otherwise, I've left a couple comments about some higher level decisions -- monkeypatching vs using env variables, and using pickles vs not. To me, both (but especially the latter) boil down to readability/simplicity and ease of contribution for other devs, especially people new to python or to the project. I'm thinking about how easy it will be to replicate this for the other Score ETL classes and I think pickling puts up both an implementation and a cognitive barrier. 

If it makes sense, let's sync early this week if you have time! Otherwise let me know what you think and then we can figure out where to take this from here. Thanks again for taking this on :)

data/data-pipeline/data_pipeline/etl/score/tests/conftest.py
switzersc-usds: Sounds like maybe this is python/pytest convention, so I don't have a strong opinion here, although I do still prefer env variables over monkeypatching to set attrs that are env-related ^.^

data/data-pipeline/data_pipeline/etl/score/schemas.py
```diff
@@ -0,0 +1,135 @@
...
+class StateInfo:
+    fips: str
+    state_name: str
+    state_abbreviation: str
+    region: str
+    division: str
+
+
+@dataclass
+class Score:
```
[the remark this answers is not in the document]
switzersc-usds: Actually, some of these might be floats.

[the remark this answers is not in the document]
NatHillardUSDS: Yeah, so actually this file is not currently in use and can be removed. This was part of an idea that we could start using a local schema to validate. I have an example of using this in another repo that I can share offline. I think actually we could and should use [pandera](https://pandera.readthedocs.io/en/stable/schema_models.html?highlight=inputschema) to create a schema that we can annotate input/output datasets with - a major problem we see is with typing, for instance, which this can ensure for us. Additionally it can put certain sanity checks in place, such as making sure percentiles are between 0-100 etc.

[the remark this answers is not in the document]
switzersc-usds: Talked sync - we'll look into this in the future but for now remove since we're not using

data/data-pipeline/data_pipeline/etl/score/tests/conftest.py
switzersc-usds: Ideally this is done with dynaconf and env settings. Right now that's not working because we're building these values from variables or constants instead of setting them in settings.toml. We should try setting these in settings.toml. We'll do this in a future test PR (since this works as is and isn't too scary ^.^).

Note that we're calling `reload` in these methods too and that's something we shouldn't have to do (it's hacky!).

data/data-pipeline/data_pipeline/etl/score/tests/conftest.py
switzersc-usds: Learned more synchronously: We're using these to preserve types in the dataframe so that we can accurately compare those. There may be other ways to solve this problem that are clearer than pickles (eg loading from sqlite or using tiangolo/sqlmodel or using snapshots; see comment below).

switzersc-usds: Sounds like we need pickles for now, so let's add some docs and a class method to create the pickles. Let's at least add to the README for this PR. We'll create a new issue for spiking on using pytest-snapshot.

switzersc-usds (approved):
Woohoo! Thanks for the updates! 🎉

switzersc-usds:
I will resolve the conflicts and merge!

Issue #352, opened 2021-07-13
Update the Contact page to match Sprint 4 design


katherinedm-usds:
**Description**
We have updated content and layout change to the Content page as part of overall improving site usability and understandability. There might still be content tweaks later, but the Contact page should be updated to reflect the latest.

**Links to user research or other resources**
@janelien confirm Sprint 4 design mockups are ready for devs to use

**Tasks**
- [ ] Jane confirm Sprint 4 design mockups are ready for devs to use
- [ ] Dev update Contact content
- [ ] Dev remove Get Involved on the side - this is treated differently now in the About page design

**Definition of "Done"**
Site matches Sprint 4 design.

janelien:
latest contact page designs

![J40_Sprint 4_Contact.png](https://images.zenhubusercontent.com/60998bc7b85eebae77bd2048/fcd3211b-4517-4094-a707-e1ea156462fc)

![J40_sprint4_mobile_contact.png](https://images.zenhubusercontent.com/60998bc7b85eebae77bd2048/5f22735b-9072-4ead-aacc-3970c5ff81ad)

BethMattern:
We decided to have only one email address for this page and it has been created - screeningtool.feedback@usds.gov

switzersc-usds:
Jane and Tom will discuss this today

janelien:
Could I please get a link to where I should look? I was looking here: https://d2zjid6n5ja2pt.cloudfront.net/justice40-tool/main/en/contact/ but it doesn't have this particular update (two emails are listed).

janelien:
Looks good now.

PR #511, opened 2021-08-12
fix contact fix footer

TomNUSDS:
Minor tweaks so we can close some issues.

See: #220 #352

### DIFF
--- a/client/src/pages/contact.tsx
+++ b/client/src/pages/contact.tsx
@@ -12,18 +12,17 @@ interface ContactPageProps {
 
 const ContactPage = ({location}: ContactPageProps) => {
   const generalemail = 'screeningtool.feedback@usds.gov';
-  const techemail = 'screeningtool.support@usds.gov';
 
   return (
     <Layout location={location}>
 
       <J40MainGridContainer fullWidth={true}>
-        <AlertWrapper />
+        <AlertWrapper/>
       </J40MainGridContainer>
 
       <J40MainGridContainer>
-        <Grid row><Grid col>
-          <section className={'usa-prose'}>
+        <section className={'usa-prose'}>
+          <Grid row><Grid col>
             <h2><FormattedMessage
               id={'contact.pageheader'}
               description={'H2 header for contact page'}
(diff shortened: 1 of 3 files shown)


client/src/pages/contact.tsx
```diff
@@ -12,18 +12,17 @@ interface ContactPageProps {
 
 const ContactPage = ({location}: ContactPageProps) => {
   const generalemail = 'screeningtool.feedback@usds.gov';
```
vim-usds, line 14: Can these two be camelCased?

client/src/pages/contact.tsx
```diff
@@ -12,18 +12,17 @@ interface ContactPageProps {
...
 
       <J40MainGridContainer fullWidth={true}>
-        <AlertWrapper />
+        <AlertWrapper/>
       </J40MainGridContainer>
 
       <J40MainGridContainer>
-        <Grid row><Grid col>
-          <section className={'usa-prose'}>
+        <section className={'usa-prose'}>
```
vim-usds, line 24: On some of the pages I think I remember seeing the pattern of Container > row > col > section. Is there something special here that we have to break from that pattern?

vim-usds (changes requested):
Looks great! Just a few minor questions.

client/src/pages/contact.tsx
TomNUSDS: Yeah. I'm not sure the correct approach here.  HTML spec says: 

> The <section> HTML element represents a generic standalone section of a document, which doesn't have a more specific semantic element to represent it. Sections should always have a heading, with very few exceptions.

So, it's kind of like a higher level `<h#/><p/>` part of the document?

It sounds like it's a fallback for html5 if the content doesn't fit nicely into `<article>`, `<aside>`, `<main>`, `<nav>`, etc.

One challenge of using Trussworks react library is that I'm not aware of if/where it uses html5 tags.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

Not sure how close we're sticking to html5 tags in the rest of our page output?

client/src/pages/contact.tsx
TomNUSDS: Good catch, it should be. (again, seems like linter should be catching this?)

TomNUSDS: fixed

client/src/pages/contact.tsx
TomNUSDS: Removed `<section>` since it's already in a `<main>`

TomNUSDS:
Re-review when you have time?

vim-usds (approved):
LGTM

PR #525, opened 2021-08-13
sets column as percentiles

vim-usds:
Hello @BethMattern, 

I think you've already validated this in Slack, however just to be sure. Can you ensure that we've selected the correct columns from the CSV files to show percentile in the map's side panel when a CBG is selected?

Thank you!


esfoobar-usds (approved):
🍏

PR #528, opened 2021-08-16
Initial draft for data provenance addition to README

switzersc-usds:
We want to make the data usable/available at every step of our data
pipeline. This starts te addition to the README that spells out the data
provenance and where each version of the data as it goes through our
pipeline lives.


esfoobar-usds (approved):
🍏

PR #533, opened 2021-08-16
update sidepanel d3 cdn

vim-usds:
testing deployment AWS creds error

### DIFF
--- a/client/src/components/DatasetContainer/tests/__snapshots__/datasetContainer.test.tsx.snap
+++ b/client/src/components/DatasetContainer/tests/__snapshots__/datasetContainer.test.tsx.snap
@@ -11,7 +11,9 @@ exports[`rendering of the DatasetContainer checks if various text fields are vis
       <div
         class="grid-container-desktop-lg"
       >
-        <div>
+        <div
+          class="undefined undefined"
+        >
           Limited data sources — Datasets may be added, updated, or removed.
         </div>
       </div>
--- a/client/src/components/J40Alert/tests/__snapshots__/j40Alert.test.tsx.snap
+++ b/client/src/components/J40Alert/tests/__snapshots__/j40Alert.test.tsx.snap
@@ -1,8 +1,20 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`rendering of the J40Alert tests the rendering of J40Alert 1`] = `
+exports[`rendering of the J40Alert tests the rendering of J40Alert with padding 1`] = `
 <DocumentFragment>
-  <div>
+  <div
+    class="undefined undefined"
+  >
+    Limited data sources — Datasets may be added, updated, or removed.
+  </div>
+</DocumentFragment>
+`;
+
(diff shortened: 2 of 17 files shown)


client/src/components/DatasetContainer/tests/__snapshots__/datasetContainer.test.tsx.snap
```diff
@@ -11,7 +11,9 @@ exports[`rendering of the DatasetContainer checks if various text fields are vis
       <div
         class="grid-container-desktop-lg"
       >
-        <div>
+        <div
+          class="undefined undefined"
```
TomNUSDS, line 15: That the snapshot has `"undefined undefined"` are an indicator that's something wrong. I don't know if it's part of this changelist but it's something to investigate?

client/src/components/J40Alert/tests/__snapshots__/j40Alert.test.tsx.snap
```diff
@@ -1,8 +1,20 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`rendering of the J40Alert tests the rendering of J40Alert 1`] = `
+exports[`rendering of the J40Alert tests the rendering of J40Alert with padding 1`] = `
 <DocumentFragment>
-  <div>
+  <div
+    class="undefined undefined"
```
TomNUSDS, line 6: `"undefined undefined"` again.. same code generating?

client/src/data/constants.tsx
```diff
@@ -4,21 +4,28 @@ import {defineMessages} from 'react-intl';
 
 // URLS
 export const DOWNLOAD_ZIP_URL = 'https://justice40-data.s3.amazonaws.com/data-pipeline/data/score/downloadable/Screening+Tool+Data.zip';
-export const FEATURE_TILE_BASE_URL = 'https://d2zjid6n5ja2pt.cloudfront.net';
+
+// This is the URL for the CDN that delivers the tile data
+export const FEATURE_TILE_BASE_URL = 'https://d3jqyw10j8e7p9.cloudfront.net';
```
TomNUSDS, line 9: Why did this change from `d2zjid6n5ja2pt` to `d3jqyw10j8e7p9` and when will it change again?

TomNUSDS (approved):
some minor questions, but otherwise lgtm

client/src/components/J40Alert/tests/__snapshots__/j40Alert.test.tsx.snap
vim-usds: Yeah, this is actually the same PR as last time. Same issue has already been captured in a ticket

client/src/data/constants.tsx
vim-usds: Thanks Tom.. this is a good question. @esfoobar-usds ?

Not sure why it changed, but right now we're looking into the ramifications of this change. One of them being CORS. My guess is that we wanted to separate the client from the map data layer?

PR #544, opened 2021-08-16
About page sprint4

### LINKED ISSUE #348: Update About page to match Sprint 4 design and content
**Description**
The content for the About page has been updated to reduce the amount of text, and layout has been changed as well. The site should be updated accordingly.

@janelien confirm when designs for Sprint 4 are ready 

**Tasks**
- [ ] Update About layout
- [ ] Update About text
- [ ] (to confirm with Jane if this is the behavior, About page should be the landing page

**Definition of "Done"**
 - [ ] All content and layout updates as provided by Jane have been incorporated into the site. There may still be minor tweaks to content and bug fixes, etc. later.

TomNUSDS:
Brings the About page (aka index page) in line with the design spec.
Fixes #348 

TODO:
 - Need to convert string so they can be easily localized. This will be a second phase since it's easier to read/review before this conversion. (Added issue so we don't forget https://github.com/usds/justice40-tool/issues/545)


client/src/components/AboutCard/AboutCardsContainer.tsx
```diff
@@ -0,0 +1,17 @@
...
+import {Grid} from '@trussworks/react-uswds';
+
+type AboutCardsContainerProps = {
+  className?: string
+}
+
+// note: this assumes a J40MainGridContainer container!
+const AboutCardsContainer = (props: React.PropsWithChildren<AboutCardsContainerProps>) => {
+  return (
+    <Grid row gap={'lg'} className={'j40-aboutcard-container ' + (props.className || '')}>
```
vim-usds, line 11: In `(props.className || '')`, why is the `or empty` string needed? Seems that since the prop is optional, we may not need that? Not sure tho?

client/src/components/AboutCard/__snapshots__/AboutCard.test.tsx.snap
```diff
@@ -0,0 +1,56 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`rendering of the AboutCard checks if component renders 1`] = `
+<DocumentFragment>
+  <div
+    class="grid-gap-lg tablet:grid-col"
```
vim-usds, line 6: The advantage of using global classname is that these are defined in jest

client/src/pages/index.tsx
```diff
@@ -3,10 +3,29 @@ import {defineMessages} from 'react-intl';
 import {Grid} from '@trussworks/react-uswds';
 import {FormattedMessage, useIntl} from 'gatsby-plugin-intl';
 
-import AreasOfFocusList from '../components/areasOfFocusList';
-import AlertWrapper from '../components/AlertWrapper';
+// import AreasOfFocusList from '../components/areasOfFocusList';  // TODO: remove areasOfFocusList file
```
vim-usds, line 6: Should these be removed?

client/src/pages/index.tsx
```diff
@@ -54,151 +68,127 @@ const IndexPage = ({location}: IndexPageProps) => {
   return (
     <Layout location={location}>
 
-      <J40MainGridContainer fullWidth={true}>
-        <AlertWrapper />
-      </J40MainGridContainer>
+      <J40MainGridContainer className={'j40-about-page'}>
```
vim-usds, line 71: I'm guessing something like this might allow us to add the headers from the design?

```jsx
<J40MainGridContainer className={'j40-about-page'}>

   // Seems as though we can use the padding token from USWDS directly in-line 
   // (kinda like tailwind CSS!). After we get our h1 - hN defined from design 
   // we may not have to overide them like this. Just one of a hundred ways
   // to implement this headings

    <h2 className={'padding-top-3'}>How to get started</h2>
    <AboutCardsContainer>
    :
    :
 ```
 
 Might need to also reduce the about-card's top margin:
 ```css
   .j40-aboutcard-card {
    margin: 3rem 0 3rem 0;
 ```

File: client/src/styles/global.scss [hunk 5]
```diff
@@ -417,3 +417,72 @@ $primary-color: #112f4e;
...
+
+  .j40-about-section-divider {
+    max-width: 1px;
+    max-height: 100px;
+    border-left: 1px black solid;
+    border-top: 1px black solid;
+  }
+
+  .j40-about-cards.usa-card-group > .usa-card {
+    .usa-card__container {
```
vim-usds, line 445: kudos for figuring out how to use the USWDS cards styles! These overrides seem like such a pain!

client/src/pages/index.tsx
```diff
@@ -54,151 +68,127 @@ const IndexPage = ({location}: IndexPageProps) => {
   return (
     <Layout location={location}>
 
-      <J40MainGridContainer fullWidth={true}>
```
vim-usds, line 57: Has design removed the Alert on this page? Sprint 4 designs show them to exist. Not sure tho...

vim-usds (commented):
Really awesome componentizing and working with USWDS! 

Been so removed from this work that I'm envious! Thanks for allowing me a moment to dip back into what I love doing (which is creating react components!)

client/src/components/AboutCard/AboutCardsContainer.tsx
TomNUSDS: @NatHillardUSDS Can chime in, but I think without this `|| ''` you'll get `undefined`

client/src/components/AboutCard/__snapshots__/AboutCard.test.tsx.snap
TomNUSDS: Oh good point!

client/src/pages/index.tsx
TomNUSDS: Yes! I'll remove it and update the PR

client/src/pages/index.tsx
TomNUSDS: Good call.

client/src/styles/global.scss
TomNUSDS: Yeah, may not be the best approach? It's something we can refactor if we want.

client/src/components/AboutCard/AboutCardsContainer.tsx
vim-usds: if undefined is a type we can add it to the props definition as `string || undefined`

TomNUSDS: I think the html output literally has `undefined` in it. I think in most use cases `props.className` is typed as optional... (e.g. `className?: string`) it can be undefined.

client/src/pages/index.tsx
TomNUSDS: Oh, good catch. Re-added for now.

TomNUSDS:
Reopen to generate deploy link.

TomNUSDS:
@vim-usds can your re-approve so I can merge? Got sign-off from design.

vim-usds (approved):

client/src/components/AboutCard/AboutCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+                    target="_blank"
+                    rel="noreferrer">{props.actionText}</a> :
+                  <a
+                    className={'j40-aboutcard-link'}
+                    href={props.actionUrl}>{props.actionText}</a>
+                }
+              </div>
+            </Grid>
+          </Grid>
+          <Grid col={1}>&nbsp;</Grid>
```
vim-usds, line 70: I think we globally replace all `&nbsp;` with ```{` `}``` ? Could that work?

vim-usds (approved):
Looks great. Just two small things. If those are good we can merge!

client/src/pages/index.tsx
TomNUSDS: I'll start working on fixing alerts as soon as the PR is in.

client/src/components/AboutCard/AboutCard.tsx
TomNUSDS: done.

PR #549, opened 2021-08-18
Adds National Risk Index data to ETL pipeline

### LINKED ISSUE #523: As a developer, I want to ingest the FEMA National Risk Index for Natural Hazards (NRI) data set, so that it can potentially be used in the screening tool
**Description**
The FEMA National Risk Index for Natural Hazards (NRI) data needs to be available in the screening tool so that it could be visualized on the map, be included in score prototypes, be compared to our score, and/or be included for all census block groups in the csv/excel download file. 

The NRI data identifies communities most at risk to 18 natural hazards. This data is only available at the Census tract level, so for now we would apply the same indicators to all census block groups contained within a census tract.

The source of the NRI data:
](https://hazards.geoplatform.gov/portal/apps/MapSeries/index.html?appid=ddf915a24fb24dc8863eed96bc3345f8)

**Tasks**
- [ ]  Obtain NRI data files
- [ ]  Import the data 
- [ ]  Apply tract-level data to all census block groups contained in that tract 

**Definition of "Done"**
- [ ] NRI data at the census tract level is applied to census block groups so that it is available to be used by the screening tool
- [ ] NRI data is available to be displayed in the map UI
- [ ] NRI data is available to be downloaded in the CSV

widal001:
## Summary

Creates ETL class and tests for FEMA National Risk Index dataset to close #523 

## Changes Proposed

- Creates the `etl/sources/national_risk_index/` directory with the following modules
  - `etl.py` module with a `NationalRiskIndexETL` class that inherits from the base ETL class
  - `test_etl.py` module with unit tests for the `NationalRiskIndexETL` class 
  - `test_data` directory with sample data used to create fixtures for the unit tests
  - `README.md` provides some basic metadata about the data source and outlines transformations made during the ETL process
- Adds `national_risk_index` to the list of datasets in `etl/constants.py`

## Blockers/Questions

- [x] I'm still working on implementing the unit tests for this ETL class. I'll check this off when I'm done
- [x] I still need to update the `README.md` with additional details about this data source. I'll check this off when I'm done.
- [x] As a default I added all of the data from the NRI source to the csv that gets written during the `load()` step. Do we want to include all columns or just a relevant subset? If it's the latter, I can amend the `transform()` step to capture just a subset.
- [x] Right now I'm not actively pulling the data extracted from this source into the `ScoreETL` workflow, and as far as I can tell from the code, it won't get automatically picked up. Should I add this to the workflow before we merge?

## Additional Context

@NatHillardUSDS @esfoobar-usds I'd love to use one of the office hours this week to review this PR and discuss some other thoughts I have about how to streamline the process of adding new datasets to the ETL workflow.

## Instructions to Review

1. Checkout the PR locally
2. Run the unit tests: `poetry run pytest` and all tests should pass
3. Run the ETL script live for this data set: `poetry run etl -d national_risk_index`
4. Inspect the code and tests in `data_pipeline/etl/sources/national_risk_index/`


switzersc-usds:
Thanks for this PR, @widal001! I'll start reviewing this today, and yes let's discuss in office hours! To answer your questions from the checklist above:

> As a default I added all of the data from the NRI source to the csv that gets written during the load() step. Do we want to include all columns or just a relevant subset? If it's the latter, I can amend the transform() step to capture just a subset.

For the other ETLs, I believe we're including all data. I think this is fine as a pattern because it gives us easier access to anything we may need later, should we end up needing it. If we start to run into storage or performance issues then we can look at reducing the number of columns we load.

> Right now I'm not actively pulling the data extracted from this source into the ScoreETL workflow, and as far as I can tell from the code, it won't get automatically picked up. Should I add this to the workflow before we merge?

Let's keep it separate for now, so no need to add to this PR. My understanding is that we'll want to use this in the comparison tool first, which is currently still a python notebook and separate from the ScoreETL workflow anyway.

switzersc-usds:
Tests and code run locally as of current version! 🎉 I'll try again once you update based on office hours convo :)

widal001:
Changes requested during last week's Office Hours review:

- [ ] TODO: Move `test_etl.py` and `test_data/` under a `tests/` dir in `national_risk_index/` dir
- [ ] TODO: Move the `tmp_dir` into a `tests/` folder in `data/tmp`
- [ ] TODO: Promote `mock_paths` and `mock_etl` to `conftest.py` fixtures, and make it return the base class
- [ ] TODO: Change equality checking of dataframes to use `df.equals()`

widal001:
@esfoobar-usds Something I noticed with the last two commits I pushed is that `black` seems to format the code differently depending on which version of python is running. The [Data Checks test failed for 3.9](https://github.com/usds/justice40-tool/runs/3417868247?check_suite_focus=true) before I committed all of the changes to the files made by running `poetry run black data_pipeline` locally, but the same checks passed for 3.7 and 3.8. I'm wondering if that's actually the source of the quirks we were discussing via slack last week.

data/data-pipeline/data_pipeline/tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pytest.fixture
+def mock_etl(monkeypatch, mock_paths) -> None:
+    """Creates a mock version of the base ExtractTransformLoad class and resets
+    global the variables for DATA_PATH and TMP_PATH to the local mock_paths
+    """
+    data_path, tmp_path = mock_paths
+    monkeypatch.setattr(ExtractTransformLoad, "DATA_PATH", data_path)
+    monkeypatch.setattr(ExtractTransformLoad, "TMP_PATH", tmp_path)
```
switzersc-usds, line 33: Thoughts on using env variables for paths rather than monkeypatching and setting the attributes? It seems to me simpler and requiring less code. 

I saw this pattern in @NatHillardUSDS unit test PR as well and now I'm wondering if this is just python convention; in my experience on Ruby projects using env vars has always been preferable to achieve this sort of thing.

widal001: Hmm I have used this approach on a number of projects as well, and it's especially elegant with the `dynaconf` package since you can set variables either as environment variables or in a `settings.toml` 

Do you have a link to that line in Nate's PR? I'm currently having trouble envisioning how environment variables would allow us to circumvent the use of monkeypatching since the paths used in testing are different than those used when we're running the code live, which is one of the use cases that pytest suggests using monkeypatching for: https://docs.pytest.org/en/6.2.x/monkeypatch.html

Sorry if I'm missing the obvious workaround here!

widal001: Also if we were to adopt use of environment variables instead of setting class attributes, I'd probably recommend scoping that into the ETL checks issue that I'm working on currently since it would require some refactoring of all of the ETL classes that inherit from the base `ExtractTransformLoad` class

widal001: @switzersc-usds Ahhh I see your note [here](https://github.com/usds/justice40-tool/pull/509/files#r703030402) I do think this could be an elegant adjustment if we have separate environments in the `settings.toml` for `[development]` and `[testing]` 

I like the idea of setting `DATA_PATH` and `TMP_PATH` as config variables in the settings. There might be some implementation quirks involved in moving to that approach, but nothing we couldn't work through. That being said, it seems like something worth tackling in #617 or #576 to avoid scope creep for this particular PR.

switzersc-usds: Discussed this a bit with Nat and Jorge when reviewing his PR and it looks like we need to refactor our use of dynaconf settings so that we set path names there (which will function essentially as env vars). We can proceed with this method for now and the next time we add some new tests we'll do that refactor.

esfoobar-usds (approved):
🍏

Issue #535, opened 2021-08-16
Create "COMBINE-TILEFY" Github Action


switzersc-usds:
**Description**
Create one github action that combines the score into differnt zoom levels, writes the geojson files to S3, and the creates/loads the tiles. 

**Definition of Done**
- [x] Refactor `application.py` & `generate.py` to be parameterized by a `--datasource` (local is default, but if you pass in `aws` then we look for those resources externally) flag and making an external S3 call where it is using our Score.csv & GeoJSON 
- [x] Github Action (clone and use the Score Action) definition yml and call SYNC after so that we write to S3
- [x] Switch to using python click CLI instead of Poetry commands
- [x] Remove poetry commands from pyproject.toml and from README
- [ ] Making sure that the github errors and logging happen
- [ ] Write test that produces failure when the file isn't in S3

ginabeena:
Take a look at the Click library for passing optionals on the CLI

PR #548, opened 2021-08-17
Create combine-job.yml

ginabeena:
This is the github action for this work here: https://app.zenhub.com/workspaces/justice40-60993f6e05473d0010ec44e3/issues/usds/justice40-tool/535

PR #559, opened 2021-08-19
Combine Job (ZenHub Issue #535) - Do Not Merge

ginabeena:
Before this can be mereged:

- [x] Github Action Code complete & PRed
- [x] base class implemented as scaffold with cli argument passing (`-ds` & `--data-source`) 
- [ ] actually query S3 & write to S3
- [ ] add test that fails on a bad datasource being passed
- [ ] add test that fails if it cannot actually query the right files in S3
- [x] delete `/infrastructure` directory


data/data-pipeline/data_pipeline/application.py
```diff
@@ -105,10 +105,18 @@ def score_full_run():
...
 
-    score_geo()
+    if datasource is "" or " ":
+        score_geo("local")
+    if not datasource:
+        score_geo("local")
+    if datasource is "aws":
+        score_geo("aws")
+    else:
+        logger.warn('This data source is unimplemented, %s' % datasource)
```
ginabeena, line 120: this feels like the worst python code ever written 😂

ginabeena: cc: @esfoobar-usds 😂

ginabeena:
Github Action for this work: https://github.com/usds/justice40-tool/pull/548

data/data-pipeline/data_pipeline/application.py
ginabeena: https://github.com/usds/justice40-tool/pull/559/commits/7a943b749fa773bda830ae7374e6d88bb3a83e5f cleaned it up a bit cause empty string is apparently falsey like JavaScript and used `elif`

esfoobar-usds: I would use Click.Choice like so:

```
@click.option('--data-source;, type=click.Choice(['local', 'staging', 'production'], default='local')
```

Then you can pass `data_source` to the runner.

@vim-usds and I discussed having similar options for the client frontend.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -167,3 +167,9 @@ def load(self) -> None:
             self.SCORE_LOW_GEOJSON, driver="GeoJSON"
         )
         logger.info("Completed writing usa-low")
+
+
+class AWSGeoScoreETL(ExtractTransformLoad):
```
esfoobar-usds, line 172: Since the only thing that changes is the source data location, I wonder if subclassing is necessary?

esfoobar-usds (changes requested):
Happy to pair on these suggestions.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
ginabeena: *shrug* I absolutely can change this cause I have no idea what I'm doing in terms of correct python.

switzersc-usds:
@esfoobar-usds is picking up this work in #535

PR #563, opened 2021-08-20
Hotfix for Readme and ACS File name

esfoobar-usds:


data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -41,10 +41,10 @@ def __init__(self):
 
         self.STATE_MEDIAN_INCOME_FTP_URL = (
             settings.AWS_JUSTICE40_DATASOURCES_URL
-            + "/2014_to_2019_state_median_income.zip"
```
vim-usds, line 44: wondering if this should be extracted to a constant somewhere?

esfoobar-usds: For ETL purposes, those are the constants (that's why they're all CAPS) and they're set at the ETL file level because they're not used anywhere else.

switzersc-usds (approved):

Issue #567, opened 2021-08-23
Task: Update packages.json


TomNUSDS:
## Describe the task
A bunch of dependencies in the packages.json are out of date, update and test

## Acceptance Criteria

- [ ] Update
- [ ] Test

## Additional context

* Updating `@trussworks/react-uswds` to 2.x once we've replaced the footer. #564

PR #568, opened 2021-08-23
update npm packages

TomNUSDS:
Fixes: #567

Depends on #564 being integrated to make sure trusswork's footer has been removed first.


TomNUSDS:
Build failed because #564 not merged in.

vim-usds (commented):
Trying to update all libraries is not something I'd recommend. I believe when/if we have unit and e2e tests in place to ensure each update doesn't break anything (at compile and run time) then it might be ok to merge this

Issue #326, opened 2021-07-12
As a score developer, I want to split the ScoreETL class into having one ScoreETL subclass per score, so it's easier to read and update.


lucasmbrown-usds:
**Description**
We'll have multiple versions of the score running at any given time, for historical purposes as well as experimentation.

Right now `ScoreETL` calculates 5 unique scores. We should have one of these score (sub)classes per unique score. This will make it easier to update. 

As part of this refactor, we can also move some of the data cleanup logic into the ETL scripts for the data, so that's not happening redundantly as part of the score. 

**Solution**
[TBA]

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**

esfoobar-usds:
On my last conversation with @switzersc-usds, we want to not do Score generation as ETL and more like a class based modular approach.

BethMattern:
There is more work to be done now that #618 is complete. New tasks should be added/estimated/priortized.

switzersc-usds:
Moving this epic to the icebox. We're not sure whether more work here is valuable until we have an immediate need to for example show different scores on the map, host the score data in different locations, or do more complex score versioning / methodology experimentation.

PR #618, opened 2021-09-05
Initial refactor for Score ETL

switzersc-usds:
This begins the refactor for #326. So far it keeps everything in the etl_score.py file but ssplits things into functions/methods so that they will be easier to pull out into classes later. 

I tested this as I went by having a separate git-tracked folder containing the contents of `data/score/csv/full` (one of the outputs of the score etl) originally creeated from the `main` branch. Then I would move newly generated contents from this branch to that folder to replace the original, and then do `git status` to see if anything had changed. Nothing showed a having changed.

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/score/etl_score.py
+++ b/data/data-pipeline/data_pipeline/etl/score/etl_score.py
@@ -258,6 +258,7 @@ class ScoreETL(ExtractTransformLoad):
         )
 
     def _join_cbg_dfs(self, census_block_group_dfs: list) -> pd.DataFrame:
+        logger.info("Joining Census Block Group dataframes")
         census_block_group_df = functools.reduce(
             lambda left, right: pd.merge(
                 left=left, right=right, on=self.GEOID_FIELD_NAME, how="outer"
@@ -276,6 +277,7 @@ class ScoreETL(ExtractTransformLoad):
         return census_block_group_df
     
     def _join_tract_dfs(self, census_tract_dfs: list) -> pd.DataFrame:
+        logger.info("Joining Census Tract dataframes")
         census_tract_df = functools.reduce(
             lambda left, right: pd.merge(
                 left=left,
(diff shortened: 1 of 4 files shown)


switzersc-usds:
Next step for refactor: create Score class and subclasses for different scores.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -544,9 +470,129 @@ def transform(self) -> None:
...
         )
 
-        self.df["Score F (communities)"] = (
-            self.df[meets_socio_field_name] & self.df[meets_burden_field_name]
+        df["Score F (communities)"] = (
+            df[meets_socio_field_name] & df[meets_burden_field_name]
         )
+        return df
+
+    def _prepare_initial_df(self, data_sets: list) -> pd.DataFrame:
```
switzersc-usds, line 478: Ideally a lot of this happens on the ETL side instead of in here

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -544,9 +470,129 @@ def transform(self) -> None:
...
+        census_block_group_df[
+            self.GEOID_TRACT_FIELD_NAME
+        ] = census_block_group_df[self.GEOID_FIELD_NAME].str[0:11]
+
+        df = census_block_group_df.merge(
+            census_tract_df, on=self.GEOID_TRACT_FIELD_NAME
+        )
+
+        if len(census_block_group_df) > 220333:
+            raise ValueError("Too many rows in the join.")
```
switzersc-usds, line 506: If GEOID10s are read as numbers instead of strings, the initial 0 is dropped, and then we get too many CBG rows (one for 012345 and one for 12345).

switzersc-usds: ^ put as comment

switzersc-usds: are there other errors we should be raising?

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -544,9 +470,129 @@ def transform(self) -> None:
...
+        
+        # Rename columns:
+        renaming_dict = {
+            data_set.input_field: data_set.renamed_field
+            for data_set in data_sets
```
switzersc-usds, line 511: data_sets is used for 3 things:
- renaming (which we want to happen in ETL not here)
- bucketing, which we use for scores C and F (so should probably just live in the Score parent or subclasses for those)
- listing the columns we want to keep for use in score calc from the ETL'ed dataset (520 below)
- also used to set df data types to numeric (line 524) but we don't need to do that this way

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -544,9 +470,129 @@ def transform(self) -> None:
...
+        }
+
+        df.rename(
+            columns=renaming_dict,
+            inplace=True,
+            errors="raise",
+        )
+
+        columns_to_keep = [data_set.renamed_field for data_set in data_sets]
+        df = df[columns_to_keep]
```
switzersc-usds, line 521: let's keep this in the score gen code, not move to ETL

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -544,9 +470,129 @@ def transform(self) -> None:
...
+
+        # Convert all columns to numeric.
```
switzersc-usds, line 523: can do this without `data_sets`

switzersc-usds: could do this at the same time as running percentiles (eg if we move this to its own class)

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -464,62 +390,62 @@ def transform(self) -> None:
...
+            df[self.MEDIAN_INCOME_AS_PERCENT_OF_STATE_FIELD_NAME] < 0.80
+        ) & (df[self.HIGH_SCHOOL_FIELD_NAME] > 0.2)
 
-        self.df[meets_socio_field_name] = (
-            self.df[ami_and_high_school_field_name]
-            | (self.df[self.POVERTY_FIELD_NAME] > 0.40)
-            | (self.df[self.LINGUISTIC_ISOLATION_FIELD_NAME] > 0.10)
-            | (self.df[self.HIGH_SCHOOL_FIELD_NAME] > 0.4)
+        df[meets_socio_field_name] = (
+            df[ami_and_high_school_field_name]
```
switzersc-usds, line 418: let's rename `ami_and_high_school_field_name` to be explicit this is a bool field

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -464,62 +390,62 @@ def transform(self) -> None:
...
         ami_and_high_school_field_name = "Low AMI, Low HS graduation"
         meets_socio_field_name = "Meets socioeconomic criteria"
         meets_burden_field_name = "Meets burden criteria"
 
-        self.df[ami_and_high_school_field_name] = (
-            self.df[self.MEDIAN_INCOME_AS_PERCENT_OF_STATE_FIELD_NAME] < 0.80
-        ) & (self.df[self.HIGH_SCHOOL_FIELD_NAME] > 0.2)
+        df[ami_and_high_school_field_name] = (
+            df[self.MEDIAN_INCOME_AS_PERCENT_OF_STATE_FIELD_NAME] < 0.80
+        ) & (df[self.HIGH_SCHOOL_FIELD_NAME] > 0.2)
```
switzersc-usds, line 415: maybe make this clearer that this is for individuals with less than high school degre

switzersc-usds:
I will add "to dos" as comments based on our review yesterday and then convert this to an open PR instead of a draft so we can merge it in.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
esfoobar-usds: I think this should be a type check on the ETL classes and the score should assume the right inputs

data/data-pipeline/data_pipeline/etl/score/etl_score.py
esfoobar-usds: These should belong on data subclass that is used by scores C and F (Python supports multi-subclassing)

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -21,11 +21,11 @@ def __init__(self):
...
-            "C16002_001E",
-            "C16002_004E",
-            "C16002_007E",
-            "C16002_010E",
-            "C16002_013E",
+            "C16002_001E", # Estimate!!Total
+            "C16002_004E", # Estimate!!Total!!Spanish!!Limited English speaking household
+            "C16002_007E", # Estimate!!Total!!Other Indo-European languages!!Limited English speaking household
+            "C16002_010E", # Estimate!!Total!!Asian and Pacific Island languages!!Limited English speaking household
+            "C16002_013E", # Estimate!!Total!!Other languages!!Limited English speaking household
```
esfoobar-usds, line 28: Love this inline descriptors!

esfoobar-usds (approved):
Amazing work overall! 🍏

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
lucasmbrown-usds: Agreed - great stuff @switzersc-usds !

PR #624, opened 2021-09-07
Additional VSCode and Poetry tasks added

esfoobar-usds:


data/data-pipeline/.vscode/launch.json
```diff
@@ -9,63 +9,100 @@
...
       "type": "python",
       "request": "launch",
       "module": "data_pipeline.application",
-      "args": ["score-full-run"]
+      "args": [
+        "score-full-run"
+      ]
+    },
+    {
+      "name": "Score Geo",
```
switzersc-usds, line 53: what does this do exactly? wondering if there's a verb we can use that'll make this clearer

switzersc-usds (changes requested):
one small change request to use a more descriptive or longer name for "Score Geo". right now I'd have to look at the docs or code to figure out what it does, so I think the name should give more of a hint

data/data-pipeline/.vscode/launch.json
esfoobar-usds: That is what we called the "Combine" lambda function. Should we call it "Combine Score and GeoJSON"?

switzersc-usds (approved):
🎉 thanks!

PR #629, opened 2021-09-08
Add Score D to USA Low

esfoobar-usds:


vim-usds (approved):
Will this change add it both low and high tiles?

esfoobar-usds:
No, this will add two properties inside the tiles called `"D_SCORE . You should update the frontend to read that property on the lower zoom level.

vim-usds:
@esfoobar-usds 
I might be missing something decoding the lower tiles from the [map on the main branch](https://d2zjid6n5ja2pt.cloudfront.net/justice40-tool/main/en/cejst#4.04/32.57/-103.14) results in the lower tiles still showing the E_SCORE?

```json
{
          "type": "Feature",
          "properties": {
            "E_SCORE": 0.9339749925937375
          },
          "geometry": {
            "type": "Polygon",
            "coordinates": [ ... ]
           }
  }
  ```
  
   I wonder if we need to invalidate the cache?

esfoobar-usds:
No, I haven't uploaded them yet. I'm stuck in a rounding issue. I'll let you know when this is up.

vim-usds:
> No, I haven't uploaded them yet. I'm stuck in a rounding issue. I'll let you know when this is up.

Ok cool! Thanks!

esfoobar-usds:
Good thing with this PR: size of tiles should be extremely smaller! :)

PR #649, opened 2021-09-09
Removes content methodology page

### LINKED ISSUE #630: Task: Remove content from data and methodology page
## Describe the task
The current content is not applicable and may confuse stakeholders that we're showing this to. Let's remove all content for now and just add the text "In progress" to this page. 

Note that because of version control we will be able to easily reinstate the style, content, etc of this page when we decide to bring it back. 

Note that we are leaving the page and its link in the nav bar because we want people to know we intend to put text here. 

## Acceptance Criteria

- [x] data and methodology page just says "In progress"

## Additional context

* Add any other context about the task here.
* Or here

vim-usds:
closes #630


TomNUSDS (approved):
LGTM. 

Thoughts:
It might be better to use a feature flag if there's a chance we'll put it back (or maybe just remove code instead of just commenting it out)?

Not sure if this also should have an updated testing snapshot and/or localized string thingy?

Issue #620, opened 2021-09-07
Task: Update download file name to not have spaces


switzersc-usds:
## Describe the task
It's best practice not to use spaces in file names because some operating systems have trouble with this. 

## Acceptance Criteria

- Update backend code to use file names with no spaces

## Additional context

* This is followup work from #578

Issue #596, opened 2021-09-02
Download file does not include cumulative index score


BethMattern:
**Describe the bug**
The .xls and .csv files in the download packet do not include the cumulative index score or any indication as to whether or not the census block group is prioritized. Both files should include Score D and any new score we add over the next 2 weeks. 

For now, just add score D. For each new score, we will make a requirement for the score issue to add that score to the file at the time of that score's development.

**To Reproduce**
1. Click on the download link
2. Open zip file
3. Open .xls 
4. See that there is no score value or prioritized yes/no indicator

**Expected behavior**
The score value, score percentile and top 25th percentile True/False data should be included as columns in the file.

switzersc-usds:
Putting in icebox for now because we don't know which score we'll be using for this and what that type of score is (in/out or percentile), and we don't expect to know by the end of the next sprint.

PR #659, opened 2021-09-10
Update downloadable zip file

switzersc-usds:
* Don't use spaces in the name, as per #620
* Add the score D columns, as per #596


esfoobar-usds (approved):
🍏

esfoobar-usds (approved):
🍏

PR #662, opened 2021-09-10
Run all Census, ETL, Score, Combine and Tilefy in one command

### LINKED ISSUE #595: Run all backend ETL scripts in one command
Right now it takes 4 commands to generate all the tiles to be run locally. We should make this just one command. We should also make sure this command gets run on docker-compose up IF the tiles do not already exist.

esfoobar-usds:
Fixes #595

This PR [introduces ](https://github.com/usds/justice40-tool/blob/esfoobar-usds/595-all-etl-in-one-script/data/data-pipeline/data_pipeline/application.py#L145)a new CLI command called `data_full_run` which can be run to do all the end-to-end process to generate tiles (Census download, run all ETL processes, Score generation, Combine Score with Census GeoJSON and generate tiles).

The CLI function has an optional flag called `--check` or `-c` that makes the whole process run ONLY if a local empty file called `first_run.txt` has NOT been written on the local `data` directory.

The idea is that the command [is called](https://github.com/usds/justice40-tool/blob/esfoobar-usds/595-all-etl-in-one-script/data/data-pipeline/Dockerfile#L36) with the `-c` flag only the first time the container is built from `docker-compose up`, so that the user can easily check out the code from Github and run `docker-compose up`. On subsequent `docker-compose up`s, the script won't run.

This leaves the option that the whole end-to-end process can be called manually or from CI/CD pipelines without checking the file by not passing the `-c` flag.

Since the file is on the `/data_pipeline/data` folder which is Git ignored, the file never gets added to the repo -- i.e., it will only be locally set on the user's directory as a semaphore.


data/data-pipeline/data_pipeline/utils.py
```diff
@@ -198,6 +199,21 @@ def temp_folder_cleanup() -> None:
     remove_all_from_dir(data_path / "tmp")
 
 
+def check_first_run() -> bool:
+    """Checks if a local semaphore has been set to only run once the
+    calling function"""
```
switzersc-usds, line 204: This statement is confusing to me; I think it's missing a phrase or verb? Specifically `set to only run once the calling function`: once the calling function has done what? or is it once _per_ calling function? or once the function has been called?

data/data-pipeline/data_pipeline/utils.py
```diff
@@ -198,6 +199,21 @@ def temp_folder_cleanup() -> None:
...
+
+    data_path = settings.APP_ROOT / "data"
+    file = "first_run.txt"
+
+    if not os.path.isfile(data_path / file):
+        cmd = f"touch {data_path}/{file}"
+        call(cmd, shell=True)
```
switzersc-usds, line 211: does this file ever get deleted?

data/data-pipeline/data_pipeline/utils.py
```diff
@@ -198,6 +199,21 @@ def temp_folder_cleanup() -> None:
...
 
 
+def check_first_run() -> bool:
+    """Checks if a local semaphore has been set to only run once the
+    calling function"""
+
+    data_path = settings.APP_ROOT / "data"
+    file = "first_run.txt"
+
+    if not os.path.isfile(data_path / file):
```
switzersc-usds, line 209: the multiple `not`s in the conditional logic (starting with `if check and not check_first_run()` makes this hard to follow. thoughts on refactoring to not use the `not`s? 

maybe renaming the method would also help, eg `def first_run_started() -> bool` (my Ruby brain is looking for question-type methods for booleans). this would allow you to change the original conditional phrase  to be `if check and first_run_started()`, which IMO is much clearer.

data/data-pipeline/data_pipeline/application.py
```diff
@@ -121,5 +122,55 @@ def generate_map_tiles():
...
+    census_reset(data_path)
+    data_folder_cleanup()
+    score_folder_cleanup()
+    temp_folder_cleanup()
+
+    logger.info("*** Downloading census data")
+    etl_runner("census")
+
+    logger.info("*** Running all ETLs")
+    etl_runner()
```
switzersc-usds, line 161: does this re-run census since it is supposed to run all ETLs, or does it skip census?

switzersc-usds (commented):
Some questions and suggestions to refactor for clarity.

data/data-pipeline/data_pipeline/utils.py
esfoobar-usds: Per the description I just added to the PR above, no, it has to be manually deleted by the user, but it's not checked in to the repo since it's on a git ignored directory and only influences the application if the command is called with the `-c` flag.

data/data-pipeline/data_pipeline/utils.py
esfoobar-usds: The file signals that a full run was completed, not that it has started. Maybe let's connect IRL to talk more about the logic?

data/data-pipeline/data_pipeline/application.py
esfoobar-usds: It executes the Census ETL, yes. The Census ETL does have some caching logic to not re-download files that already exist in the file system but it does go through the SHP files extraction regardless.

switzersc-usds (approved):
Wahoo!

PR #665, opened 2021-09-10
adds initial prefix

vim-usds:


.github/workflows/deploy_main.yml
```diff
@@ -79,11 +83,13 @@ jobs:
...
         env:
-          DISTRIBUTION: "ED03LPVC4OXSW" # TODO: update to new CDN after access is provided
+          DISTRIBUTION: ${{env.WEB_CDN_ID}}
           PATHS: "/*"
           AWS_REGION: "us-east-1"
           AWS_ACCESS_KEY_ID: ${{ secrets.CLIENT_DEV_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.CLIENT_DEV_AWS_SECRET_ACCESS_KEY }}
+      - name: Remove staging artifacts
+        run: aws s3 rm s3://usds-geoplatform-justice40-website/justice40-tool/ --exclude "*" --include "$(echo $PR_NUMBER)-*" --recursive
       - name: Update PR with deployed URL
```
switzersc-usds, line 92: Do we want this in the `deployed_main` action? I think it should only be in the staging one since it's updating an open PR (and for deploying main there would be no open PR).

switzersc-usds: (I realize that this isn't a change in this PR though, so we can address separately if it's too much scope creep)

.github/workflows/deploy_main.yml
```diff
@@ -8,6 +8,10 @@ on:
     branches: [main]
     paths:
       - "client/**/*"
+env:
```
switzersc-usds, line 11: Up on line 7, it looks like this runs when a PR is closed (` types: [closed]`). Should this be `merged` instead or additionally (if that's an option)? I don't see any code in this action that suggests that it only runs on merge (as opposed to a PR being closed without merging), so I want to confirm that this is only being run on merge... 

Now that I think about it though, while the deploy stuff should only happen on merge of the PR, the stale artifact cleanup should happen on any type of close, regardless of merge.

.github/workflows/deploy_main.yml
```diff
@@ -79,11 +83,13 @@ jobs:
...
         uses: chetan/invalidate-cloudfront-action@master
         env:
-          DISTRIBUTION: "ED03LPVC4OXSW" # TODO: update to new CDN after access is provided
+          DISTRIBUTION: ${{env.WEB_CDN_ID}}
           PATHS: "/*"
           AWS_REGION: "us-east-1"
           AWS_ACCESS_KEY_ID: ${{ secrets.CLIENT_DEV_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.CLIENT_DEV_AWS_SECRET_ACCESS_KEY }}
+      - name: Remove staging artifacts
+        run: aws s3 rm s3://usds-geoplatform-justice40-website/justice40-tool/ --exclude "*" --include "$(echo $PR_NUMBER)-*" --recursive
```
switzersc-usds, line 91: So `--include "$(echo $PR_NUMBER)-*" --recursive` means that the only objects that will be removed are only any files and directories that have the `$PR_NUMBER` in the path name? What exactly does `--exclude "*"` do?

.github/workflows/deploy_staging.yml
```diff
@@ -66,7 +68,7 @@ jobs:
           path: ./public
       - name: Set DESTINATION_FOLDER for branch
         run: |
-          echo "DESTINATION_FOLDER=$(echo $GITHUB_SHA | cut -c 1-6)" >> $GITHUB_ENV
+          echo "DESTINATION_FOLDER=$(echo $PR_NUMBER-$GITHUB_SHA | cut -c 1-11)" >> $GITHUB_ENV
```
switzersc-usds, line 71: I'm curious about the change to use PR number as well as the Git SHA. Were we running into issues with just the SHA?

switzersc-usds (commented):
Lots of questions! Let's discuss before approving :)

.github/workflows/deploy_staging.yml
vim-usds: Since a given PR could have multiple builds (and therefore multiple SHAs and therefore multiple artifacts), this would allow us to group all the artifacts by PR and remove them all in a single command. :)

switzersc-usds: oh nice! cool

.github/workflows/deploy_main.yml
vim-usds: If we were to place it in the staging and if the PR was still open, the previous staging links would not load. Sometimes it's helpful to compare staging links on a given PR

.github/workflows/deploy_main.yml
vim-usds: While this makes sense to me, checking the PR types, I don't think 'merged' is an option?

https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request

.github/workflows/deploy_main.yml
vim-usds: I'm placing the exclude first meaning, first exclude everything. Then include only the artifacts with the PR prefix. It's an added safety check to ensure that only the folder with the PR prefix get's removed. A --dryrun flag was ran and it seemed to work as expected.

switzersc-usds: It looks like maybe this would delete folders for pull requests that end with a given number, not match a given number. For example if a pull request number is 122, this code would delete a folder with 122- AND 1122- in the name, right? Which I don't think we'd want, even though it seems like an unlikely edge case.

vim-usds: @switzersc-usds : I don't think it would delete the folder with 1122, to be sure, ran a quick test:

Created both folder prefixes, 3 folders (1 sub-folder) with a 111 prefix and one with 2111 prefix:
Screen Shot 2021-10-05 at 12 23 02 PM

Then ran the delete command twice:

[Run 1 ](https://github.com/usds/justice40-tool/runs/3806950036?check_suite_focus=true): PR set to 2111 - Expecting only 1 folder to be removed:

Screen Shot 2021-10-05 at 12 27 09 PM

Meets expectation to delete one and only 1 folder

[Run 2](https://github.com/usds/justice40-tool/runs/3807086047?check_suite_focus=true): PR set to 111 - Expecting 3 folders (and 1 sub-folder) to be removed:
Screen Shot 2021-10-05 at 12 29 56 PM

Meets expectation to delete 3 and only 3 folders (and 1 sub-folder).

switzersc-usds: Oh great, thanks!

.github/workflows/test-remove-artifacts.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
```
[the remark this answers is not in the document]
vim-usds: Nope! Removing

.github/workflows/deploy_main.yml
esfoobar-usds: Yeah, @switzersc-usds, `closed` is the right action, since it's meant to be kicked off when the review is finished and the PR is closed, which signals the final merge.

esfoobar-usds (approved):
🍏  awesome work!

PR #670, opened 2021-09-13
Update to score G from score D

vim-usds:
In order for this to work, the PBF files for both the high and low zoom tiles will require the Score G key. This can be validated by running the [following](https://github.com/usds/justice40-tool/blob/main/client/VIEW_MAP_DATA.md#decoding-a-pbf-file) on the high and low zoom tiles.

TL;DR - an example of a low zoom tile (the z value in the url is 3):

```
curl https://d3jqyw10j8e7p9.cloudfront.net/data-pipeline/data/score/tiles/low/3/1/3.pbf -o d3_313.pbf

tippecanoe-decode d3_313.pbf 3 1 3 | jq
```

The output of tippecanoe-decode should have Score G key in the payload.

The same should happen with a high zoom tile.

### DIFF
--- a/client/src/data/constants.tsx
+++ b/client/src/data/constants.tsx
@@ -26,8 +26,8 @@ export const FEATURE_TILE_LOW_ZOOM_URL = featureURLForTilesetName('low');
 export const PERFORMANCE_MARKER_MAP_IDLE = 'MAP_IDLE';
 
 // Properties
-export const SCORE_PROPERTY_HIGH = 'Score D (percentile)';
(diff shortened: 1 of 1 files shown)


esfoobar-usds:
@lucasmbrown-usds the Score G has no percentile, i.e it's just a boolean.:

```
| Score G (communities) | Score G |
| True                  | 1       |  
```

@vim-usds that means the code in this PR won't work. Let's connect when you're up.

esfoobar-usds (approved):
🍏

PR #671, opened 2021-09-13
hotfix wrong score tile csv path

esfoobar-usds:


vim-usds (approved):
@esfoobar-usds - approving however not sure of a way to actually validate functionality. Maybe sometime In the future I can get set up to actually do a real PR for backend stuff where I'm pulling the branch and validating.  🤯   Thx!

PR #672, opened 2021-09-14
Prototype G

lucasmbrown-usds:
@esfoobar-usds or others: to run this, make sure you rerun at least the following ETLs: 

* `poetry run etl --dataset census_acs_median_income` 
* `poetry run etl --dataset census_acs`

The output we desire is Score G (will be all 1s and 0s, with ~25% of the communities as 1s).


data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -564,13 +679,15 @@ def _prepare_initial_df(self, data_sets: list) -> pd.DataFrame:
 
         # Calculate median income variables.
         # First, calculate the income of the block group as a fraction of the state income.
-        # TODO: handle null values for CBG median income, which are `-666666666`.
```
lucasmbrown-usds, line 567: Moved this to the ETL.

esfoobar-usds (approved):
🍏

esfoobar-usds:
Merging these changes since there are some map generation fixes as well.

@vim-usds remember to create a new PR so that the client frontend uses "Score G (percentile)" instead of "Score G" before we do a new map tile s3 sync.

PR #673, opened 2021-09-14
Gherkin cypress spike

### LINKED ISSUE #390: [Spike] Set up Cucumber/Gherkin to be used in the front end test suite
**Description**
We should adopt Cucumber/Gherkin for describing test cases, to make it easier to write and maintain them and to create a common vocabulary around testing and test scenarios. This will make QA a lot easier and faster and make it easier to maintain the codebase.

**Solution**
We should setup [cypress-gherkin-preprocessor](https://github.com/TheBrainFamily/cypress-cucumber-preprocessor) to bridge cypress and gherkin for describing test cases. 

- Using cypress out of the box helps engineers, but is less visible to non-engineers / makes it harder to write and maintain test suites. 

- https://github.com/TheBrainFamily/cypress-cucumber-preprocessor
- https://cucumber.io/
- https://cucumber.io/docs/gherkin/

**Tasks**
- [x] Adopt gherkin-cypress-cucumber
- [x] Setup a test or two to show how it would work 

**Definition of "Done"**
- [x] An initial pattern for creating more gherkin tests
- [x] More fleshed out issues to write remaining gherkin tests

vim-usds:
Closes #390 

Adds initial testing of navigation from all pages to the about page.


client/src/pages/cejst.tsx
```diff
@@ -1,5 +1,8 @@
 import React from 'react';
+import {defineMessages} from 'react-intl';
 import {Link} from 'gatsby-plugin-intl';
+import {useIntl} from 'gatsby-plugin-intl';
+import {Grid} from '@trussworks/react-uswds';
```
switzersc-usds, line 5: Just curious - how come these are necessary to import now? What about writing these gherkin tests inspired the code changes here?

switzersc-usds (approved):
Looks good to me! Just had one question to understand the explore the tool plan refactor more. Stoked to get these tests started!!

client/src/pages/cejst.tsx
vim-usds: Great question!

The basic Gherkin test currently tests that the title of each page is set correctly. To add the title require refactoring a bit. 
1. moving away from custom layouts to one that uses the Grid system (trussworks)
2. internationalizing the content as this was something that needs to be done and since I was here touching the code decided to wrap it into this branch

PR #681, opened 2021-09-14
dependabot bump pillow

esfoobar-usds:


switzersc-usds (approved):

esfoobar-usds:
@switzersc-usds @lucasmbrown-usds heads up, testing for this PR is failing because we need to add encoding to any `open` command, per [PEP-0597](https://www.python.org/dev/peps/pep-0597/). I will add to the current codebase but take into consideration for future open commans.

PR #686, opened 2021-09-15
(Not urgent) Minor updates to scoring comparison tool

lucasmbrown-usds:
Adds a few formatting changes and column changes to scoring comparison outputs.


esfoobar-usds (approved):
🍏

PR #695, opened 2021-09-16
Add action for ping test and alert to Slack

switzersc-usds:
Based on @sverchdotgov 's https://github.com/sverchdotgov/github-actions-experiments/blob/master/.github/workflows/scheduled-email.yml

This is set to run every hour. We can modify as needed. It is also using our cloudfront URL and we'll need to change this to our new URL once we have it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,33 @@
+# .github/workflows/ping-test.yml
+name: Trigger Run of site ping test
+on:
+  # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+  schedule:
+    # Run every hour
+    - cron: '0 * * * *'
+  # https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'     
+        required: true
+        default: 'warning'
+      tags:
+        description: 'Ping test'  
+jobs:
+  sitePingCheck:
+    name: Slack Notification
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,33 @@
...
+    - uses: Leocardoso94/is-my-site-up@v1.2
+      with:
+        site: "https://d2zjid6n5ja2pt.cloudfront.net/justice40-tool/main/en/"
+    # https://www.ravsam.in/blog/send-slack-notification-when-github-actions-fails/
+    - uses: ravsamhq/notify-slack-action@v1
+      if: always()
+      with:
+        status: ${{ job.status }}
+      env:
+        SLACK_WEBHOOK_URL: ${{ secrets.ACTION_MONITORING_SLACK }}
```
esfoobar-usds, line 33: Does the secret define which Slack channel to notify?

esfoobar-usds (approved):
🍏 with a minor question

.github/workflows/main.yml
sverchdotgov: Yup, I think when you create the webhook, it's a webhook for a specific channel.

I copy-pasted this, so maybe the secret should be called something like "JUSTICE40_TOOL_SLACK_ALERTS" so people know what the secret is for.

This section may have bugs since I wasn't able to test with a real secret, but once you merge and set the secret hopefully it'll be obvious what's wrong

switzersc-usds:
Updated to run every 5 minutes and use a more descriptive secret. Added the secret and going to merge to test out!

PR #704, opened 2021-09-17
Update ping test workflow

switzersc-usds:
Update name of workflow and only send slack notification on failure


esfoobar-usds (approved):
🍏

PR #706, opened 2021-09-17
Add env variable to set path prefix

vim-usds:
- When routing traffic to our new URL, the path prefix needs to be blank
- adding path prefix to be set by env var to yml files


esfoobar-usds (approved):
🍏

PR #707, opened 2021-09-17
Add analytics script

vim-usds:
- also add the geoplatform official site to yml file


TomNUSDS (approved):
LGTM. Usually these things require some kind of key that's associated with a "host" domain to use. DAP doesn't?

vim-usds:
Thank you! Gonna wait until Monday to merge.. Friday at 5pm is just asking for trouble! Plus I got weekend plans! 🤣 

> LGTM. Usually these things require some kind of key that's associated with a "host" domain to use. DAP doesn't?

Yeah I think so. The URL in the script tag alludes to some UID: 

`GWT_UAID:["UA-33523145-1"]`

Have to look into how to use that or if it's even needed. Looks strikingly similar to Google Analytics? 🤔

TomNUSDS:
> Have to look into how to use that or if it's even needed. Looks strikingly similar to Google Analytics? 🤔

I think that DAP is built on top of Google Analytics!

Issue #628, opened 2021-09-08
Add timestamp to the output files from the score generation


switzersc-usds:
This will make it a lot easier to know what version of the package you've downloaded, as we're sending around to people at different times. 

Let's add the timestamp of generation in the file to the output of "downloadable". We just want the timestamps on the csv and xlsx files, not the zip file name (because that would require more updates on the front end).

We will add the version (v0.1) to the column ("Community of Focus (v0.1)"). See other issue for that.

switzersc-usds:
While we're doing this, let's also change the two spreadsheets to use the word "communities" instead of the word "usa" -- so communities-{timestamp}.csv and communities-{timestamp}.xlsx

PR #708, opened 2021-09-17
Change downloadable file names

switzersc-usds:
Address #628 

This PR:
* Remove constants because we're dynamically creating these
* Update to "communities" for the descriptor word based on team convo
* Add timestamp in 2020-09-20-0930 format because I personally think this is the best ^.^ We can change it though
* Adds CLI for score post so we can easily just run that from the terminal

Still to do:
- [x] Get tests passing
- [x] Get command line working
- [x] Confirm this works

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
+++ b/data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
@@ -1,6 +1,7 @@
 import json
 import zipfile
 from pathlib import Path
+from dynaconf.loaders.py_loader import try_to_load_from_py_module_name
 
 import pandas as pd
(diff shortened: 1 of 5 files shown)


data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -1,6 +1,7 @@
 import json
 import zipfile
 from pathlib import Path
+from dynaconf.loaders.py_loader import try_to_load_from_py_module_name
```
switzersc-usds, line 4: oh actually i dont know if this needs to be here / why i added it. maybe its leftover from a merge conflict.

esfoobar-usds (approved):
Amazing! 🍏

Issue #513, opened 2021-08-13
[SPIKE] Investigate why styles imported from scss modules are undefined / removed in snapshots


vim-usds:
Investigate why in some cases, classNames are removed or marked as undefined in snapshots when styles are imported from scss modules.

PR #744, opened 2021-09-23
Update some additional dataset links

vim-usds:
- update snapshots

Note - aware of undefined in snapshots, ticket here
https://github.com/usds/justice40-tool/issues/513


TomNUSDS (approved):
LGTM. 

Something to think about for the future: could these urls need localization at some point? (e.g. `https://www.epa.gov/ejscreen/technical-documentation-ejscreen` -> `https://www.epa.gov/ejscreen/es/technical-documentation-ejscreen` or something)

vim-usds:
> LGTM.
> 
> Something to think about for the future: could these urls need localization at some point? (e.g. `https://www.epa.gov/ejscreen/technical-documentation-ejscreen` -> `https://www.epa.gov/ejscreen/es/technical-documentation-ejscreen` or something)

Good point. I imagine that the translation service would look for a `es` version of the link (in this case EPA). We are actually working on that now

PR #758, opened 2021-09-28
Attempt to merge Lighthouse Fix branch from Nat

vim-usds:
The a11y cypress tests pass for all pages except for the about page. This page requires an h1 tag.


vim-usds:
For both mobile and desktop, all pages on the above staging URL has an A11y lighthouse score of 100% ! 

- About = 100% ✅ 
- Explore = 100% ✅ 
- Methodology = 100% ✅ 
- Contact = 100% ✅ 

The Axe library is more stringent it seems than Lighthouse.

TomNUSDS (approved):
Massive change. But it LGTM.

vim-usds:
closes #420

PR #759, opened 2021-09-29
Removes GA Gatsby plugin

vim-usds:
- places original GA tag back into head


TomNUSDS (approved):
LGTM. I guess there's not snapshot updates required since it's going into the <head>?

vim-usds:
> LGTM. I guess there's not snapshot updates required since it's going into the ?

Yeah I think we have to add a snapshot test on the layout file. There's a ticket to makes sure all files and components have tests.

PR #766, opened 2021-10-01
Update site with latest download file date + size

switzersc-usds:
I re-generated the download files today as part of #628 so needed to update the website to match.


client/src/pages/__snapshots__/methodology.test.tsx.snap
```diff
@@ -313,12 +313,12 @@ exports[`rendering of the DatasetContainer checks if various text fields are vis
               <div>
                 <div>
                   <div>
-                    Draft communities list v0.1 (143MB)
```
vim-usds, line 316: Love that the snapshots were updated!

vim-usds (approved):
LGTM!

Issue #389, opened 2021-07-20
As a developer I want to chop up my backend application code in such a way that they are all Lambda functions on AWS.


ginabeena:
![image.png](https://images.zenhubusercontent.com/60c7a57fe54d169cb4b228e2/b94cbd60-c2c3-473d-8f5a-4fc7ba61c9ed)

Using this diagram as a model, chop up our infrastructure and rework it so that every section gets it's own function definition inside of the `/functions/` directory inside of `/infrastructure` as well as pushing static content that never changes to our S3 bucket so that the front-end may query it. 

- [x] add infrastructure diagram and explanations of how the infra works into the `/infrastructure` directory
- [x] push TigerJSON for states to S3 and delete infrastructure code that does that
- [x] run script for geopandas combining the state TigerJSON to create single national geojson file, document that, and push that to S3 bucket (whether or not that's a lambda is lower priority) - yes, script for now. this is static data so doesn't need to change.
- [ ] create **SCORE lambda** which combines datasets to generate a CSV for score information and writes that to S3
- [ ] create **COMBINE lambda** which consumes a tippecanoe event of high & low zoom granularity for national GeoJSON and outputs high and low granularity zoom MbTiles compressed files 
- [ ] create **TILE-FY Production Data lambda** which writes our MbTiles data into our Production S3 bucket
- [ ] ~create **TILE-FY Stage Data lambda** _configuration_ so that Github Action can push that data locally or in STG to STG S3~ Not doing STG for now

ginabeena:
Should the pushing of the MbTiles to S3 be a Github Action only? 🤔

switzersc-usds:
Summary of infra sync convo 8/16:
- Moving to use Github actions for these three functions instead of Lambdas, so that infra is all done in one place instead of in a combo of GA and Lambda
- All of our "deployment" is just reading/writing to S3, so it's not really deployment in a more traditional sense and therefore isn't so weird for GA

PR #790, opened 2021-10-07
Update diagrams in architecture docs

### LINKED ISSUE #684: Update architecture diagram in docs
Current diagram is out of date. We should update this so we can onboard folks more easily and have more accurate documentation!

switzersc-usds:
* Updates overall diagram to represent current arch and process
(including vision for data selection)
* Updates geo data pipleline arch diagram and removes geoplatform
version since we only have one version of this for the foreseeable
future and we're using geoplatform infradstructure
* New data pipeline arch diagram uses diagram from #389 as a base

Closes #684


esfoobar-usds (approved):
This looks generally good to me, although I have a question about this line:

image

I can't remember the frontend accessing geojson files directly -- the metadata the client renders is stored on the `pbf`'s themselves. Am I missing something?

PR #794, opened 2021-10-08
Update es.json to have same fields as en.json

katherinedm-usds:
Update es.json to have same fields as en.json, translated a couple strings to Spanish


vim-usds:
> @vim-usds I don't see Spanish content on this link: http://usds-geoplatform-justice40-website.s3-website-us-east-1.amazonaws.com/justice40-tool/794-afb9606/es, maybe I'm doing something wrong?

I think this is right. What I believe @katherinedm-usds did was copy over the English json file and overwrote the Spanish json file as a starting point for the translation ( which I agree with). She changed a few lines in the overwritten Spanish json file with some Spanish. I couldn't find the lines, but if we can identify the Spanish lines she changed, we can see if that Spanish content shows up on the PR'd URL.

katherinedm-usds:
Awesome! I'll manually change the rest of the JSON values from object to string
image

katherinedm-usds:
Just had an awesome awesome sync with @vim-usds on how to make sure whatever version of es.json the translation team is working in (I have not kicked them off yet) has the latest combine beta message and a couple other small copy tweaks, and that after kicking them off we have a good reliable way to point the translation team to just the strings that need additional translation. I'll likely delete the branch I created after Vim merges what we just talked about:
 - Vim will make the en.json updates he's working on and then make a corresponding es.json that has extra ".en" keys with English string values to make it easier to keep track of which string is which and easier for the translation team, " "about.page.sub.header.1.text.1.en": "On January...", these will be ignored when being parsed, and then the regular keys such as "about.page.sub.header.1.text.1": "El 27 de enero de 2021..." will show the Spanish text. 
 - I'll pass this version of the es.json to the translation team
 - If during translation or after additional changes to the English copy are made, Vim can run a script that corresponds the English copy change to the Spanish key and es.json line number and to start I'll just manually keep track of what new strings will need additional translation

PR #798, opened 2021-10-14
Migrates global footer styles to USDS

### LINKED ISSUE #640: Make colors consistent and use public-sans font
Leftover from #571

vim-usds:
- enables colors in settings file
- updates snapshots

closes #640 

This ticket is really an epic as it touches all files and components. Now that colors, fonts, spacing, sizing are able to be used from USWDS, the approach is, that each time a component is touched, it will be migrated to use USWDS styles

### DIFF
--- a/client/src/styles/global.scss
+++ b/client/src/styles/global.scss
@@ -35,9 +35,8 @@ There are 3 things that should be included in this file:
 
 /***************** GLOBAL STYLES **************************************************************/
 
-$primary-color: #112f4e;
-$j40-blue-background-color: #EFF6FB;
-$sidePanelBorderColor: #f2f2f2;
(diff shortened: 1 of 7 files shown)


client/src/styles/global.scss
```diff
@@ -122,7 +124,7 @@ components include:
 
   span {
     // make sure the open close chevron is colored correctly
-    color: $primary-color;
+    color: $primary-color; // Todo color: remove this global
```
TomNUSDS, line 127: I wonder if this can be removed now? I think it was from a time when the top nav was inverted (dark themed).

vim-usds:
> Yay. Thanks for cleaning this up! A lot of it was a mess that I left behind.

Absolutely! It's still not super clear on how to style.. global vs modules! I'm leaning toward global, might have some repeated styles that could be abstracted

PR #815, opened 2021-10-26
Big ole score refactor

### LINKED ISSUE #326: As a score developer, I want to split the ScoreETL class into having one ScoreETL subclass per score, so it's easier to read and update.

switzersc-usds:
Closes #326 

This is a big refactor that pulls out most scores into their own classes that are child classes of a simple Score base class, and that get aggregated by a new ScoreCalculator. It also removes the `data_sets` list.

There are probably lots of improvements and further refactors that can be made, so let's try to figure out what should be in scope for this current PR during review.


data/data-pipeline/data_pipeline/score/score_d.py
```diff
@@ -0,0 +1,29 @@
+from data_pipeline.score.score import *
+
+
+class ScoreD(Score):
+    def add_columns(self) -> pd.DataFrame:
+        logger.info("Adding Scores D and E")
```
switzersc-usds, line 6: Ik ept scores d and e together because they use the same fields and mostly the same logic, so it didn't seem worth splitting out and moving shared logic to the parent class (which would decrease code legibility).

data/data-pipeline/data_pipeline/score/score_a.py
```diff
@@ -0,0 +1,13 @@
+from data_pipeline.score.score import *
```
switzersc-usds, line 1: This is what's causing the lint to fail. I will update all these to name the classes/modules we need.

data/data-pipeline/data_pipeline/score/score.py
```diff
@@ -0,0 +1,14 @@
+import pandas as pd
+import data_pipeline.score.field_names as FN
+
+from data_pipeline.utils import get_module_logger
+
+logger = get_module_logger(__name__)
```
esfoobar-usds, line 6: We don't need the logger here if it's not used within the file.

esfoobar-usds (changes requested):
Great work overall, just a couple of changes.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -1,91 +1,19 @@
...
-from pathlib import Path
 import pandas as pd
 
 from data_pipeline.etl.base import ExtractTransformLoad
+from data_pipeline.score.score_calculator import ScoreCalculator
+import data_pipeline.score.field_names as FN
+
 from data_pipeline.utils import get_module_logger
-from data_pipeline.etl.score.score_calculator import ScoreCalculator
+from . import constants
```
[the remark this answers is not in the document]
switzersc-usds: @esfoobar-usds this is what we're doing in `etl_score_post.py`. It's quite a lot of constants; I'm concerned that if we listed out all of them it kind of defeats the purpose of having a constants file (because it puts the constant names in multiple places that then have to be maintained, and then it takes up lots of redundant space). 

I could refactor this and the identical line in `etl_score_post.py` to ue the pattern I use elsewhere for the field names set of constants (`import data_pipeline.score.field_names as FN`) so that this line becomes `import data_pipeline.etl.constants as CONSTANTS`. Thoughts?

data/data-pipeline/data_pipeline/score/score.py
switzersc-usds: It's used in all of the child classes. I could redefine it in every child class, but that seems less DRY and also the more we move out of this parent class to every child class, the less it feels like we're utilizing the inheritance structure (which, admittedly, maybe we don't need).

data/data-pipeline/data_pipeline/score/score.py
```diff
@@ -0,0 +1,14 @@
+import pandas as pd
```
switzersc-usds, line 1: I don't know that this is a helpful parent class; furthermore, linting is not happy with me importing libraries here that I use only in child classes.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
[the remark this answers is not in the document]
esfoobar-usds: I see -- even though we don't use `score_post` on Docker -- I would need to theckout your branch and run a Docker instance from it to see if get any issues.

The key line is this, we install the `data_pipeline` as a Poetry package, and thus the relative imports don't work -- at least that's what Nat and I saw at the beginning.

Did you mean `import data_pipeline.etl.score.constants as CONSTANTS`?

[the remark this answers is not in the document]
switzersc-usds: > Did you mean import data_pipeline.etl.score.constants as CONSTANTS? 

Oh, yes

Okay! Want to run this on Docker and see what happens? I can also do this tomorrow.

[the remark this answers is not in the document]
lucasmbrown-usds: Any reason we wouldn't be fine with just `import data_pipeline.etl.score.constants`, and then when we need to reference them, using `constants.FIELD_NAME`?

data/data-pipeline/data_pipeline/score/score.py
lucasmbrown-usds: Hmm, you do use it here though, in the typings? It's still not happy with that?

data/data-pipeline/data_pipeline/score/score_a.py
```diff
@@ -0,0 +1,13 @@
+from data_pipeline.score.score import Score, FN, pd, logger
```
lucasmbrown-usds, line 1: It seems a little odd to me we're not just directly importing pandas and the logger here – is there a reason we prefer to inherit them?

data/data-pipeline/data_pipeline/score/score.py
```diff
@@ -0,0 +1,14 @@
...
+import data_pipeline.score.field_names as FN
```
lucasmbrown-usds, line 2: nit: i generally prefer being verbose and clear over concise with room for misinterpretation, and `FN` is throwing me off since `fn` is often used to mean `function` (that's what I thought this meant at first). 

Any reason we can't just leave it at `import data_pipeline.score.field_names` and reference, e.g., `field_names. POVERTY_PERCENTILE_FIELD`?

lucasmbrown-usds: Also maybe we just import this directly in the child classes, and not here in the parent?

data/data-pipeline/data_pipeline/score/score_a.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ScoreA(Score):
+    def add_columns(self) -> pd.DataFrame:
+        logger.info("Adding Score A")
+        self.df["Score A"] = self.df[
```
lucasmbrown-usds, line 7: Nit: should we move `"Score A"` (and the names of all the other scores) into a field name in the constants?

data/data-pipeline/data_pipeline/score/score.py
```diff
@@ -0,0 +1,14 @@
...
+
+logger = get_module_logger(__name__)
+
+
+class Score:
+    def __init__(self, df: pd.DataFrame) -> None:
+        self.df = df
+
+    def add_columns(self) -> pd.DataFrame:
+        pass
```
lucasmbrown-usds, line 14: nit: want to do a `raise NotImplementedError` or make this abstract method to force all child classes to implement it?

data/data-pipeline/data_pipeline/score/score_c.py
```diff
@@ -0,0 +1,88 @@
+from data_pipeline.score.score import Score, FN, pd, logger
+
+
+class ScoreC(Score):
+    def __init__(self, df: pd.DataFrame) -> None:
+        self.BUCKET_SOCIOECONOMIC = {
```
lucasmbrown-usds, line 6: This is a tiny nit, but I do loved namedtuples for use cases like this where you're naming fields in a dict. It gives you a little more safety and clarity about what's happening (I was confused at first when i saw `bucket["fields"]` below).

data/data-pipeline/data_pipeline/score/score_calculator.py
```diff
@@ -0,0 +1,65 @@
...
+from data_pipeline.utils import get_module_logger
+
+logger = get_module_logger(__name__)
+
+
+class ScoreCalculator:
+    def __init__(self, df: pd.DataFrame):
+        # Define some global parameters
+        self.df = df
+        self.LOW_INCOME_THRESHOLD = 0.60
```
lucasmbrown-usds, line 23: Is this still necessary here?

data/data-pipeline/data_pipeline/score/score_calculator.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def calculate_scores(self) -> pd.DataFrame:
```
lucasmbrown-usds, line 25: woohoooooooo very nice and clean!

data/data-pipeline/data_pipeline/score/score.py
```diff
@@ -0,0 +1,14 @@
...
+from data_pipeline.utils import get_module_logger
+
+logger = get_module_logger(__name__)
+
+
+class Score:
+    def __init__(self, df: pd.DataFrame) -> None:
+        self.df = df
+
+    def add_columns(self) -> pd.DataFrame:
```
lucasmbrown-usds, line 13: I like this method name! Very descriptive of what it's doing.

data/data-pipeline/data_pipeline/score/score_d.py
lucasmbrown-usds: Makes sense!!

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -125,35 +49,32 @@ def add_definition_l_factors(self):
 
         return self.df
 
-    def climate_factor(self) -> bool:
+    def _climate_factor(self) -> bool:
```
lucasmbrown-usds, line 52: Nice making these private!

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -263,22 +179,22 @@ def workforce_factor(self) -> bool:
...
             | (
-                self.df[self.MEDIAN_INCOME_FIELD]
+                self.df[FN.MEDIAN_INCOME_PERCENT_AMI_FIELD]
                 # Note: a high median income as a % of AMI is good, so take 1 minus the threshold to invert it.
                 # and then look for median income lower than that (not greater than).
                 < 1 - self.ENVIRONMENTAL_BURDEN_THRESHOLD
             )
             | (
-                self.df[self.POVERTY_LESS_THAN_100_FPL_FIELD]
+                self.df[FN.POVERTY_LESS_THAN_100_FPL_FIELD]
```
lucasmbrown-usds, line 192: Unfortunately it looks like this points to the wrong field -- `POVERTY_LESS_THAN_100_FPL_FIELD` is `"Percent of individuals < 100% Federal Poverty Line"`, but we actually want `"Percent of individuals < 100% Federal Poverty Line (percentile)"`. Because `POVERTY_LESS_THAN_100_FPL_FIELD` is used in some other scores, we should probably create a new field, like `POVERTY_LESS_THAN_100_FPL_FIELD_PERCENTILE`, or maybe more grammatically accurately,  `POVERTY_LESS_THAN_100_FPL_PERCENTILE_FIELD`. 

Most of these variables are going to point to the percentile, and we'll get pretty inaccurate results if we don't fix that up. 

The ones I know we need to change to a new field name that has ` "percentile"` added to it (I haven't checked all of them): 
* MEDIAN_INCOME_PERCENT_AMI_FIELD
* UNEMPLOYMENT_FIELD
* ASTHMA_FIELD (it looks like ASTHMA_FIELD_PERCENTILE already exists in field_names though)
* DIABETES_FIELD (it looks like DIABETES_FIELD_PERCENTILE already exists in field_names though)
* etc

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -0,0 +1,113 @@
+# Suffixes
```
lucasmbrown-usds, line 1: I know we had talked about maybe having constants set on the ETL class, but I like this approach a lot better. The fields get referenced and handled in lots of places so it makes sense to have them created here! Nice work.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -323,7 +38,27 @@ def extract(self) -> None:
             ejscreen_csv, dtype={"ID": "string"}, low_memory=False
         )
         self.ejscreen_df.rename(
-            columns={"ID": self.GEOID_FIELD_NAME}, inplace=True
+            columns={
+                "ID": self.GEOID_FIELD_NAME,
```
lucasmbrown-usds, line 42: Nice cleanup! 

If you want, we can add a TODO to move this into the EJSCREEN ETL.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -741,120 +248,103 @@ def _prepare_initial_df(self, data_sets: list) -> pd.DataFrame:
...
+            FN.POVERTY_FIELD,
+            FN.HIGH_SCHOOL_ED_FIELD,
+            FN.UNEMPLOYMENT_FIELD,
+            FN.HT_INDEX_FIELD,
+        ]
+        non_numeric_columns = [
+            self.GEOID_FIELD_NAME,
+            FN.PERSISTENT_POVERTY_FIELD,
+        ]
+        columns_to_keep = non_numeric_columns + numeric_columns
```
lucasmbrown-usds, line 304: Looks good!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -741,120 +248,103 @@ def _prepare_initial_df(self, data_sets: list) -> pd.DataFrame:
...
 
             logger.info(
-                f"For data set {data_set.renamed_field}, the min value is {min_value} and the max value is {max_value}."
+                f"For data set {col}, the min value is {min_value} and the max value is {max_value}."
             )
 
-            df[f"{data_set.renamed_field}{self.MIN_MAX_FIELD_SUFFIX}"] = (
-                df[data_set.renamed_field] - min_value
-            ) / (max_value - min_value)
+            df[f"{col}{FN.MIN_MAX_FIELD_SUFFIX}"] = (df[col] - min_value) / (
```
lucasmbrown-usds, line 331: Nice cleanup here!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -741,120 +248,103 @@ def _prepare_initial_df(self, data_sets: list) -> pd.DataFrame:
...
+        # calculate scores
+        self.df = ScoreCalculator(df=self.df).calculate_scores()
 
     def load(self) -> None:
         logger.info("Saving Score CSV")
-        self.SCORE_CSV_PATH.mkdir(parents=True, exist_ok=True)
+        constants.DATA_SCORE_CSV_FULL_DIR.mkdir(parents=True, exist_ok=True)
 
-        self.df.to_csv(self.SCORE_CSV_PATH / "usa.csv", index=False)
+        self.df.to_csv(constants.DATA_SCORE_CSV_FULL_FILE_PATH, index=False)
```
lucasmbrown-usds, line 350: Nice.

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -1,101 +1,25 @@
...
         self.ENVIRONMENTAL_BURDEN_THRESHOLD: float = 0.90
+        super().__init__(df)
 
-    def add_definition_l_factors(self):
-        self.df["Climate Factor (Definition L)"] = self.climate_factor()
-        self.df["Energy Factor (Definition L)"] = self.energy_factor()
+    def add_columns(self) -> pd.DataFrame:
+        logger.info("Adding Score L")
+
+        self.df["Climate Factor (Definition L)"] = self._climate_factor()
```
lucasmbrown-usds, line 13: I'm wondering if these names should also go in the `field_names` constants?

lucasmbrown-usds (changes requested):
I left lots of comments throughout, but overall this look really really great. You've deeply cleaned up the spaghetti and put a lot of nice clean order on top of it. I'm loving how simple the score classes are and they seem very clean. Overall the choices throughout look really good!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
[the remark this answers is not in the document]
switzersc-usds: im fine with that!

data/data-pipeline/data_pipeline/score/score.py
switzersc-usds: should've put this comment on the following line - it's not happy with importing `FN`. @lucasmbrown-usds

data/data-pipeline/data_pipeline/score/score_a.py
switzersc-usds: That might be my lack of experience with python then. If it's more conventional to import them in every class that needs them than to inherit them, I'm fine with that.

data/data-pipeline/data_pipeline/score/score_a.py
switzersc-usds: I thought about that! I opted out of it as out of scope but I'm happy to move them in.

data/data-pipeline/data_pipeline/score/score.py
switzersc-usds: oh yeah that would be better

data/data-pipeline/data_pipeline/score/score_calculator.py
switzersc-usds: nope!

data/data-pipeline/data_pipeline/score/score_l.py
switzersc-usds: done!

data/data-pipeline/data_pipeline/score/score.py
switzersc-usds: done!

data/data-pipeline/data_pipeline/score/score_a.py
switzersc-usds: changed!

data/data-pipeline/data_pipeline/score/score.py
switzersc-usds: moved to child classes

data/data-pipeline/data_pipeline/etl/score/etl_score.py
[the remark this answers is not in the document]
switzersc-usds: changed this!

data/data-pipeline/data_pipeline/score/score_c.py
switzersc-usds: cool! i forgot those existed! done

data/data-pipeline/data_pipeline/score/score_l.py
switzersc-usds: ah okay!

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -205,24 +205,24 @@ def _workforce_factor(self) -> bool:
...
             )
             | (
-                self.df[field_names.LINGUISTIC_ISO_FIELD]
+                self.df[field_names.LINGUISTIC_ISO_PERCENTILE_FIELD]
                 > self.ENVIRONMENTAL_BURDEN_THRESHOLD
             )
         )
         return (
-            self.df[field_names.HIGH_SCHOOL_ED_FIELD] > 0.05
+            self.df[field_names.HIGH_SCHOOL_ED_PERCENTILE_FIELD] > 0.05
```
lucasmbrown-usds, line 227: Ah! This is the rare one where it's actually supposed to stay as `field_names.HIGH_SCHOOL_ED_FIELD`.

lucasmbrown-usds (approved):

PR #830, opened 2021-10-28
Add version number

### LINKED ISSUE #778: Add version number to the download link label on the Explore the tool page
Text should read:

Download the draft v.0.1 list of communities of focus and datasets used. Last updated: 10/01/21.

vim-usds:
closes #778

### DIFF
--- a/client/src/data/copy/explore.tsx
+++ b/client/src/data/copy/explore.tsx
@@ -358,14 +358,14 @@ export const DOWNLOAD_DRAFT = {
     values={{
       downloadDraft:
         <a href={METHODOLOGY_COPY.DOWNLOAD_ZIP_URL}>
-          {'Download the draft list'}
+          {`Download the draft list v${METHODOLOGY_COPY.VERSION_NUMBER}`}
         </a>,
       dateUpdated: METHODOLOGY_COPY.DOWNLOAD_LAST_UPDATED,
     }}
   />,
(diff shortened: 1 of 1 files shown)


client/src/data/copy/explore.tsx
```diff
@@ -358,14 +358,14 @@ export const DOWNLOAD_DRAFT = {
     values={{
       downloadDraft:
         <a href={METHODOLOGY_COPY.DOWNLOAD_ZIP_URL}>
-          {'Download the draft list'}
+          {`Download the draft list v${METHODOLOGY_COPY.VERSION_NUMBER}`}
```
switzersc-usds, line 361: @vim-usds where does this get set?

switzersc-usds (approved):
🚢

client/src/data/copy/explore.tsx
vim-usds: @switzersc-usds , great question. All (if not close to all) of the copy has moved here:

https://github.com/usds/justice40-tool/tree/main/client/src/data/copy

It's organized by page. I was getting confused where piece of copy was. Now it's easier to find and easier to maintain. Also, we don't have to dig through components!

In this particular case, it was decided to place the version number on the methodology page. So the version number is defined here:

https://github.com/usds/justice40-tool/blob/1b17af84c824e1836f461788b8aa32b37e47df28/client/src/data/copy/methodology.tsx#L28

PR #849, opened 2021-11-03
Display score L on map

### LINKED ISSUE #834: Display score L on map
Update map to display score L results.

esfoobar-usds:
Fixes #834


data/data-pipeline/data_pipeline/etl/sources/census/etl_utils.py
```diff
@@ -107,7 +107,7 @@ def check_census_data_source(
         # check if census data is found locally
         if not os.path.isfile(census_data_path / "geojson" / "us.json"):
             logger.info(
-                "No local census data found. Please use '-d aws` to fetch from AWS"
+                "No local census data found. Please use '-s aws` to fetch from AWS"
```
vim-usds, line 110: What does -s and -d do? Is this the same as the AWS CLI commands we run in the github action yml files?

vim-usds (approved):

data/data-pipeline/data_pipeline/etl/sources/census/etl_utils.py
switzersc-usds: -s specifies the data source for the census data (either aws or local). it was -d for a minute but that's confusing with a flag used elsewhere, which specifies a source data set  for indicator data (eg ejscreen or hud)

vim-usds (approved):

PR #863, opened 2021-11-05
Add orientation deck to docs

switzersc-usds:
Adding the new community orientation deck to the repo. We're choosing to store it here so that we can version control it and always have an up to date link to it instead of having to email out a hard copy every time we make updates. 

Major props and thank you to @KristineNixon for getting this started! 

Once we get this merged I'll send out to the group.


esfoobar-usds (approved):
🍏

Issue #864, opened 2021-11-05
Change affordable and sustainable housing formula


BethMattern:
Change the affordable and sustainable housing formula to be as follows:

(pre-1960s >= 90th percentile and median house value < 50th percentile) AND
housing cost burden >= 90th

PR #882, opened 2021-11-09
Add median house value to Definition L

lucasmbrown-usds:
#864. This adds Median house value from Census ACS to the ETL as well as Definition L.


esfoobar-usds (approved):
🍏

Issue #883, opened 2021-11-10
As a score developer, I want to update the FEMA risk index based on guidance from FEMA.


lucasmbrown-usds:
Put in three separate risk index measures for building, population, agriculture. 

Population EAL Rate = Eal Valp / Population
Agriculture EAL Rate = Eal Vala / Agrivalue
divide EAL_VALB (Expected Annual Loss - Building Value) by BUILDVALUE (Building Value ($)).

PR #884, opened 2021-11-10
Issue 883: update FEMA risk index measure

lucasmbrown-usds:
#883. Creates three measures as described by FEMA. 

Put in three separate risk index measures for building, population, agriculture. 

This updates Definition L to include those as well, but that will require further review in comparison tool.


esfoobar-usds (approved):
🍏

PR #885, opened 2021-11-10
Add territory boundary data

### LINKED ISSUE #804: Add territory boundary data from Census to data pipeline
This is the result of the spike #724. 

Tasks:
- [ ] Add fips code to list of census areas to download shapefiles for Census ETL
- [ ] Exclude territory FIPS codes from the ETLs that break because of no data
- [ ] Add territories back at the Score Post step so that they get tiles baked

There is a separate front end issue for showing territory on the map.

esfoobar-usds:
Fixes #804

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/sources/census/etl_utils.py
+++ b/data/data-pipeline/data_pipeline/etl/sources/census/etl_utils.py
@@ -16,13 +16,17 @@ from data_pipeline.utils import (
 logger = get_module_logger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


esfoobar-usds:
@lucasmbrown-usds @switzersc-usds this is ready for review. I have updated the `fips_states_2010.zip` file in S3 with the added territories as well.

ahlusar1989:
Thanks @switzersc-usds and @esfoobar-usds!! I will try again and give y'all an update.

PR #897, opened 2021-11-16
Move survey button to footer

vim-usds:
- remove SurveyFab component as it's no longer a FAB
- create a new SurveyButton component as it's a normal button now
- place button in heading
- add tests
- add pageStyles module


switzersc-usds (approved):
🍏

PR #901, opened 2021-11-17
Use tract instead of block group when calling census API

### LINKED ISSUE #871: Convert Census ACS to tract level

switzersc-usds:
Closes #871

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
+++ b/data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
@@ -100,7 +100,7 @@ class CensusACSETL(ExtractTransformLoad):
 
         self.df = pd.concat(dfs)
 
-        self.df[self.GEOID_FIELD_NAME] = self.df.index.to_series().apply(
+        self.df[self.GEOID_TRACT_FIELD_NAME] = self.df.index.to_series().apply(
             func=self._fips_from_censusdata_censusgeo
         )
 
@@ -179,7 +179,7 @@ class CensusACSETL(ExtractTransformLoad):
         self.OUTPUT_PATH.mkdir(parents=True, exist_ok=True)
 
         columns_to_include = [
-            self.GEOID_FIELD_NAME,
+            self.GEOID_TRACT_FIELD_NAME,
             self.UNEMPLOYED_FIELD_NAME,
             self.LINGUISTIC_ISOLATION_FIELD_NAME,
             self.MEDIAN_INCOME_FIELD_NAME,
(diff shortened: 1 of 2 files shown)


data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -77,7 +77,7 @@ def extract(self) -> None:
                     src="acs5",
                     year=self.ACS_YEAR,
                     geo=censusdata.censusgeo(
-                        [("state", fips), ("county", "*"), ("block group", "*")]
+                        [("state", fips), ("county", "*"), ("tract", "*")]
```
lucasmbrown-usds, line 80: @switzersc-usds -- do we also need to change this line from `etl_score.py`: 

Move `self.census_df` from 

```
census_block_group_dfs = [
            self.ejscreen_df,
            self.census_df,
            self.housing_and_transportation_df,
            self.census_acs_median_incomes_df,
            self.national_risk_index_df,
        ]
```

to move `self.census_df` into `census_tract_dfs`?

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -77,7 +77,7 @@ def extract(self) -> None:
                     src="acs5",
```
lucasmbrown-usds, line 77: Do we need to change `GEOID_FIELD_NAME` to `self.GEOID_TRACT_FIELD_NAME` in the results?

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
switzersc-usds: I was just looking at that. I could do that all at once, or move each df over to the tracts df list as I go. Probably moving them over as I go makes sense, so we're keeping track of them.

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
switzersc-usds: Oh, I didn't realize we had a tract field name. Using that makes sense.

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
lucasmbrown-usds: Right - and moving it as we go means that the score should (🤞 ) still work even as we're moving1

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
switzersc-usds: Updated

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
switzersc-usds: Updated

lucasmbrown-usds (approved):
LGTM! 🎉

PR #903, opened 2021-11-17
Move Housing and Transportation Index to tracts

### LINKED ISSUE #902: Change HT Index to tracts from block groups

switzersc-usds:
Update data download URL to use tract as focus, use tract field name,
and move this dataset to the tracts df list in etl_score.

Closes #902


lucasmbrown-usds (approved):
LGTM! Maybe you already know this, but this is *not* the housing data we use in Definition L, although it is good to update this old data anyways. 

The housing data in Def L comes from `hud_housing` ETL directory.

Issue #751, opened 2021-09-24
Clean up urban/rural ETL 


BethMattern:
Clean up urban/rural ETL by moving loading and transforming code from notebook into the ETL class

lucasmbrown-usds:
While doing this, you can look at information on this ticket for more information about how this data was originally generated: https://github.com/usds/justice40-tool/issues/355#issuecomment-920241787

VincentLaUSDS:
I started a new branch here: https://github.com/usds/justice40-tool/tree/justice40-tool-iss-751 where I at least added a comment to make it clearer.

PR #913, opened 2021-11-19
[ISS-751] Updating Comments

VincentLaUSDS:
Re: https://github.com/usds/justice40-tool/issues/751 I would advocate for merging in this PR which is just adding comments at least to say where the CSV comes from


lucasmbrown-usds (approved):
LGTM! I think it's failing the linter do to very long lines, but thanks for putting this together!

PR #916, opened 2021-11-22
Esfoobar usds/835 census tracts geojson

### LINKED ISSUE #835: [SPIKE] Find new geographic boundary data for census tracts
We want to change our data pipeline to use tracts. Currently the census boundary data we use is only block groups, so we should figure out how to get tract boundary data to use instead. Worst case scenario we use geopandas to merge CBG data from existing source, but this would be time intensive and may have issues.

Make sure this includes territory boundaries.

The output of this spike is either completing this work or creating a new issue with more details as a result of the investigation.

esfoobar-usds:
Fixes #835

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/sources/census/etl.py
+++ b/data/data-pipeline/data_pipeline/etl/sources/census/etl.py
@@ -156,7 +156,7 @@ class CensusETL(ExtractTransformLoad):
             None
         """
         ## write to individual state csv
-        tractd10_list = self.TRACT_PER_STATE[fips_code]
+        tractid10_list = self.TRACT_PER_STATE[fips_code]
         csv_path = self._path_for_fips_file(fips_code, GeoFileType.CSV)
         with open(
             csv_path, mode="w", newline="", encoding="utf-8"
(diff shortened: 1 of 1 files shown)


data/data-pipeline/data_pipeline/etl/sources/census/etl.py
```diff
@@ -153,22 +156,22 @@ def _load_into_state_csvs(self, fips_code: str) -> None:
             None
         """
         ## write to individual state csv
-        geoid10_list = self.CBG_PER_STATE[fips_code]
+        tractd10_list = self.TRACT_PER_STATE[fips_code]
```
lucasmbrown-usds, line 159: Tiny nit: should this be `tractid10_list` instead of `tractd10_list`?

lucasmbrown-usds (approved):
Overall looks very simple and straightforward! 

Am I right in saying basically this PR just changes the URL and the variable names?

data/data-pipeline/data_pipeline/etl/sources/census/etl.py
esfoobar-usds: Absolutely should be!

esfoobar-usds:
> Am I right in saying basically this PR just changes the URL and the variable names?

It does a bit more than that in terms of the data being written, but yes, in summary, we point the Census Tiger data extract to the tract end point and then regenerates all census CSVs to be tract-level.

PR #918, opened 2021-11-23
EJScreen Tracts

### LINKED ISSUE #837: Change EJSCREEN data import to use tract level data
We have draft data at the tract level that we can use until this is formally released in December. We will store the data as a secret in settings for now, although this may break the data pipeline run for open source contributors. This is a temporary situation and should be resolved in Dec.

esfoobar-usds:
Fixes #837


lucasmbrown-usds (approved):
Looking good!

PR #922, opened 2021-11-24
Issue 919: Fix too many tracts issue

lucasmbrown-usds:
It turns out the CNT data had duplicate tract values, so I removed that source since it's no longer in score L. 

Also created some cleaner sanity checking methods around the dataframes.

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/base.py
+++ b/data/data-pipeline/data_pipeline/etl/base.py
@@ -34,7 +34,8 @@ class ExtractTransformLoad:
     GEOID_TRACT_FIELD_NAME: str = "GEOID10_TRACT"
     # TODO: investigate. Census says there are only 217,740 CBGs in the US. This might be from CBGs at different time periods.
     EXPECTED_MAX_CENSUS_BLOCK_GROUPS: int = 250000
-    EXPECTED_MAX_CENSUS_TRACTS: int = 73076
+    # TODO: investigate. Census says there are only 73,057 tracts in the US. This might be from tracts at different time periods.
+    EXPECTED_MAX_CENSUS_TRACTS: int = 74027
 
     def __init__(self, config_path: Path) -> None:
         """Inits the class with instance specific variables"""
--- a/data/data-pipeline/data_pipeline/etl/constants.py
+++ b/data/data-pipeline/data_pipeline/etl/constants.py
@@ -1,10 +1,5 @@
 DATASET_LIST = [
     {
-        "name": "tree_equity_score",
-        "module_dir": "tree_equity_score",
-        "class_name": "TreeEquityScoreETL",
-    },
-    {
         "name": "census_acs",
         "module_dir": "census_acs",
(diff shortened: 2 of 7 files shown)


data/data-pipeline/data_pipeline/etl/constants.py
```diff
@@ -79,6 +69,16 @@
         "module_dir": "census_decennial",
         "class_name": "CensusDecennialETL",
     },
+    {
```
lucasmbrown-usds, line 72: Note: I moved these to the end of the list since they're slow and not used in the score. This way when you run `etl-run`, you'll fill in all the score-necessary data first.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -37,31 +36,9 @@ def extract(self) -> None:
             constants.DATA_PATH / "dataset" / "ejscreen_2019" / "usa.csv"
         )
         self.ejscreen_df = pd.read_csv(
-            ejscreen_csv, dtype={"ID": "string"}, low_memory=False
-        )
-        # TODO move to EJScreen ETL
```
lucasmbrown-usds, line 42: Cleaned this up as the TODO suggests.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -70,26 +47,9 @@ def extract(self) -> None:
         )
         self.census_df = pd.read_csv(
             census_csv,
-            dtype={self.GEOID_FIELD_NAME: "string"},
-            low_memory=False,
-        )
-
-        # Load housing and transportation data
-        housing_and_transportation_index_csv = (
```
lucasmbrown-usds, line 78: Removed this as the offending data frame, since it's not used in score L.

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -34,7 +34,8 @@ class ExtractTransformLoad:
     GEOID_TRACT_FIELD_NAME: str = "GEOID10_TRACT"
     # TODO: investigate. Census says there are only 217,740 CBGs in the US. This might be from CBGs at different time periods.
     EXPECTED_MAX_CENSUS_BLOCK_GROUPS: int = 250000
-    EXPECTED_MAX_CENSUS_TRACTS: int = 73076
+    # TODO: investigate. Census says there are only 73,057 tracts in the US. This might be from tracts at different time periods.
```
esfoobar-usds, line 37: Shouldn't this `TODO` be removed?

esfoobar-usds (approved):
🍏 with one small comment

data/data-pipeline/data_pipeline/etl/base.py
lucasmbrown-usds: No I just added it! EXPECTED_MAX_CENSUS_TRACTS is about 90 tracts higher than what census says we should expect.

PR #923, opened 2021-11-24
County mapping to Tracts

### LINKED ISSUE #874: Convert county mapping to use tracts, not CBGs
We should look into whether we can use Geocorr (which we use for other mapping earlier in the process). Maybe we  don't want to do that though in case we drop Geocorr in the future. Whoever takes this should figure this out by leading the covnersation and understanding what we get from Geocorr and when.

esfoobar-usds:
Fixes #874


data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -58,12 +60,14 @@ def _extract_score(self, score_path: Path) -> pd.DataFrame:
...
+    def _extract_national_tract(
+        self, national_tract_path: Path
+    ) -> pd.DataFrame:
+        logger.info("Reading national tract file")
         return pd.read_csv(
-            national_cbg_path,
-            names=["GEOID10"],
-            dtype={"GEOID10": "string"},
+            national_tract_path,
+            names=[self.GEOID_TRACT_FIELD_NAME],
```
lucasmbrown-usds, line 69: Love seeing all these constants change, so clean and clear.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -150,15 +155,19 @@ def _create_score_data(
...
         # merge state + county with score
         score_county_state_merged = score_df.merge(
-            county_state_merged, on="GEOID", how="left"
+            county_state_merged,
+            on="GEOID",  # GEOID is the county ID
+            how="left",
         )
 
         # check if there are census cbgs without score
         logger.info("Removing CBG rows without score")
```
lucasmbrown-usds, line 164: nit: should we switch all references to CBGs to tracts?

lucasmbrown-usds (approved):
TBH, I don't fully understand each aspect of the work in these parts of the pipelines, but the code LGTM.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
esfoobar-usds: Yes inded

PR #926, opened 2021-11-26
Tract work score geo

### LINKED ISSUE #925: Tract Work for Score Geo
The score-geo script needs to be updated to use tracts instead of CBGs.

esfoobar-usds:
Fixes #925


lucasmbrown-usds (approved):
LGTM, very simple!

PR #933, opened 2021-11-30
Issue 932 fix too many tracts in ACS join

lucasmbrown-usds:
Because of where this sneaky little `append` was in the code, if a call to the Census API failed (as it does for places like Guam and Virgin Islands), then it still appends the most recent value of `response`, which in this case was the data for Wyoming and Puerto Rico, which got appended multiple times.


vim-usds (approved):
Nice catch!!

lucasmbrown-usds:
FYI I think this bug has been around for a while, and it's only my recent addition of that sanity check to the code that caught it.

Issue #940, opened 2021-11-30
Update link for accessing FEMA data


BethMattern:
We need to get the data from FEMA here:
https://hazards.fema.gov/nri/Content/StaticDocuments/DataDownload//NRI_Table_CensusTracts/NRI_Table_CensusTracts.zip

PR #952, opened 2021-12-02
Revisions to FEMA measure and new link for FEMA data

### LINKED ISSUE #920: As a data developer, I want to make updates to the FEMA measures, so they more accurately capture climate change risk.
**Description**
Two tasks related to FEMA risk index. Data dictionary available online or here: [NRIDataDictionary.csv](https://github.com/usds/justice40-tool/files/7597458/NRIDataDictionary.csv).

~~First, drop the buildings and agriculture measures from Definition L, just leaving behind population (injuries and fatalities).~~

First, update to point to the *new* data we got from FEMA. 

Recalculate the measures in FEMA risk index to only include those natural disasters that have been linked in any association with climate change. So this means removing the following disaster types: 

* Earthquake 
* Tsunami
* Volcano 
* Lightning 

For all remaining disaster types, calculate a single combined population risk rate with the following formula: 
    
    numerator: the sum of each field "{disaster type}  - Expected Annual Loss - Population" such as field `AVLN_EALP`.
    
    denominator: "Population"

For all remaining disaster types, calculate a single combined agriculture value risk rate with the following formula: 

    numerator: the sum of each field "{disaster type}  - Expected Annual Loss - Agricultural Value" such as field `DRGT_EALA`.
    
    denominator: "Agricultural Value ($)" aka `AGRIVALUE`

For all remaining disaster types, calculate a single combined building value risk rate with the following formula: 

    numerator: the sum of each field "{disaster type}  - Expected Annual Loss - Building Value" such as field `EAL_VALB`.
    
    denominator: "Building Value ($)" aka `BUILDVALUE`

We may not use all of these, but we should go ahead and calculate them.

Note that not every disaster type has every loss type. E.g., avalanche may not have an agricultural value measure. This is fine and it can just be skipped.

ahlusar1989:
Fixes #920 and #940

@switzersc-usds  and @lucasmbrown-usds I am not sure if this is necessarily correct. 

I took this "numerator: the sum of each field "{disaster type}  - Expected Annual Loss - Agricultural Value" such as field `DRGT_EALA`." to mean "for a given tract aggregate the expected annual loss". The other divisor is self-explanatory.


data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -93,33 +96,85 @@ def transform(self) -> None:
...
+            "SWND",  # Strong Wind
+            "TRND",  # Tornado
+            "WFIR",  # Wildfire
+            "WNTW",  # Winter Weather
+        ]
+
+        # Some disaster categories do not have agriculture value column
+        agri_columns = [str(x) + "_EALA" for x in disaster_categories if str(x) + "_EALA" in list(df_nri.columns)]
+
+        disaster_population_sum_series = df_nri[
```
ahlusar1989, line 126: Numerators

saran-ahluwalia:
@lucasmbrown-usds There is a unit test failing related to GEOID10 not being in the index. I am a bit puzzled as the integration test seems to work when I run things locally.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -52,7 +54,7 @@ def __init__(self):
 
         # Note: also need to edit transform step to add fields to output.
         self.COLUMNS_TO_KEEP = [
-            self.GEOID_TRACT_FIELD_NAME,
```
saran-ahluwalia, line 55: Upstream changes.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -82,6 +84,7 @@ def transform(self) -> None:
         logger.info("Transforming National Risk Index Data")
 
         NRI_TRACT_COL = "TRACTFIPS"  # Census Tract Column in NRI data
+        TRACT_COL = self.GEOID_TRACT_FIELD_NAME  # Census Tract column name
```
lucasmbrown-usds, line 87: Why are you going back to using TRACT_COL? Are you sure you branched off of `mega-tracts` / the latest on `main`?

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
saran-ahluwalia: Not sure if this is related

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -93,33 +96,92 @@ def transform(self) -> None:
...
+            "LNDS",  # Landslide
+            "RFLD",  # Riverine Flooding
+            "SWND",  # Strong Wind
+            "TRND",  # Tornado
+            "WFIR",  # Wildfire
+            "WNTW",  # Winter Weather
+        ]
+
+        # Some disaster categories do not have agriculture value column
+        agri_columns = [str(x) + "_EALA" for x in disaster_categories if str(x) + "_EALA" in list(df_nri.columns)]
```
lucasmbrown-usds, line 124: Oooh very stylish!! 🎉

lucasmbrown-usds: Nit: instead of `str(x)` you could just `f"{x}_EALA"`

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -93,33 +96,92 @@ def transform(self) -> None:
...
         df_nri[self.EXPECTED_BUILDING_LOSS_RATE_FIELD_NAME] = (
-            df_nri[self.EXPECTED_ANNUAL_LOSS_BUILDING_VALUE_INPUT_FIELD_NAME]
-            / df_nri[self.BUILDING_VALUE_INPUT_FIELD_NAME]
+            disaster_building_sum_series / df_nri[self.BUILDING_VALUE_INPUT_FIELD_NAME]
+        )
+
+        # Reduce columns.
+        # Note: normally we wait until writing to CSV for this step, but since the file is so huge,
+        # move this up here for performance reasons.
+        df_nri = df_nri[
```
lucasmbrown-usds, line 164: Yeah, this also looks like old code, making me think you branched off an older version of this repo. This is no longer necessary now that we're at the tract level.

lucasmbrown-usds (changes requested):
Overall this looks awesome!! Some super elegant code along the way, too, major style points. 

I left a couple of nits. The major comment is that it looks like you branched off some old code that's been updated since mega-tracts, so maybe you want to rebase off `main`. 

Also, we're planning to update our reference data (see https://github.com/usds/justice40-tool/issues/940). You could do that as part of this PR or as a separate PR!

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -93,33 +96,92 @@ def transform(self) -> None:
...
         )
 
         # divide EAL_VALB (Expected Annual Loss - Building Value) by BUILDVALUE (Building Value ($)).
         df_nri[self.EXPECTED_BUILDING_LOSS_RATE_FIELD_NAME] = (
-            df_nri[self.EXPECTED_ANNUAL_LOSS_BUILDING_VALUE_INPUT_FIELD_NAME]
-            / df_nri[self.BUILDING_VALUE_INPUT_FIELD_NAME]
+            disaster_building_sum_series / df_nri[self.BUILDING_VALUE_INPUT_FIELD_NAME]
+        )
+
+        # Reduce columns.
```
lucasmbrown-usds, line 161: I think you can delete 161 - 184.

ahlusar1989: Done

saran-ahluwalia:
> 

I changed the file path to address #940

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -99,28 +101,63 @@ def transform(self) -> None:
...
+
+        buildings_columns = [f"{x}_EALB" for x 
+            in disaster_categories if f"{x}_EALB" in list(df_nri.columns)]
+
+        disaster_population_sum_series = df_nri[
+            population_columns
+        ].sum(axis=1)
+
+        disaster_agriculture_sum_series = df_nri[
+            agri_columns
```
lucasmbrown-usds, line 138: This is super nitpicky, so feel free to ignore, but for readability do we want to make sure all the variable names match with the same way of describing the category? 

E.g., `disaster_agriculture_sum_series` and `agriculture_columns` might be more consistent than `agri_columns`. 

And `disaster_building_sum_series` has singular building compared to `buildings_columns`. 

But feel free to ignore, this is super nitpicky.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -61,6 +61,8 @@ def test_transform(self, mock_etl):
             dtype={BLOCK_COL: "string", TRACT_COL: "string"},
         )
 
+        etl.df = expected
+
```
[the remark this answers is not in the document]
lucasmbrown-usds: Hmm, I'm not sure you've updated these tests correctly. I would expect you to need to update the three CSVs in the tests folder. 

I'm also not sure what your intention is here – why would you set `etl.df = expected`? `DATA_DIR / "transform.csv"` is the expected output of the `etl.transform()` step, so you want to check that the results are equal. 

I think what you want to do is temporarily, after these lines, 

```
        # execution
        etl.transform()
```

Write the output to CSV and then use that to replace the text fixture in `DATA_DIR / "transform.csv"`. Once you've replaced the text fixture, you can delete the lines that write the output to CSV.

lucasmbrown-usds (changes requested):
Overall looking good, but it seems like you forced the test to pass rather than actually updating the test fixtures in the CSVs. Let me know if you want to pair on that part.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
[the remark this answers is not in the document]
ahlusar1989: @lucasmbrown-usds :Ah - thanks - yeah - root cause was the deletion of  `self.df = df_nri` in line 122 which made df no loner an attribute (hence, no longer accessible) in the unit test.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/input.csv
```diff
@@ -1,6 +1,3 @@
...
...RISKS,RFLD_RISKR,SWND_EVNTS,SWND_AFREQ,SWND_EXPB,SWND_EXPP,SWND_EXPPE,SWND_EXPA,SWND_EXPT,SWND_HLRB,SWND_HLRP,SWND_HLRA,SWND_HLRR,SWND_EALB,SWND_EALP,SWND_EALPE,SWND_EALA,SWND_EALT,SWND_EALS,SWND_EALR,SWND_RISKS,SWND_RISKR,TRND_EVNTS,TRND_AFREQ,TRND_EXPB,TRND_EXPP,TRND_EXPPE,TRND_EXPA,TRND_EXPT,TRND_HLRB,TRND_HLRP,TRND_HLRA,TRND_HLRR,TRND_EALB,TRND_EALP,TRND_EALPE,TRND_EALA,TRND_EALT,TRND_EALS,TRND_EALR,TRND_RISKS,TRND_RISKR,TSUN_EVNTS,TSUN_AFREQ,TSUN_EXPB,TSUN_EXPP,TSUN_EXPPE,TSUN_EXPT,TSUN_HLRB,TSUN_HLRP,TSUN_HLRR,TSUN_EALB,TSUN_EALP,TSUN_EALPE,TSUN_EALT,TSUN_EALS,TSUN_EALR,TSUN_RISKS,TSUN_RISKR,VLCN_EVNTS,VLCN_AFREQ,VLCN_EXPB,VLCN_EXPP,VLCN_EXPPE,VLCN_EXPT,VLCN_HLRB,VLCN_HLRP,VLCN_HLRR,VLCN_EALB,VLCN_EALP,VLCN_EALPE,VLCN_EALT,VLCN_EALS,VLCN_EALR,VLCN_RISKS,VLCN_RISKR,WFIR_EVNTS,WFIR_AFREQ,WFIR_EXPB,WFIR_EXPP,WFIR_EXPPE,WFIR_EXPA,WFIR_EXPT,WFIR_HLRB,WFIR_HLRP,WFIR_HLRA,WFIR_HLRR,WFIR_EALB,WFIR_EALP,WFIR_EALPE,WFIR_EALA,WFIR_EALT,WFIR_EALS,WFIR_EALR,WFIR_RISKS,WFIR_RISKR,WNTW_EVNTS,WNTW_AFREQ,WNTW_EXPB,WNTW_EXPP,WNTW_EXPPE,WNTW_EXPA,WNTW_EXPT,WNTW_HLRB,WNTW_HLRP,WNTW_HLRA,WNTW_HLRR,WNTW_EALB,WNTW_EALP,WNTW_EALPE,WNTW_EALA,WNTW_EALT,WNTW_EALS,WNTW_EALR,WNTW_RISKS,WNTW_RISKR,NRI_VER
```
lucasmbrown-usds, line 1: Do you mind if we do like 10-20 rows, for robustness?

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/output.csv
```diff
@@ -1,6 +1,3 @@
 GEOID10_TRACT,FEMA Risk Index Expected Annual Loss Score,Expected population loss rate (Natural Hazards Risk Index),Expected agricultural loss rate (Natural Hazards Risk Index),Expected building loss rate (Natural Hazards Risk Index)
-05007040300,11.5,0.241594948,0.206607506,0.399500316
-05001020100,12.5,0.281343259,0.207119742,0.535089827
-15007040500,13.5,0.280726185,0.269219337,0.493093767
-15001021010,14.5,0.221568983,0.360886597,0.305051347
-15001021101,15.5,0.805488269,0.204161204,0.30618612
+15007040300,18.619940187472693,3.5067096117382227e-06,0.0087782351557864,7.19506190745842e-05
```
lucasmbrown-usds, line 2: Huh, why are these in scientific notation?

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/transform.csv
```diff
@@ -1,6 +1,3 @@
...
...688884e-06,1.829685538e-06,7.2723262261e-05,Very Low,92.36922872583236,0.000479434788604,3643.704393392753,0.336028207064017,3736.4096503256496,14.973490276784354,Relatively Low,16.607054548516683,Relatively Low,0.0,6.5330996904e-05,912658000.0,8385.0,63726000000.0,147860.5647200878,64638805860.56472,0.008966238954864,5.978380718e-06,0.002107946289716,Very Low,534.6107152638473,3.274959403e-06,24.889691462507123,0.020362504191456,559.5207692305459,5.870692520235838,Very Low,6.7469108144685,Very Low,7.0,0.031969309,198555247.5326174,978.4678896234336,7436355961.138096,7634911208.670713,0.001559313957881,3.873415999e-06,Very Low,9898.016764864928,0.000121164102392,920.8471781755048,10818.863943040433,23.658087226460832,Relatively High,20.211588413559333,Relatively Moderate,0.0,0.0,0.0,0.0,0.0,0.0,0.000541106984497,3.737122925e-06,No Rating,0.0,0.0,0.0,0.0,0.0,No Expected Annual Losses,0.0,No Rating,,,,,,,,,,,Insufficient Data,,,,,,,Insufficient Data,,Insufficient Data,0.0,0.0,0.0,0.0,0.0,0.0,0.0,8.5057095e-08,1.05693624e-07,1.782e-12,No Rating,0.0,0.0,0.0,0.0,0.0,0.0,No Expected Annual Losses,0.0,No Rating,November 2021,3.5067096117382227e-06,0.008778235155786459,7.19506190745842e-05
```
lucasmbrown-usds, line 2: Same question about scientific notation here.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -91,9 +92,10 @@ def test_load(self, mock_etl):
         expected = pd.read_csv(DATA_DIR / "output.csv", dtype={BLOCK_COL: str})
         # execution
         etl.load()
-        output = pd.read_csv(output_path, dtype={BLOCK_COL: str})
 
+        output = pd.read_csv(output_path, dtype={BLOCK_COL: str})
```
lucasmbrown-usds, line 96: i think this `BLOCK_COL` should be replaced everywhere with the census tract column instead. 

That is on me though! I'm the one who didn't update these tests enough when switching to tracts.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -56,6 +56,7 @@ def test_transform(self, mock_etl):
         # setup - read in sample output as dataframe
```
lucasmbrown-usds, line 56: I can't highlight the lines above, but could we get rid of everything that references blocks? E.g., `acs_dst = etl.BLOCK_GROUP_CSV`. 

As with the other comment, this is my fault! I'm the one who didn't update these tests enough when switching to tracts.

lucasmbrown-usds (changes requested):

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/output.csv
ahlusar1989: Good question - I thought that since this was just an artifact of numerical instability after the division operation. I did not see setting the numpy floating point limit or ceiling. We could coerce like this `math.ceil((some_ridiculous number * 1e15)/1e15)`, but that just looks cheap to me.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/input.csv
ahlusar1989: Sure thing

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -99,28 +101,71 @@ def transform(self) -> None:
...
+            "ISTM",  # Ice Storm
+            "LNDS",  # Landslide
+            "RFLD",  # Riverine Flooding
+            "SWND",  # Strong Wind
+            "TRND",  # Tornado
+            "WFIR",  # Wildfire
+            "WNTW",  # Winter Weather
+        ]
+
+        self.df = df_nri
```
lucasmbrown-usds, line 122: Hmm, this seems totally off to me. Why did you move this up here from line 122? You're essentially ensuring that none of your work gets saved?

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/output.csv
lucasmbrown-usds: I think we can just fix the `to_csv` parameters: https://stackoverflow.com/a/23006399/1569221

lucasmbrown-usds (changes requested):
LGTM, except for the move of `self.df = df_nri` which I don't really understand?

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
ahlusar1989: @lucasmbrown-usds Good catch - I forgot to move that back down!!

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/output.csv
ahlusar1989: This is still posing problems. We do this as well to standardize: 

`df_data = df_data.round(3)`
`df_data.to_csv('somename.csv')`

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
ahlusar1989: That's fine - as always - the main work is done but the unit tests hold us up 😮‍💨

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/output.csv
ahlusar1989: I should have added this as well.

I added this to check why the assertion call keeps failing.

```
        difference_locations = np.where(etl.df != expected)
        changed_from = etl.df.values[difference_locations]
        changed_to = expected.values[difference_locations]
        df_2 = pd.DataFrame({'from': changed_from, 'to': changed_to}, index=np.arange(len(changed_to)))
        print(df_2)
```

The `stdout` is attached below. Notice that visually the differences are not apparent to the human eye using the suggestion provided in that post shared above. The load test is just fine by the way, which is why I am at a complete loss as to how to "fix" the test. That's what is blocking me. 

Screen Shot 2021-12-03 at 7 10 34 AM

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -49,24 +49,29 @@ def test_transform(self, mock_etl):
         input_src = DATA_DIR / "input.csv"
         input_dst = etl.INPUT_CSV
         acs_src = DATA_DIR / "acs.csv"
+        # Lucas raised this as a lingering artifact. We have a
```
lucasmbrown-usds, line 52: Can you just delete `(acs_src, acs_dst)`? I think it'll all work fine to just do `for src, dst in [(input_src, input_dst)]:` since you no longer need that data file to be copied.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/output.csv
```diff
@@ -1,6 +1,56 @@
 GEOID10_TRACT,FEMA Risk Index Expected Annual Loss Score,Expected population loss rate (Natural Hazards Risk Index),Expected agricultural loss rate (Natural Hazards Risk Index),Expected building loss rate (Natural Hazards Risk Index)
```
lucasmbrown-usds, line 1: Is this what the ETL outputs now? I'm sorry to be super nitpicky but I'm wondering if we can write a lot more digits here (like 10? idk) because the `Expected population loss rate` is so low that all the readings are clumped together at 6 significant digits.

lucasmbrown-usds (changes requested):

ahlusar1989:
@lucasmbrown-usds Regarding the tolerance for asserting two NumPy arrays is equal, I am not clear on how 10 digits versus 6 digits will affect the numerical precision or have a material impact on scoring and eligibility? 

My thought from a numerical precision perspective is that we clamp across the board in the class method to something that is human-readable, in conjunction with counting the number of eligible factors. I welcome your thoughts.

esfoobar-usds:
One quick point of feedback: I personally think it's useful to title the PRs with a summary of what the PR is about and leave the issue you're resolving as part of the description, i.e. "Fixes #920"

ahlusar1989:
For some reason - I can't comment inline to [this](https://github.com/usds/justice40-tool/pull/952#discussion_r762013639). Yes - this is the output. That is an artifact of the following suggestion in that link you shared earlier:

`        output.to_csv("output_2.csv", float_format='{:f}'.format, encoding='utf-8', index=False)
`

 I honestly think that to reduce the precision we can either just floor or clamp the floating-point precision to three digits if aesthetics and human-readability is required.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -173,5 +173,5 @@ def load(self) -> None:
         # write nationwide csv
         self.OUTPUT_DIR.mkdir(parents=True, exist_ok=True)
         self.df[self.COLUMNS_TO_KEEP].to_csv(
-            self.OUTPUT_DIR / "usa.csv", index=False
+            self.OUTPUT_DIR / "usa.csv", index=False, float_format="%.10f"
```
lucasmbrown-usds, line 176: 🎉

lucasmbrown-usds (approved):
Looks good! I think we're done!!

PR #956, opened 2021-12-03
Add 4 additional territory buttons

### LINKED ISSUE #910: Add shortcuts for territories on map
For MVP, we want to make sure territories are included and easy to navigate to. For now, let's add shortcuts in the same pattern we already have. Very soon we will update this with new designs so that we don't have one long list of shortcuts slapped to the side of the page. We know that this will also cause a poorer experience on mobile, hence prioritizing improving this very soon after adding the initial shortcuts.

The territories are 
- Puerto Rico (PR) already implemented
- Guam (GU)
- American Samoa (AS)
- Commonwealth of Northern Mariana Islands (MP)
- US Virgin Islands (VI)

vim-usds:
Closes #910


switzersc-usds (approved):
🚢

switzersc-usds:
This looks good code-wise, but I just noticed on the QA branch, the shortcuts have what looks like a box around the initials:

image

Issue #936, opened 2021-11-30
Add proximity to hazardous waste sites to Remediation and reduction of legacy pollution formula for score L


BethMattern:
Change the formula for Remediation and reduction of legacy pollution in score L to be:

Proximity to RMP sites ≥ 90th OR
Proximity to NPL sites (superfund sites) ≥ 90th OR
Proximity to hazardous waste sites ≥ 90th

ahlusar1989:
@BethMattern @lucasmbrown-usds @switzersc-usds 

For hazardous waste sites, it may be helpful to add this [reference](https://oehha.ca.gov/media/downloads/calenviroscreen/report/ces3report.pdf)

PR #959, opened 2021-12-03
Integrate proximity to waste sites into pollution factors

ahlusar1989:
Addresses #936


data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -205,10 +205,13 @@ def _pollution_factor(self) -> bool:
 
         pollution_criteria = (
             self.df[field_names.RMP_FIELD + field_names.PERCENTILE_FIELD_SUFFIX]
-            > self.ENVIRONMENTAL_BURDEN_THRESHOLD
+            >= self.ENVIRONMENTAL_BURDEN_THRESHOLD
```
lucasmbrown-usds, line 208: This change from `>` to `>=` is fine, but if we do that, we should do it on all the criteria throughout Score L. Do you mind doing that? In practice I don't think it'll matter much because there are rarely exact ties but we should be consistent.

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -205,10 +205,13 @@ def _pollution_factor(self) -> bool:
...
         ) | (
             self.df[field_names.NPL_FIELD + field_names.PERCENTILE_FIELD_SUFFIX]
-            > self.ENVIRONMENTAL_BURDEN_THRESHOLD
+            >= self.ENVIRONMENTAL_BURDEN_THRESHOLD
+        ) | (
+            self.df[field_names.TSDF_FIELD + field_names.PERCENTILE_FIELD_SUFFIX]
```
lucasmbrown-usds, line 213: 🎉

lucasmbrown-usds (changes requested):
Overall looks good!!

data/data-pipeline/data_pipeline/score/score_l.py
ahlusar1989: @lucasmbrown-usds I am adhering to what @BethMattern stated in the issue. Your logic for across-the-board changes is fine - I just wanted to bring that to your attention.

lucasmbrown-usds: Sounds good. Can you go ahead and make this change everywhere then in Score L, for consistency?

ahlusar1989: Yup

lucasmbrown-usds (approved):
Looks great, thanks!

Issue #921, opened 2021-11-24
As a score developer, I want to list the thresholds exceeded by a community in Definition L.


lucasmbrown-usds:
**Description**
Right now definition L produces a single binary output variable. 

In addition to that, we want to identify the individual thresholds exceeded by a census tract. 

By individual threshold, we mean a criteria like `diabetes >= 90th percentile AND % low income >= 65th percentile`. For categories with more than one criteria in them, like health burdens which contains criteria related to diabetes, asthma, heart disease, and life expectancy, each criteria will be handled individually

Whether or not a census tract exceeds a threshold will need to be included in the .xls file, so there will need to be a column with a yes/no value for each individual threshold.

Design is still considering how this might be visualized in the UI.

List of all thresholds:
Climate Change:
- IF at or above the 90th percentile for FEMA’s expected population loss rate AND is low income 

Clean Energy and Energy Efficiency:
- IF at or above the 90th percentile for energy cost burden score And is low income
- PM2.5 and is low income 

Clean Transportation: 
- IF at or above the 90th percentile for diesel particulate matter And is low income
- traffic proximity and volume And is low income
 
Affordable and Sustainable Housing:
- IF at or above the 90th percentile for lead paint AND the median house value is less than 90th percentile
- IF at or above the 90th percentile for housing cost burden And is low income

Remediation and Reduction of Legacy Pollution :
- IF at or above the 90th percentile for proximity to RMP sites And is low income
- IF at or above 90th percentile for proximity to superfund sites And is low income
- IF at or above 90th percentile for proximity to hazardous waste facilities And is low income

Critical Clean Water and Waste Infrastructure:
- IF at or above the 90th percentile for wastewater discharge And is low income

Health Burdens:
-  IF at or above the 90th percentile for diabetes And is low income
- IF at or above the 90th percentile for asthma And is low income
- IF at or above the 90th percentile for heart disease And is low income
- lIF at or above the 90th percentile fo rife expectancy And is low income

Training and Workforce Development:
- IF at or above the 90th percentile for unemployment and Where the high school degree achievement rate for adults 25 years and older is less than 90%
- IF at or above the  90th percentile for Linguistic isolation and Where the high school degree achievement rate for adults 25 years and older is less than 90%
- IF at or above the 90th percentile for % households at or below 100% federal poverty level Where the high school degree achievement rate for adults 25 years and older is less than 90%
[...]

Issue #844, opened 2021-11-02
Generate score for territories


BethMattern:
While there is a lot of data used by Score L that is not available for territories, we should be able to utilize the 'training and workforce development' portion of the score to determine priority census tracts for territories.

switzersc-usds:
After we implement this, we will have a conversation early next week about whether this makes sense and what next steps are.

PR #969, opened 2021-12-03
Create indicators for all thresholds exceeded by a community in Definition L

ahlusar1989:
Addresses #921 

@lucasmbrown-usds  - Not happy about this one. I have some ideas to reduce the amount of boilerplate but wanted to just ensure this meets the basic requirements we discussed this morning.


data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -88,40 +110,59 @@ def _climate_factor(self) -> bool:
...
+        )
+
+        threshold_one = (
             self.df[
                 field_names.ENERGY_BURDEN_FIELD
                 + field_names.PERCENTILE_FIELD_SUFFIX
             ]
             >= self.ENVIRONMENTAL_BURDEN_THRESHOLD
-        ) | (
+        ) & FPL_200
```
saran-ahluwalia, line 137: AH!!!!

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -160,3 +160,51 @@
 EJSCREEN_AREAS_OF_CONCERN_STATE_95TH_PERCENTILE_COMMUNITIES_FIELD_NAME = (
     "EJSCREEN Areas of Concern, State, 95th percentile (communities)"
 )
+
+# Community Threshold Field Names for Derived columns (Issue 921)
+
+# Climate Change
+FEMA_LOSS_RATE_LOW_INCOME = "At or above the 90th percentile for FEMA's expected loss rate and is low income"
```
lucasmbrown-usds, line 167: Can we programatically create these using f-strings, so they update automatically if we change the threshold? 

That'll also help with consistency (e.g., the PM2.5 field has a different name style than every other field).

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -160,3 +160,51 @@
...
+    "At or above the 90th percentile for diabetes and is low income"
+)
+ASTHMA_LOW_INCOME = (
+    "At or above the 90th percentile for asthma and is low income"
+)
+HEART_DISEASE_LOW_INCOME = (
+    "At or above the 90th percentile for heart disease and is low income"
+)
+LIFE_EXPECTANCY_INCOME = (
+    "At or above the 90th percentile fo life expectancy and is low income"
```
lucasmbrown-usds, line 196: another example error that would be cleaned up by creating them programmatically

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -160,3 +160,51 @@
...
+)
+# Workforce
+UNEMPLOYMENT_HIGH_SCHOOL_ATTAINMENT = """At or above the 90th percentile"
```
lucasmbrown-usds, line 199: nit: can we simplify this to just `and low HS education`?

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -64,6 +65,27 @@ def _climate_factor(self) -> bool:
         # Low income: In 60th percentile or above for percent of block group population
         # of households where household income is less than or equal to twice the federal
         # poverty level. Source: Census's American Community Survey]
+
+        # I am thinking that we make constants for these since these thresholds are used
```
lucasmbrown-usds, line 69: Yeah, agreed. Let's just create it once. Maybe add a _create_low_income_threshold method that runs before anything else and generates this field, and then we can just reference the field?

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -64,6 +65,27 @@ def _climate_factor(self) -> bool:
...
+
+        did_we_meet_threshold = (
+            self.df[
+                field_names.EXPECTED_POPULATION_LOSS_RATE_FIELD_NAME
+                + field_names.PERCENTILE_FIELD_SUFFIX
+            ]
+            >= self.ENVIRONMENTAL_BURDEN_THRESHOLD
+        ) & FPL_200
+
+        self.df[field_names.FEMA_LOSS_RATE_LOW_INCOME] = did_we_meet_threshold
```
lucasmbrown-usds, line 87: nit: `did_we_meet_threshold` is a fairly vague name since it doesn't reference population loss. 

I'm wondering, can you just set: 

```
self.df[field_names.FEMA_LOSS_RATE_LOW_INCOME] = (
            self.df[
                field_names.EXPECTED_POPULATION_LOSS_RATE_FIELD_NAME
                + field_names.PERCENTILE_FIELD_SUFFIX
            ]
            >= self.ENVIRONMENTAL_BURDEN_THRESHOLD
        ) & FPL_200
```

and skip the assignment?

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -64,6 +65,27 @@ def _climate_factor(self) -> bool:
...
+
         climate_criteria = (
             (
                 self.df[
```
lucasmbrown-usds, line 91: I would strongly vote that we do *not* maintain redundant code with the same logic here. For instance, your logic only checks population loss, but this logic checks 3 FEMA fields. 

Can you just create `self.df[field_names.FEMA_LOSS_RATE_LOW_INCOME]` and the other two fields, and then the result returned is `self.df[field_names.FEMA_LOSS_RATE_LOW_INCOME] | <other field> | <other field>`?

lucasmbrown-usds (changes requested):
Thanks for diving into this. Overall I don't really understand approaching this by maintaining a lot of duplicate code. Why create two series with the same logic, when you could just create one series and use it twice? This is the same feedback as I had when we initially discussed this approach verbally.

data/data-pipeline/data_pipeline/score/score_l.py
saran-ahluwalia: Yup - my idea exactly

data/data-pipeline/data_pipeline/score/score_l.py
saran-ahluwalia: I'll look at this more closely

data/data-pipeline/data_pipeline/score/score_l.py
saran-ahluwalia: Yeah, exactly, that was my thought as well. I was planning to do that after your reviewed the core logic.

data/data-pipeline/data_pipeline/score/field_names.py
saran-ahluwalia: @lucasmbrown-usds - Right, as I mentioned in the description, I am not happy either about the amount of boilerplate. 
I have some ideas to reduce the amount of boilerplate but wanted to just ensure this meets the basic requirements we discussed this morning.

data/data-pipeline/data_pipeline/score/field_names.py
ahlusar1989: In hindsight, I couple use the template strings. I was just trying to follow precedent and a pattern I see throughout - other than #844.

data/data-pipeline/data_pipeline/score/field_names.py
ahlusar1989: Yup

data/data-pipeline/data_pipeline/score/score_l.py
ahlusar1989: Upon, review, that's certainly an idea I can roll with. It's just that in the event the interpreter hits a compilation error I know it's the variable instantiated rather than the assignment in which then I have to consider both the left and right - hand side.

data/data-pipeline/data_pipeline/score/score_l.py
saran-ahluwalia: @lucasmbrown-usds Just to be clear for FEMA there is only 1 of the 3 that is asked to be added as an indicator:

_Climate Change_:

_IF at or above the 90th percentile for FEMA’s expected population loss rate AND is low income_

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -93,6 +94,22 @@ def _combine_island_areas_with_states_and_set_thresholds(
 
         return df, threshold_column_name
 
+    def create_low_income_threshold(
```
lucasmbrown-usds, line 97: Nit: let's make this `_create_low_income_threshold` to indicate that it's a private method that shouldn't be used outside of the class. https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name#:~:text=Python%20doesn't%20have%20real,not%20part%20of%20the%20API.

data/data-pipeline/data_pipeline/score/field_names.py
lucasmbrown-usds: Any reason not to do this now?

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -188,3 +188,51 @@
...
+    "At or above the 90th percentile for asthma and is low income"
+)
+HEART_DISEASE_LOW_INCOME = (
+    "At or above the 90th percentile for heart disease and is low income"
+)
+LIFE_EXPECTANCY_INCOME = (
+    "At or above the 90th percentile for life expectancy and is low income"
+)
+# Workforce
+UNEMPLOYMENT_LOW_HS_EDUCATION = """At or above the 90th percentile"
```
lucasmbrown-usds, line 227: I think I left a comment here asking if we could make the latter half of this something short, like `and has low HS achievement`

data/data-pipeline/data_pipeline/score/score_l.py
lucasmbrown-usds: Otherwise looks great!! Nice docstrings 😄

data/data-pipeline/data_pipeline/score/score_l.py
lucasmbrown-usds: Yea, unfortunately this is a miscommunication, I think the ticket may be out of date. For now, make sure all 3 FEMA indicators are represented. We're still deliberating about whether to remove the other two.

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -167,40 +193,37 @@ def _climate_factor(self) -> bool:
             )
```
lucasmbrown-usds, line 193: We still have duplicate code here – are you planning to remove that as part of this PR and just calculate the same values once?

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -167,40 +193,37 @@ def _climate_factor(self) -> bool:
...
     def _energy_factor(self) -> bool:
         # In Xth percentile or above for DOE’s energy cost burden score (Source: LEAD Score)
         # AND
         # Low income: In 60th percentile or above for percent of block group population
         # of households where household income is less than or equal to twice the federal
-        # poverty level. Source: Census's American Community Survey]
-        energy_criteria = (
+        # poverty level. Source: Census's American Community Survey]  
+
+        threshold_one = (
```
lucasmbrown-usds, line 205: Nit: can we use a more meaningful name here? Like `energy_burden_threshold`? Just helps to reduce future errors.

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -167,40 +193,37 @@ def _climate_factor(self) -> bool:
...
-            ]
-            >= self.LOW_INCOME_THRESHOLD
-        ) & energy_criteria
+        self.df[
+            field_names.ABOVE_90TH_FOR_COST_BURDEN_LOW_INCOME
+        ] = threshold_one & self.FPL_200_SERIES
+
+        self.df[field_names.PM25_LOW_INCOME] = threshold_two & self.FPL_200_SERIES
+
+        return (threshold_one | threshold_two) & self.FPL_200_SERIES
```
lucasmbrown-usds, line 226: Why not replace this with: 

```
return (field_names.ABOVE_90TH_FOR_COST_BURDEN_LOW_INCOME | self.df[field_names.PM25_LOW_INCOME])
``` 

That would reduce the duplicate logic of checking `threshold_one & self.FPL_200_SERIES`

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -212,25 +235,34 @@ def _transportation_factor(self) -> bool:
         # Low income: In 60th percentile or above for percent of block group population
         # of households where household income is less than or equal to twice the federal
         # poverty level. Source: Census's American Community Survey]
-        transportation_criteria = (
+
+        threshold_one = (
```
lucasmbrown-usds, line 239: Same comments from energy factor apply everywhere.

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -401,7 +433,7 @@ def _workforce_factor(self) -> bool:
                     field_names.UNEMPLOYMENT_FIELD
                     + field_names.PERCENTILE_FIELD_SUFFIX
                 ]
-                >= self.ENVIRONMENTAL_BURDEN_THRESHOLD
```
lucasmbrown-usds, line 404: Nit: merge conflicts. I think you want this to go back to `>=`, here and everywhere.

lucasmbrown-usds (changes requested):
Some progress! 

By the way, this isn't properly documented in the ticket which is unfortunate, but we're also gonna need a *count* of total criteria exceeded (e.g., an int from 0 to 16 or whatever the total number of criteria is). 

Up to you how you want to create that. I could see two options: 

1. create a field like df.THRESHOLD_COUNT that you add to with each criteria exceeded, like `df.THRESHOLD_COUNT[df[field_names.LIFE_EXPECTANCY_INCOME]] += 1` or something. 

2. at the end of score L, manually reference all the ~16 criteria fields and add them together

data/data-pipeline/data_pipeline/score/field_names.py
ahlusar1989: i see - I wasn't clear what the comment meant. I just thought you meant LOW_HS

ahlusar1989: I am not clear on what "this" is and what would like to be done. Do you mind clarifying for me?

ahlusar1989:
@lucasmbrown-usds Closing this PR because I need to update local and there are way too many merge conflicts. I might as well start from scratch using the freshest stuff.

lucasmbrown-usds:
> @lucasmbrown-usds Closing this PR because I need to update local and there are way too many merge conflicts. I might as well start from scratch using the freshest stuff.

Ok!

Issue #242, opened 2021-06-28
As a contributor to J40, I want to compare the J40 score results to Redlined neighborhoods from Univ of Richmond, so that I can understand how the J40 score compares.


BethMattern:
**Description**
The J40 team is working to develop the best possible score for identifying the most underserved and overburdened communities. In order to assess the quality of the score and build credibility in the EJ community, the J40 team build a tool to compare the output of the most recent version of the J40 score with other data sources that generate scores.

We need help adding more data sources for comparison to our tool. We would like to compare to the Redlined neighborhoods from Univ of Richmond, that can be found here: https://dsl.richmond.edu/panorama/redlining/#loc=5/39.1/-94.58&text=downloads

We will only want to compare the areas that are included in the Redlined neighborhoods data, which is likely urban areas. The research question is to see the J40 score for the redlined neighborhoods and to see if these neighborhoods get high J40 scores.

**Tasks**
- [ ] ETL redlined neighborhood data
- [ ] Modify comparison tool script to load redlined neighborhood data and compare it against current score
- [ ] Modify comparison tool script to add stats about the comparison with redlined neighborhoods with the current score to the report generated at the end of the script

**Definition of "Done"**
- [ ] The J40 comparison tool can be run using the most recent version of our score
- [ ] The comparison tool can access the redlined neighborhood data
- [ ] A report / data export can be generated that outputs the results of the comparison on a census tract level

PR #978, opened 2021-12-04
Issue 242: Add HOLC Grades to data inputs

lucasmbrown-usds:
Addresses #242 . Adds HOLC grades to data inputs as well as comparison tool. Not used in any scores/definitions.


ahlusar1989 (approved):
Very nice - and glad that this is being used!

PR #984, opened 2021-12-06
Creating notebook to compare two score files for differences

lucasmbrown-usds:
I created this because I wanted to check the results of the score file before and after some massive refactors we've been doing. So far the results look good!


esfoobar-usds (approved):
🍏

switzersc-usds (approved):
fab! so glad you made this! would love to incorporate this (or something similar) into a CI/CD check in the future.

Issue #953, opened 2021-12-02
When I open the csv or xlsx file, the prefixed 0s in census tract IDs are lost, causing usability issues.


switzersc-usds:
**Describe the bug**
When I open the data file (either) in Excel, the first 0 of many of the census tract IDs is dropped and therefore the ID is made incorrect / unusable.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

lucasmbrown-usds:
FYI, I'm not sure you'll be able to fix this in a CSV -- maybe writing strings around the cell like `"01234567890"` works? For excel, you may have to write `='01234567890`, if you try `"01234567890"` and it doesn't work there either.

PR #985, opened 2021-12-06
Update etl constants to use score field_names and put strings around tract IDs in downloadable CSV

### LINKED ISSUE #977: As a data developer, I want to update `etl/score/constants.py` to use `field_names`, so the code is easier to update.
**Description**
Right now, in the file `data/data-pipeline/data_pipeline/etl/score/constants.py`, we have a few constants like TILES_SCORE_COLUMNS, TILES_SCORE_FLOAT_COLUMNS, etc that list field names. We should replace these strings with references to `field_names` to make the code easier to update and be consistent.

switzersc-usds:
Closes #977 and #953.

This PR does two things:
1. Update etl constants to use score field_names and put strings around tract IDs in downloadable CSV
2. Put strings around tract IDs in downloadable CSV to preserve leading zeros when opening the CSV in a program like Excel.

There was no need to modify the xls file creation because the string type is preserved and interpreted correctly in Excel already.

One note is that this does cause the ID in the CSV to be have quotes around it, which might be annoying because those can be hard to remove. Maybe we don't want this behavior?


data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -4,6 +4,8 @@
 import pandas as pd
 from data_pipeline.config import settings
 
+import score.field_names as field_names
```
lucasmbrown-usds, line 7: nit: why not just from score import field_names?

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -59,88 +61,88 @@
...
-    "Traffic proximity and volume (percentile)",
-    "FEMA Risk Index Expected Annual Loss Score (percentile)",
-    "Energy burden (percentile)",
-    "Wastewater discharge (percentile)",
-    "Percent pre-1960s housing (lead paint indicator) (percentile)",
-    "Diesel particulate matter (percentile)",
-    "Particulate matter (PM2.5) (percentile)",
-    "Median household income (% of AMI) (percentile)",
-    "Percent of individuals < 200% Federal Poverty Line (percentile)",
+    field_names.GEOID_TRACT_FIELD,
```
lucasmbrown-usds, line 64: So sexy!

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -59,88 +61,88 @@
...
+    field_names.STATE_FIELD,
+    field_names.COUNTY_FIELD,
+    field_names.TOTAL_POP_FIELD,
+    field_names.SCORE_D + field_names.PERCENTILE_FIELD_SUFFIX,
```
lucasmbrown-usds, line 68: Should we get rid of all the Score D/E/G stuff from the tiles?

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -59,88 +61,88 @@
...
-    "Traffic proximity and volume",
-    "FEMA Risk Index Expected Annual Loss Score",
-    "Energy burden",
-    "Housing burden (percent)",
-    "Wastewater discharge",
-    "Percent pre-1960s housing (lead paint indicator)",
-    "Diesel particulate matter",
-    "Particulate matter (PM2.5)",
-    "Total population",
+    field_names.AMI_FIELD,
```
lucasmbrown-usds, line 130: :tada:

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -155,11 +157,11 @@
...
-    "GEOID10_TRACT",
-    "County Name",
-    "State Name",
-    "Score G (communities)",
-    "Median household income (% of AMI)",
-    "Median household income (% of state median household income) (percentile)",
+    field_names.GEOID_TRACT_FIELD,
+    field_names.COUNTY_FIELD,
+    field_names.STATE_FIELD,
+    field_names.SCORE_G_COMMUNITIES,
```
lucasmbrown-usds, line 163: should this be `SCORE_L_COMMUNITIES` now?

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -276,12 +276,13 @@ def _load_downloadable_zip(
...
 
-        logger.info("Writing downloadable csv")
-        downloadable_df_copy.to_csv(csv_path, index=False)
-
         logger.info("Writing downloadable excel")
         downloadable_df_copy.to_excel(excel_path, index=False)
 
+        logger.info("Writing downloadable csv")
+        downloadable_df_copy[self.GEOID_TRACT_FIELD_NAME] = "\"" + downloadable_df_copy[self.GEOID_TRACT_FIELD_NAME] + "\""
+        downloadable_df_copy.to_csv(csv_path, index=False)
```
lucasmbrown-usds, line 284: instead of manually adding the quotes, we could also add this to the `to_csv`: 

`downloadable_df_copy.to_csv(.... quoting=csv.QUOTE_NONNUMERIC)` See https://docs.python.org/3/library/csv.html 

But manually adding is also fine!

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -528,7 +528,7 @@ def _workforce_factor(self) -> bool:
 
         median_income_threshold = (
             self.df[
-                field_names.MEDIAN_INCOME_PERCENT_AMI_FIELD
+                field_names.MEDIAN_INCOME_AS_PERCENT_OF_AMI_FIELD
```
lucasmbrown-usds, line 531: nice catch

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -276,12 +276,13 @@ def _load_downloadable_zip(
             inplace=False,
         )
 
-        logger.info("Writing downloadable csv")
-        downloadable_df_copy.to_csv(csv_path, index=False)
-
         logger.info("Writing downloadable excel")
         downloadable_df_copy.to_excel(excel_path, index=False)
```
lucasmbrown-usds, line 280: That's cool that Excel preserves the string type!

lucasmbrown-usds (commented):
Overall looking good!

data/data-pipeline/data_pipeline/score/score_l.py
ahlusar1989: +1 ☝️

data/data-pipeline/data_pipeline/etl/score/constants.py
switzersc-usds: This was purely to change the file, not do the work of updating everything to have exactly what we want (since that is still TBD and captured in a separate issue).

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
switzersc-usds: Oh cool, I didn't know about that! I'll update it.

data/data-pipeline/data_pipeline/etl/score/constants.py
switzersc-usds: Ditto on above; happy to make changes but figured I'd wait til we have the final list of fields we want and where we want them.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
switzersc-usds: Actually couldn't get this to work in 15 minutes, so I'm going to keep as is (doesn't seem worth continuing to troubleshoot or doing a bigger refactor to how we invoke CSV write).

data/data-pipeline/data_pipeline/etl/score/constants.py
switzersc-usds: updated to match the import style elsewhere

data/data-pipeline/data_pipeline/etl/score/constants.py
lucasmbrown-usds: Ok got it!

data/data-pipeline/data_pipeline/etl/score/constants.py
lucasmbrown-usds: That makes a lot of sense!

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
lucasmbrown-usds: Good call!

lucasmbrown-usds (approved):
Looks great! So fresh and so clean now!!

PR #996, opened 2021-12-06
Remove "infrastucture" directory

switzersc-usds:
We originally used the approach in the infrastructure directory when we
thought we would be using Amazon Lambda for different parts of our
deployment pipeline. We have since then moved to using Github Actions
and no longer need this code, and keeping it in `main` has caused
confusion for onboarding new folks. This commit removes the directory
(although this will still be around in version control so we can always
view it or bring it back in the future if we want to).


sverchdotgov:
I would like to update https://github.com/usds/justice40-tool/blob/switzersc-usds/cleanup-infra/INSTALLATION.md#repo-organization to document the new usage of github actions, but I don't know how you keep it in sync with https://github.com/usds/justice40-tool/blob/switzersc-usds/cleanup-infra/INSTALLATION-es.md#ejecuta-el-c%C3%B3digo.

But yeah, that's the only place in the docs that links here (and the only place that documents the infrastructure, so it could be expanded to talk about github actions).

sverchdotgov (commented):
Just commenting that this link needs updating: https://github.com/usds/justice40-tool/pull/996#issuecomment-987209315. I'm definitely up for updating those docs and merging this PR, or doing that in another PR (don't know your PR etiquette yet).

esfoobar-usds (approved):
🍏

Issue #993, opened 2021-12-06
As a data developer, I want to load the tree equity data as a CSV, not as geojson, so that it's easier to integrate into our other analytics.


lucasmbrown-usds:
**Description**
We currently load tree equity data into the tool. This was started from a PR submitted by people from the tree equity team, which is awesome. The ETL currently saves the results as geojson though, so it doesn’t play nicely with any of our other analytics.

Instead of saving as geojson, this ticket is to save the results as a CSV, with one row per census tract. 

See also https://github.com/usds/justice40-tool/issues/276.

PR #997, opened 2021-12-07
Tree Equity Output: Change output from geojson to CSV format for easier analysis

ahlusar1989:
Addresses #993 

I presume this is the reference for this [dataset](https://www.treeequityscore.org/datasources/).

Steps to reproduce - `poetry run python3 data_pipeline/application.py etl-run --dataset tree_equity_score`

@lucasmbrown-usds @switzersc-usds One thing that I have not done yet is look into into the data. There are ~ 155, 029 GEOIDs in the data and I am interested in establishing relationships with other tract-level datasets - such as HOLC and property valuation. Is this worth investigating? I also notice that there are some fields in this dataset that we may already represent. The naming conventions in the columns are also not aligned. Do we reference field names here? 

Moreover, there is an entire column that represented geological coordinates and geometry. Do we need this?


data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -87,4 +87,4 @@ def load(self) -> None:
         logger.info("Saving Tree Equity Score GeoJSON")
         # write nationwide csv
         self.CSV_PATH.mkdir(parents=True, exist_ok=True)
-        self.df.to_file(self.CSV_PATH / "tes_conus.geojson", driver="GeoJSON")
+        self.df.to_csv(self.CSV_PATH / "tes_conus.csv", index=False)
```
switzersc-usds, line 90: For consistency with other ETL classes, I'd say let's change the whole file name to `usa.csv` instead of `tes_conus.csv`

switzersc-usds:
>  One thing that I have not done yet is look into into the data.

@ahlusar1989 @saran-ahluwalia -- I'd say the next step here is probably adding this data to the comparison tool so we can start doing some of that analysis and at the very least incorporate it into our automatic report generation in that tool. It's up to you whether you want to include that in this PR, but I'd say if you can't get to that until next week or after, then let's go ahead and finish this PR and we can make a separate issue for integrating this data into the comparison tool and answering some initial questions with the data.

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
saran-ahluwalia: Got it - thanks @switzersc-usds

saran-ahluwalia:
@switzersc-usds  in terms of space considerations, I just wanted to follow up from earlier in my comment: there is an entire column that represented geological coordinates and geometry. Do we need this?

switzersc-usds:
Oh! I see. If the geographic resolution is census tract or CBG (i.e. if each row in the CSV corresponds to a tract or CBG), then I don't think we need to keep geometries. However if the resolution is a different geographic unit then I'd say keep the geometry data so analysis/combination with other data sets is easier.

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -87,4 +87,42 @@ def load(self) -> None:
         logger.info("Saving Tree Equity Score GeoJSON")
         # write nationwide csv
         self.CSV_PATH.mkdir(parents=True, exist_ok=True)
-        self.df.to_file(self.CSV_PATH / "tes_conus.geojson", driver="GeoJSON")
+        # here I remove the geometry
+        self.df = self.df[
+            [
+                "geoid",
```
switzersc-usds, line 93: since we're looking at tracts in this dataset, can you go ahead and map this field to the tract field constant (`GEOID_TRACT_FIELD`) from the `field_names` module? this will help make it easier to join with other datasets later because the join column will be named the same.

saran-ahluwalia: A fantastic idea!

saran-ahluwalia: We no longer have a reference of `GEOID_TRACT_FIELD` in field names. However, what I do see consistently is `GEOID_TRACT_FIELD_NAME` as an instance variable of any subclass of `ExtractTransformLoad`. That is what is added now.

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -83,8 +83,52 @@ def transform(self) -> None:
             pd.concat(tes_state_dfs), crs=tes_state_dfs[0].crs
         )
 
+        # rename ID to Tract ID
+        self.df.rename(
+            columns={"geoid": self.GEOID_TRACT_FIELD_NAME},
```
switzersc-usds, line 88: nice!

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -83,8 +83,52 @@ def transform(self) -> None:
             pd.concat(tes_state_dfs), crs=tes_state_dfs[0].crs
         )
 
+        # rename ID to Tract ID
+        self.df.rename(
```
switzersc-usds, line 87: this might raise a warning about changing the dataframe in place (instead of copying it), but if you're not getting an error or warning and the rename is successful, then all good!

saran-ahluwalia: Indeed, I have not been receiving any errors or warnings from Pandas 👍

lucasmbrown-usds (changes requested):

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -83,8 +83,52 @@ def transform(self) -> None:
...
+
     def load(self) -> None:
         logger.info("Saving Tree Equity Score GeoJSON")
         # write nationwide csv
         self.CSV_PATH.mkdir(parents=True, exist_ok=True)
-        self.df.to_file(self.CSV_PATH / "tes_conus.geojson", driver="GeoJSON")
+        # here I remove the geometry
+        self.df = self.df[
+            [
+                self.GEOID_TRACT_FIELD_NAME,
```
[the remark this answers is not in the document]
saran-ahluwalia: Yes it appears to be the case: https://www.treeequityscore.org/datasources/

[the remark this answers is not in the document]
saran-ahluwalia: In that case this is not the appropriate field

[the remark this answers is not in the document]
lucasmbrown-usds: If you want to keep it at the CBG level for now, we can just make sure GEOID gets mapped to `ExtractTransformLoad.GEOID_FIELD_NAME` and it won't be as confusing. Or you could spend a little time figuring out how to aggregate this into tracts, but I'm not sure that is worth the effort right now.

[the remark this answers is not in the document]
saran-ahluwalia: Also, what do we consider to be the most relevant for scoring and analysis purposes?

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -83,8 +83,52 @@ def transform(self) -> None:
...
+
     def load(self) -> None:
         logger.info("Saving Tree Equity Score GeoJSON")
         # write nationwide csv
         self.CSV_PATH.mkdir(parents=True, exist_ok=True)
-        self.df.to_file(self.CSV_PATH / "tes_conus.geojson", driver="GeoJSON")
+        # here I remove the geometry
+        self.df = self.df[
+            [
+                ExtractTransformLoad.GEOID_FIELD_NAME,
```
lucasmbrown-usds, line 99: Because this whole ETL class is actually an extension of `ExtractTransformLoad`, you can just do `self.GEOID_FIELD_NAME`

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -83,8 +83,52 @@ def transform(self) -> None:
...
+            columns={"geoid": ExtractTransformLoad.GEOID_FIELD_NAME},
+            inplace=True,
+        )
+
     def load(self) -> None:
         logger.info("Saving Tree Equity Score GeoJSON")
         # write nationwide csv
         self.CSV_PATH.mkdir(parents=True, exist_ok=True)
-        self.df.to_file(self.CSV_PATH / "tes_conus.geojson", driver="GeoJSON")
+        # here I remove the geometry
```
lucasmbrown-usds, line 96: To be nitpicky, could we clean up these comments a bit to be more readable to an external audience encountering this codebase for the first time? 

Sometimes your comments read like things that I would typically put as comments on the pull request as explanations of the current work and changes to collaborators, but those comments are ephemeral, not designed to last or make sense.

There's no "I" meaningfully in a codebase written by many authors, so I would avoid using personal pronouns.

Please use comments to be really clear and unambiguous about what you mean. I don't actually know what you mean by "remove the geometry" – do you mean something like "drop the fields related to geojson boundaries"? I  don't know.

lucasmbrown-usds (changes requested):

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
saran-ahluwalia: Good point - I will just reference the `super` constructor that is called first..

saran-ahluwalia:
Closing this PR because the repository for @ahlusar1989 does not exist anymore. I will reopen this PR under @saran-ahluwalia.

PR #998, opened 2021-12-07
Shows N/A when data is not present on selected track

### LINKED ISSUE #992: Tract are failing for added territories that have no data
Add handling for no data cases to all tracts in UI and also come up with ways to show no data

@KameronKerger and @glennettec 

Maybe we should talk about what this should look like. Thanks

vim-usds:
closes #992

### DIFF
--- a/client/src/components/AreaDetail/index.tsx
+++ b/client/src/components/AreaDetail/index.tsx
@@ -12,24 +12,26 @@ import * as constants from '../../data/constants';
 import * as EXPLORE_COPY from '../../data/copy/explore';
 import * as METHODOLOGY_COPY from '../../data/copy/methodology';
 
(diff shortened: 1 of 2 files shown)


esfoobar-usds (changes requested):
I know this is not part of this work, but I still see "Census block group" on the side panel. Can we update to "Tract" or "Tract ID"?

/cc @KameronKerger @BethMattern @glennettec

glennettec:
@esfoobar-usds @BethMattern @KameronKerger instead of showing "N/A" can we show "--" because it implies that there's no data for the indicator. "N/A" implies that the indicator is not applicable.

esfoobar-usds (approved):
I see @KameronKerger addressed the Census Block side panel title on the panel redesign meeting, so I think we can approve.

Issue #1004, opened 2021-12-07
Update HUD housing cost burden data source


BethMattern:
Updated HUD data has been released to be as of 2014 - 2018. We need to import the new data from their portal.

PR #1005, opened 2021-12-07
Update HUD Housing Burden

ahlusar1989:
Addresses #1004 

To replicate: 

1. `poetry run python3 data_pipeline/application.py etl-run --dataset hud_housing`

2. Verified output from `score-run`


lucasmbrown-usds (approved):
Looks great, thanks for the quick work! I ran this locally and LGTM. 

One option comment. I've been going through and adding estimates of the size of the download to the logger statements so that the user using these CLI commands knows whether to expect to wait a long time or not.

For instance, `logger.info("Extracting HUD Housing Data")` could be changed to `logger.info("Downloading 190MB HUD Housing Data").

saran-ahluwalia:
@lucasmbrown-usds I will add that size information and then merge

ahlusar1989:
Merged most up-to-date HUD data

Issue #967, opened 2021-12-03
Investigate creating percentiles that split the comparisons between rural only and urban only


lucasmbrown-usds:
**Description**
There are some challenging data problems introduced by comparing rural areas and urban areas using the same percentiles. 

One suggestion to address this challenge is to separate the apples from the oranges, so to speak. 

This would mean: 

1. Decide a methodology for confirming whether a tract counts as urban or rural. Currently in the codebase, we use Geocorr to identify the % rural of a tract, and mark the tract as rural if the percentage is >50% and urban otherwise. This may or may not be the right methodology. 
2. Once tracts are marked as urban or rural, create one percentile rank that *only* ranks urban tracts, and one percentile rank that *only* ranks rural tracts. 
3. Combine into a single field.
4. Use this combined field to set criteria for Definition L, so that urban thresholds will be included if they exceed 90th percentile of urban tracts and rural thresholds will be included if they exceed 90th percentile of rural tracts. 

We would then want to run the comparison tool to see the impact that this has on overall methodology.

lucasmbrown-usds:
Call the combined field (percentile*).

PR #1006, opened 2021-12-07
Issue 967: Calculate urban/rural percentiles

lucasmbrown-usds:
Adds apples-to-apples percentiles, finishing issue #967 .


esfoobar-usds (approved):
🍏

Issue #949, opened 2021-12-02
Task: Review the level of effort and barriers to entry for a new contributor to onboard into the project


sverchdotgov:
## Describe the task

Someone new to the project (like me, @sverchdotgov) run through the end to end setup process of the project and document all the pain points that we can improve on.

## Acceptance Criteria

- [ ] Make any small updates as needed
- [ ] File new issues if there are things that are high in complexity or time

## Additional context

* TODO

sverchdotgov:
So far:

- The README structure is very clear, I could get straight to https://github.com/usds/justice40-tool/blob/main/INSTALLATION.md
- I can see the different parts of the stack clearly split up here (deployment, data pipeline, frontend) https://github.com/usds/justice40-tool/blob/main/INSTALLATION.md#repo-organization

One of the things I'd like to do is figure out how much of the stack needs to be spun up before you have something useful (i.e. how integrated is everything) so I'm going to start from deployment and move up through the data pipeline and then to the frontend. If that's backwards from what someone would normally do it, then maybe it'll yield something useful.

Only thing I'll note so far is that the instructions are Mac and Windows specific, with no mention of Linux, but maybe that's ok since a lot of what's documented is the basic machine setup (like installing git).

sverchdotgov:
In looking at https://github.com/usds/justice40-tool/blob/main/infrastructure/README.md, it seems like that documentation is more a set of notes that would be useful for maintainers of the infrastructure, rather than something as curated to an outside contributor as the other documentation was.

I don't think there's much here, maybe a simple intro. The site is deployed on AWS using the [Serverless](https://www.serverless.com/framework/docs/providers/aws/cli-reference/deploy) framework, and there are some Serverless functions in there, but I don't know what they do yet.

So far, I see:

- Add an overview, just saying what services we're using and for what.
- Document what the different environments are for, and how to add a new one.
- Document what the functions are for.

But maybe this is not the thing we expect people to care most about getting involved with. In that case, maybe a simple link to the other parts of the stack (like "if you are interested in X, go here").

sverchdotgov:
Moving onto the data pipeline here: https://github.com/usds/justice40-tool/blob/main/data/data-pipeline/README.md

Notes so far:

[...]

PR #1016, opened 2021-12-08
Update documentation to make it easier for users to find the right content for them

sverchdotgov:
This is a result of some previous discussions in https://github.com/usds/justice40-tool/issues/949#issuecomment-988091659. It's not really touching the idea of the wiki yet, since we will discuss that in a follow up conversation. I just wanted to get something up there that we can all react to.

Changes:

- Added a "Datasets" section to the top level readme, so we can add download links there for people who only want the data.
- Added a "Local Quickstart" section to the top level readme that links to some docker compose setup steps, for people who just want to try running the app locally.
- Added an "Advanced Guides" section for everyone else looking to go deeper.
- Updated INSTALLATION.md to only itself document installation of more generic things like git, docker, and homebrew, and delegate more specific installation guides to the sub components where it's relevant (e.g. don't document node setup there, because someone working with the data pipeline doesn't need that).
- Added a readme in the github actions directory just to give some context, and link that from INSTALLATION.md (it's not really something requiring an install, but I guess if you wanted to "install" our github actions deployment scripts in your own repo I guess you can stretch it. It seems like the most logical place for it for now).

I'd recommend going to the README on this branch and starting there to review this: https://github.com/usds/justice40-tool/blob/sverchdotgov/new-user-docs-updates/README.md


DATASETS.md
```diff
@@ -0,0 +1,8 @@
+# Justice40 Datasets
+
+This page contains web links to the datasets that are uploaded as part of our data pipeline, if you want to access them directly.
+
+For now, we have two datasets:
```
switzersc-usds, line 5: We also have all the datasets that we ETL -- once we ETL and do some normalization, they are available as CSVs in S3. I've linked to them previously in the data-pipeline README because I found I was sometimes looking for those URLs (and also sharing them). Thoughts on adding those links here?

switzersc-usds: Oh, and we also have the data as GeoJSON, which I think would be good to highlight.

INSTALLATION.md
```diff
@@ -1,26 +1,28 @@
-# Installing the Justice40 Tool
+# Justice40 Tool Installation
 
 *[¡Lea esto en español!](INSTALLATION-es.md)*
```
switzersc-usds, line 3: Are you up for taking a pass at updating the ES version of this file too? Otherwise maybe @esfoobar-usds might be able to?

QUICKSTART.md
```diff
@@ -0,0 +1,23 @@
+# Local Quickstart
```
switzersc-usds, line 1: Flag that it would be good to have this in Spanish too. If we can't fit that in here, let's make an issue for it.

README.md
```diff
@@ -44,9 +44,17 @@ Contributions are always welcome! We encourage contributions in the form of disc
...
-The installation is a typical gatsby install and the details can be found in [INSTALLATION.md](https://github.com/usds/justice40-tool/blob/main/INSTALLATION.md)
+The intermediate steps of the data pipeline and the final output that is consumed by the frontend are all public and can be accessed directly. See [DATASETS.md](DATASETS.md) for these direct download links.
+
+## Local Quickstart
+
+If you want to run the entire application locally, see [QUICKSTART.md](QUICKSTART.md).
+
+## Advanced Guides
+
+If you have software experience or more specific use cases, the [project wiki](https://github.com/usds/justice40-tool/wiki) and [INSTALLATION.md](INSTALLATION.md) have more in depth documentation of how to work with this project.
```
switzersc-usds, line 57: I thought you were against the wiki?!

switzersc-usds (commented):
Looking good so far! My biggest question is that you link to the wiki in one of the new lines, but I thought you were against the wiki?! If we're not gonna have the wiki, we don't need to link to it. Let's talk about this soon and decide.

DATASETS.md
switzersc-usds: Oh oh, and I'd love to have a dedicated section or file on how to use our tile server API. That can be a separate PR though because it would probably good to include a decent amount of how-to content

QUICKSTART.md
```diff
@@ -0,0 +1,23 @@
...
+
+> *Important*: To be able to run the entire application, you may need to increase the memory allocated for docker to at least 8096 MB. See [this post](https://stackoverflow.com/a/44533437) for more details.
+
+Use `docker-compose` to run the application:
+
+```sh
+$ docker-compose up
+```
+
+After it initializes, you should be able to open the application in your browser at [http://localhost:8000](http://localhost:8000).
```
esfoobar-usds, line 23: I would add a text here with an estimated time of completion -- i.e., that the user should expect about 2 hours before they can see the map

DATASETS.md
sverchdotgov: This would definitely be the place for all of that. The main goal here was to find the right structure.

I don't know much about the tile server api, but that could go here as well since it's focused on "how do I get access to the data directly?".

INSTALLATION.md
```diff
@@ -1,26 +1,28 @@
...
 
-## Required Software
-You'll need the following set up in your environment before continuing to run the Justice40 Tool code.
+This page documents the installation steps for some of the software needed to work with this project. After following the steps on this page, continue on to one of the following, depending on what you are trying to do:
 
-- git, node, and an IDE
+- If you want to quickly run everything locally, see [`QUICKSTART.md`](QUICKSTART.md).
```
switzersc-usds, line 7: Probably worth calling this out more and saying STOP HERE if you don't want to go beyond getting the tool up and running fast.

DATASETS.md
sverchdotgov: @switzersc-usds @esfoobar-usds I think if we think this file is a good idea, we should agree on an update frequency, and reflect that in a github action. What do you think of that?

For now, I'll put "whenever we run our pipeline, which we currently don't run on any specific cadence". But I think for people to find this useful, we might want to provide those guarantees (although that's out of scope of this PR I think).

sverchdotgov: Also, since it's in s3 and s3 is not a real filesystem, we can't just send people to a directory and say "here are all the datasets we have available". They just get access denied unless we link to specific objects. Just a note that this will make keeping this up to date a bit more active effort.

INSTALLATION.md
sverchdotgov: Unfortunately I don't think I'd be qualified to do that, my Spanish would definitely not look right to a native speaker.

QUICKSTART.md
sverchdotgov: How is this normally done @switzersc-usds?

README.md
sverchdotgov: Yup, removed this.

QUICKSTART.md
sverchdotgov: Hmm, it didn't take me two hours to run this.

INSTALLATION.md
sverchdotgov: Done.

QUICKSTART.md
switzersc-usds: We go through our typical State Department translation process or someone on the team with Spanish chops scrubs in. cc @esfoobar-usds 

(We can also make an issue for this if we don't have time to get this in before end of sprint, in the interest of keeping issues self-contained within the sprint)

QUICKSTART.md
switzersc-usds: Good idea

INSTALLATION.md
switzersc-usds: No worries at all!

DATASETS.md
switzersc-usds: Agreed on automating the updating of this file with a Github action - that should be pretty doable. I can make a new issue for that. 

Re: S3 links, yeah, good point. We should be able to automate the link generation in the aforementioned Github Action though, since they follow a standard pattern.

sverchdotgov: Ok, I'll let you resolve this comment when that issue exists. I like the idea of auto-generating the links somehow, and maybe making a pull request to update them automatically if they change.

QUICKSTART.md
sverchdotgov: I also like the idea of a follow up issue for this, since this seems like a different kind of work.

QUICKSTART.md
sverchdotgov: I'll just put "which may take a while, possibly hours, because it has to download and process all the data", since it seems like there's variability.

sverchdotgov:
@switzersc-usds @esfoobar-usds @saran-ahluwalia I think I addressed the main comments from this and our discussion. Seems like there are two follow up tickets, for the Spanish translations and the github action to update the dataset links.

Some notes that we didn't already talk about.

- Made a "For Developers and Data Scientists" section in the top level readme.
- The datasets section probably needs some work to make it comprehensible/useful, but that can be part of the follow up.
- The geojson got access denied so there may be some permissions to figure out.

Which datasets are useful might be a conversation on its own.

switzersc-usds (approved):
Awesome updates - thank you!

PR #1018, opened 2021-12-08
Issue 970: reverse percentiles for AMI and life expectancy

### LINKED ISSUE #970: As a site user, I want to have AMI and life expectancy percentiles updated so that a high percentile identifies disadvantage.
**Description**
High life expectancy and high area median incomes are both “good” things, so for those two, unlike every other threshold, instead of using >= 90th percentile we have been using the <= 10th percentile. 

However, this is likely to be confusing to any users of the site or data download because it's inconsistent with other measures. 

For high school education, we do "low high school education" rates rather than "high school education rate", for instance. We should do the same thing here. 

Steps: 
1. Take existing percentile for AMI and life expectancy, and create new, reversed percentiles that are 1 minus those values. 
2. Update score logic to use new field and >= 90th rather than <= 10th. 
3. Update fields in CSV and Excel download packet.

lucasmbrown-usds:
Take existing percentile for AMI and life expectancy, and create new, reversed percentiles that are 1 minus those values.
Update score logic to use new field and >= 90th rather than <= 10th.

Closes #970


switzersc-usds (approved):
lgtm!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -505,10 +514,6 @@ def _prepare_initial_df(self) -> pd.DataFrame:
 
             max_value = df_copy[numeric_column].max(skipna=True)
 
-            logger.info(
```
lucasmbrown-usds, line 508: We don't use min/max anywhere so this logging felt too chatty and unnecessary.

esfoobar-usds (approved):
🍏

PR #1027, opened 2021-12-09
Place spanish content on feature ?flags=sp

### LINKED ISSUE #1017: Task: Put world icon, "English", and "Español" toggles behind a feature flag, make English version default for launch
## Describe the task
Because timing for incorporating Spanish, in particular the updates for Methodology & data, and side panel for indicators being final today, would introduce a risk of having only partial Spanish translation on the site, the team decided it would be prudent to hide the English/Spanish toggle area (the world icon + English link + Spanish link) in the upper right corner behind a feature flag. It's ok to leave the links as is, i.e. keep .../es/contact in addition to .../en/contact.

If hiding behind a feature flag isn't easily or quickly doable, it's ok to just remove the English/Spanish toggle area. We can always test the translations by replacing the /en with /es in the URLs.

vim-usds:
closes #1017 

To enable -  ?flags=sp

### DIFF
--- a/client/src/components/GovernmentBanner/GovernmentBanner.tsx
+++ b/client/src/components/GovernmentBanner/GovernmentBanner.tsx
@@ -2,17 +2,32 @@ import React from 'react';
 import {GovBanner} from '@trussworks/react-uswds';
 import {useIntl} from 'gatsby-plugin-intl';
 
+// Contexts:
+import {useFlags} from '../../contexts/FlagContext';
+
 import Language from '../Language';
 
 import * as styles from './GovernmentBanner.module.scss';
 
 const GovernmentBanner = () => {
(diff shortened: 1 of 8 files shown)


switzersc-usds:
I'm not able to QA this properly. When I add the featrure flag, the flag gets removed from the URL if it's the map page, or does not switch the content to spanish on the other pages. Not sure if I'm using the query string properly or what?

esfoobar-usds:
Works for me. You need to put the flag before the hash:
http://usds-geoplatform-justice40-website.s3-website-us-east-1.amazonaws.com/justice40-tool/1027-8ca087/en/cejst/?flags=sp#3/32.47/-86.5

esfoobar-usds (approved):
🍏

PR #1046, opened 2021-12-13
Add low HS education fields to tiles and download

lucasmbrown-usds:


data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -523,7 +523,7 @@ def _workforce_factor(self) -> bool:
             field_names.LOW_MEDIAN_INCOME_LOW_HS_EDUCATION_FIELD,
         ]
 
-        high_scool_achievement_rate_threshold = (
```
saran-ahluwalia, line 526: Perhaps too particular, but doesn't this break from the state-specific variable assignment for the other factors?

data/data-pipeline/data_pipeline/score/score_l.py
```diff
@@ -632,7 +633,7 @@ def _workforce_factor(self) -> bool:
             >= self.ENVIRONMENTAL_BURDEN_THRESHOLD
         )
 
-        island_areas_high_scool_achievement_rate_threshold = (
```
saran-ahluwalia, line 635: Same question as above ^

data/data-pipeline/data_pipeline/score/score_l.py
lucasmbrown-usds: I'm not sure I understand the question -- could you say a little more? Thanks!

saran-ahluwalia: Of course. Like we did here https://github.com/usds/justice40-tool/blob/a06555a4f01cd0eea8b80d7988288b6d6dd9a3bf/data/data-pipeline/data_pipeline/score/score_l.py#L178, or for other criterion, I was curious why we were using in-place assignment of the column here. This seems to be a departure from convention, no?

saran-ahluwalia: Got it - thanks for clarifying.

switzersc-usds (approved):

Issue #1049, opened 2021-12-13
As a data developer, I want to stop swallowing Census API errors, so that I can more easily tell if there's a problem with the census API.


lucasmbrown-usds:

vim-usds:
merged and closing

PR #1051, opened 2021-12-13
Issue 1049: Stop swallowing Census API errors

lucasmbrown-usds:
#1049. I noticed we were swallowing errors that happen on every request from the island territories. But this meant that new, unexpected errors were also getting swallowed, and the rest of the pipeline would run unaware that the source data was corrupted. 

This skips the island areas where we know Census ACS will never give us data, and raises all other errors.

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/sources/census_acs/etl_utils.py
+++ b/data/data-pipeline/data_pipeline/etl/sources/census_acs/etl_utils.py
@@ -8,6 +8,8 @@ from data_pipeline.utils import get_module_logger
 
 logger = get_module_logger(__name__)
 
+CENSUS_ACS_FIPS_CODES_TO_SKIP = ["60", "66", "69", "78"]
+
 
 def _fips_from_censusdata_censusgeo(censusgeo: censusdata.censusgeo) -> str:
     """Create a FIPS code from the proprietary censusgeo index."""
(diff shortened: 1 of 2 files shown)


data/data-pipeline/data_pipeline/etl/sources/census_acs_2010/etl.py
```diff
@@ -50,7 +50,6 @@ def __init__(self):
         ]
 
         self.EMPLOYMENT_LESS_THAN_HS_IN_LABOR_FORCE = (
-            # TODO: FIX!!!!!!
             "B23006_005E"
```
lucasmbrown-usds, line 53: I wrote this scary todo a while ago; it should never have been committed. This used to be the wrong field. Now it's the right field, it's already fixed.

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl_utils.py
```diff
@@ -22,32 +24,34 @@ def retrieve_census_acs_data(
...
+                    src=acs_type,
+                    year=acs_year,
+                    geo=censusdata.censusgeo(
+                        [("state", fips), ("county", "*"), ("tract", "*")]
+                    ),
+                    var=variables,
+                )
+                dfs.append(response)
+
+            except ValueError as e:
```
saran-ahluwalia, line 51: @lucasmbrown-usds I was thinking about the discrepancy in the number of census tracts. Do you think we should capture the number of unique tracts that are accessible (for accounting purposes)?

switzersc-usds (approved):

Issue #947, opened 2021-12-02
Task: Ensure there is a documented rollback plan for data pipeline issues


sverchdotgov:
## Describe the task

If the data pipeline breaks because of a new change, we should have confidence that we can rollback to a known good state.

This has some potential complications if upstream data sources are down or fail, so it's related to https://github.com/usds/justice40-tool/issues/946.

## Acceptance Criteria

- [ ] Document with rollback plan
- [ ] New backlog issues for known issues with rollback that we decided are not worth doing now

## Additional context

* TODO

sverchdotgov:
This is dependent on what we decide to do as a result of this: https://github.com/usds/justice40-tool/issues/946#issuecomment-989155252

Issue #946, opened 2021-12-02
[SPIKE] Investigate impact of upstream data sources failing


sverchdotgov:
## Describe the task

The pipeline pulls from live upstream endpoints, this would be an investigation into what would happen if they went down.

## Acceptance Criteria

- [ ] New or updated issues for addressing this problem
- [ ] Small document of the external dependencies that could fail and how they are mitigated

## Additional context

* TODO

sverchdotgov:
# Impact Analysis: Upstream Data Source Failures
## Pipeline Stages

This is what I think the steps are in the pipeline (with links to the github actions that run them):

- [Pull Census Data, Upload to S3](https://github.com/usds/justice40-tool/blob/main/.github/workflows/generate-census.yml#L41)
- [Run All ETLs and Scoring, Upload Combined Results to S3](https://github.com/usds/justice40-tool/blob/main/.github/workflows/generate-score.yml#L42)
- [Pull Results Of Previous Jobs From S3, Generate Tileset and GEOJson and Upload Both To S3](https://github.com/usds/justice40-tool/blob/main/.github/workflows/combine-tilefy.yml#L58)

## Failure Scenarios

### Complete Upstream Outage

#### Pipeline Behavior

I believe if there is a complete upstream outage in a data source, the ETL job will fail, causing the github action to stop and not upload anything to s3.

#### Impact

Inability to update the data, but the version of the data as of the last run will still be present. However, even if we want to update the data for an unrelated upstream source, I believe one endpoint having an outage will cause the entire github action to fail.

#### Possibile Mitigation

Change the ETL pipeline to upload the scoring for each individual source to S3 as a first stage, and add an option for skipping sources that have errors rather than failing the whole job.

### Upstream Data Corruption/Inaccuracy

#### Pipeline Behavior

If there is corruption in the data upstream, the pipeline has no way of knowing this, and will clobber the existing datasets.

#### Impact

The data will be wrong, and running the pipeline will not restore it.

#### Possible Mitigations

Since [s3 bucket versioning](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html) is enabled, in theory we can restore a previous version of the data. However, this is [not simple to do for an entire sub-path](https://serverfault.com/questions/754410/is-there-any-way-to-restore-an-entire-previous-version-of-a-folder-in-an-s3-buck/757566) so we would have to write some custom scripts.

[...]

PR #1069, opened 2021-12-16
Add worst case scenario rollback plan

sverchdotgov:
Fixes: https://github.com/usds/justice40-tool/issues/947

This is ensuring we at least have some way to rollback in the worst case scenario of the data being corrupted somehow, and our upstream data sources being down in a way that prevents us from regenerating any datasets. See https://github.com/usds/justice40-tool/issues/946#issuecomment-989155252 for an analysis of the different ways the pipeline can fail and their impacts.

This guide is very brittle to pipeline changes, and should be considered a temporary solution for the worst case. I think that someday the github actions should run more frequently and write to different paths, perhaps with a timestamp. That would make rolling back more straightforward, as a previous version of the data would already exist at some path in s3.


vim-usds:
@sverchdotgov - this is awesome!

I'm not sure about CLI access to AWS directly from our terminal. However I know that CLI access from GitHub actions work. See [here](https://github.com/usds/justice40-tool/blob/0d57dd572be027a2fc8b1625958ed68c4b900653/.github/workflows/deploy_staging.yml#L79).

I wonder if having these commands that you outlined could be placed into GitHub Action that can be triggered by the GitHub action UI, like [generate-census](https://github.com/usds/justice40-tool/blob/0d57dd572be027a2fc8b1625958ed68c4b900653/.github/workflows/generate-census.yml#L3) does. This may anyone to revert. 

Is this even possible, as some commands require python? Just my  $ .02!

switzersc-usds (approved):
🎉

vim-usds (approved):
LGTM!

sverchdotgov:
Merging since we reviewed this. If we add rollbacks as github actions in the future this may become obsolete, but it works for now.

Issue #579, opened 2021-08-27
Task: Cypress tests for AK and HI hang


vim-usds:
## Describe the task
There are 4 buttons on the map that when clicked will zoom into the map. The `map.spec.js` test should test that the map pans to the appropriate state. The `tests` object (below) contains the 4 territory buttons.

```
  const tests = {
    'Lower 48': '3.19/38.07/-95.87',
    // Todo: Investigate why Alaska and Hawaii cause the test to hang
    // 'Alaska': '3/63.28/-162.39',
    // 'Hawaii': '5.89/20.574/-161.438',
    'Puerto Rico': '7.65/18.2/-66.583',
  };
```

When adding Alaska or Hawaii back into the test object, the cypress test hangs intermittently. This test was run 20 consecutive times:

![Screen Shot 2021-08-27 at 3 18 02 PM](https://user-images.githubusercontent.com/86254807/131194159-83b6fdca-5119-4d72-888c-d015c07885bf.png)

Some questions worth exploring:
1. what causes this Alaska to intermittently work 40% of the time?
2. The way Alaska hangs 40% of the time, is the way HI hangs 100% percent of the time.

Debugging shows that the test loop runs for cy.* command and the tests hangs after the loop is completed.

```
    it(`Can focus on ${territory} `, () => {
      cy.getMap().then((map) => {
        cy.get(`[aria-label="Focus on ${territory}"]`).click();
        cy.waitForMapIdle(map);
        cy.url().should('include', zxy);
      });
    });
```

When the map is hung, moving the map slightly causes the test to un-hang and then fails on zxy mismatch, since the map was moved. This implies that it's possible that the `once` map event that `cy.waitForMapIdle()` is using may be getting missed by the listener.

Splitting up the test into 4 different `it(...)`s rather than the current looping over the test object was not effective. Forcing synchronicity (via async / await) also did not work.

PR #1082, opened 2021-12-21
Fix territory shortcuts when census tract is selected

### LINKED ISSUE #1056: Shortcuts broken after tract selection
**Describe the bug**
When I zoom in anywhere and select a tract to view, and THEN click the "AS" shortcut, the map moves to the left slightly instead of taking me to American Samoa.

**To Reproduce**
1. Go to the Explore the Tool page.
2. Zoom into any census tract NOT near American Samoa. 
3. Click the tract.
4. Click the AS shortcut
5. See that the map only moves slightly to the left.

**Expected behavior**
I expected the map to immediately jump to and zoom in on American Samoa. If you click the AS shortcut multiple times, it finally will move you to American Samoa, but it should only need one click.

**Screenshots**

 - OS: [e.g. iOS] MacOS Mojave
 - Browser [e.g. chrome, safari]- Firefox
 - Version [e.g. 22] - 89

**Additional context**

vim-usds:
# Problem
It seems that when a tract is selected and/or the zoom level was above 8 (z>8), when clicking on a _territory shortcut_, e.g., `AS`, it would move the map over a tiny bit. This is because the click was registering as a click on the _map_ and NOT the territory shortcut `AS`.

### Solution (TL;DR)
We had two click handlers (one for _territory shortcuts_ and one for the _map_) firing each time the map was clicked. This meant either handler could handle the event. Normally `preventDefault()` or `stopPropagation()` are used to solve this, however for some reason it was not working in this case. 

Now we have a single click handler (aka [Event Delegation Pattern](https://dmitripavlutin.com/javascript-event-delegation/)) to handle both the _territory shortcut_ clicks and _map_ clicks. Now we decide based on the event's payload if it's a _map_ event or a _territory shortcut_ event.  

# Purpose of PR

- Re-architects the territory controls which was done by Nat. I believe it's important that we exhaustively test this. While I was hoping this would have fixed the automated tests for this [existing issue](https://github.com/usds/justice40-tool/issues/579), unfortunately it doesn't and so we will need manual testing for now. 
- Centers `AS` more on the map
- Center `HI` more on the map
- the initial map position(z/lat/long) of the US is when we click on the `48`

## How to test
Click on the territories:
1. with various zoom levels, 
2. when various types of tracts (urban and rural) are selected or not selected
3. after using search
4. after selecting a tract after searching
5. after using the zoom controls
6. anything else you can think of!

Thx!
closes #1056

### DIFF
--- a/client/src/components/territoryFocusControl.tsx
+++ b/client/src/components/territoryFocusControl.tsx
@@ -1,53 +1,17 @@
 import {useIntl} from 'gatsby-plugin-intl';
 import React from 'react';
-import {LngLatBoundsLike} from 'mapbox-gl';
+import {MapEvent} from 'react-map-gl';
 
 import * as styles from './territoryFocusControl.module.scss';
 import * as EXPLORE_COPY from '../data/copy/explore';
-import * as constants from '../data/constants';
 
 interface ITerritoryFocusControl {
-  goToPlace(bounds: LngLatBoundsLike): void;
+  onClick(event: MapEvent | React.MouseEvent<HTMLButtonElement>):void;
 }
 
 
-const TerritoryFocusControl = ({goToPlace}: ITerritoryFocusControl) => {
+const TerritoryFocusControl = ({onClick}: ITerritoryFocusControl) => {
   const intl = useIntl();
-
-  const onClickTerritoryFocusButton = (event: React.MouseEvent<HTMLButtonElement>) => {
-    event.stopPropagation();
-    const buttonID = event.target && (event.target as HTMLElement).id;
-
-    switch (buttonID) {
-      case '48':
-        goToPlace(constants.LOWER_48_BOUNDS);
-        break;
-      case 'AK':
-        goToPlace(constants.ALASKA_BOUNDS);
-        break;
-      case 'HI':
-        goToPlace(constants.HAWAII_BOUNDS);
-        break;
-      case 'PR':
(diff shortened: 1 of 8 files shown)


esfoobar-usds (approved):
I can confirm this fixes the issue.

🍏

client/src/components/territoryFocusControl.tsx
```diff
@@ -1,54 +1,18 @@
...
-import {LngLatBoundsLike} from 'mapbox-gl';
+import {MapEvent} from 'react-map-gl';
 
 import * as styles from './territoryFocusControl.module.scss';
 import * as EXPLORE_COPY from '../data/copy/explore';
-import * as constants from '../data/constants';
 
 interface ITerritoryFocusControl {
-  goToPlace(bounds: LngLatBoundsLike): void;
+  onClick(event: MapEvent | React.MouseEvent<HTMLButtonElement>):void;
```
saran-ahluwalia, line 9: Slick 👍

vim-usds:
> I can confirm this fixes the issue.
> 
> 🍏

Thanks @esfoobar-usds !!

client/src/components/territoryFocusControl.tsx
vim-usds: Thx! This is still kinda head scratcher as to why the IDE complains about the `feature` property not being in the event payload. I have to look into this. Not a show stopper but more of a personal preference to not have any linting warnings.

Issue #1075, opened 2021-12-17
As a data developer, I want to develop a great testing framework for one of our ETL classes, so I have confidence I haven't introduced bugs into the code.


lucasmbrown-usds:
**Description**
Right now, only one of our ETL classes (national_risk_index) has tests on it. 

We should develop a single preferred approach to ETL testing (the national_risk_index seems like a really good start), and then implement tests for all of the current ETL classes.

See also discussion at https://github.com/usds/justice40-tool/issues/1048 about updating pickles and CSVs.

lucasmbrown-usds:
Objective of this work: 
* Have a great testing framework for ETL we feel will catch most errors
* Apply that framework to all classes

Punchlist of work to do: 
- [ ] Start with NRI tests, review to see if we want to make any updates 
- [ ] Create one new test based on that for another ETL class
- [ ] Open pull request for review
- [ ] When we feel pretty settled, create tests for every ETL class

lucasmbrown-usds:
1/27 brainstorm with @emma-nechamkin and @widal001 :

* Conditional skip on tests in pytest - pass a flag when running pytest or set an environment variable to force that test to run. 
* Should it be in the testing code or not? Yes, should be near tests
* Don't just rely on data snapshot tests -- explicitly write unit tests that check the logic of the code directly.
* Add more validation steps that answer questions like: are all the tracts we expect to see represented?

Issue #1048, opened 2021-12-13
As a data developer, I want to easily update the data pickles


lucasmbrown-usds:
**Description**
If we added a simple global constant like `UPDATE_PICKLES = False` to the `data/data-pipeline/data_pipeline/etl/score/tests/test_score_post.py` file, then when you need to update the pickles, you just set that parameter to True. 

And within each test, it would be something like: 

```
if UPDATE_PICKLES:
  data_path = Path.cwd()
  score_transformed_actual.to_pickle(data_path / "data_pipeline" / "etl" / "score" / "tests" / "snapshots" / "score_transformed_expected.pkl", protocol=4)
else:
   pdt.assert_frame_equal(
        score_transformed_actual, score_transformed_expected, check_dtype=False
    )
```

lucasmbrown-usds:
Potentially copy or expand upon framework developed by https://github.com/usds/justice40-tool/pull/1249

Issue #1013, opened 2021-12-08
As a data developer, I want to improve the loading of data for the `etl_score` file.


lucasmbrown-usds:
**Description**
Right now we load each ETL's results using something like:

```
        # EJSCreen csv Load
        ejscreen_csv = (
            constants.DATA_PATH / "dataset" / "ejscreen_2019" / "usa.csv"
        )
        self.ejscreen_df = pd.read_csv(
            ejscreen_csv,
            dtype={self.GEOID_TRACT_FIELD_NAME: "string"},
            low_memory=False,
        )
```

This requires knowing the CSV path (redundant to where it's defined in the class) and which fields need to be loaded with a certain dtype. 

Replace this with a method on every ETL class like `get_df` that checks whether the CSV exists,if it doesn't then ETLs it, and if it does then loads and returns it as a dataframe. 

We could also potentially add a boolean to the ETL base class `LOAD_FOR_SCORE_PROCESS` that's true or false, and `NUMERIC_COLUMNS_FOR_SCORE_PROCESS` etc. So that all info about ETL is kept in the ETL process.

PR #619, opened 2021-09-07
Add ETL Contract Checks

widal001:
## Summary

Creates a series of methods on the `ExtractTransformLoad` base class that simplify and validate adherence to a standardized ETL contract for dataset sources. Fixes #617 

## Changes Proposed

- Changes to `data_pipeline/etl/base.py`
  - Adds class GLOBAL `CENSUS_CSV` that can be used to read in the definitive list of Census Block Group FIPS codes from the census data
  - Creates an `__init__()` method to instantiate class attributes that can be used by each source ETL class and can be checked for compliance against the contract
    -  `self.NAME` Used to create output path and populate logger info
    -  `self.SOURCE_URL` Used to extract source data in self.extract()
    -  `self.OUTPUT_PATH` Location where output csv is written in self.load()
    -  `self.SCORE_COLS` List of columns that will be used in the score and should appear in the output csv
    -  `self.FIPS_CODES` Dataframe of Census Block Group FIPS codes used to standardize the columns in the output csv
  - Create a private method `self._get_census_fips_codes()` which extracts the Census Block Group and Census Tract FIPS codes from the Census data and stores it in `self.FIPS_CODES`
  - Implements `self._get_yaml_config()` which parses a yaml file for the class and uses it to set class attributes in `__init__()`
  - Create an `self.validate_output()` method that can be optionally called to check the compliance of the output of ETL classes with the conditions below:
    - An output csv has been created in the `data/dataset/{data_source_name}` directory
    - The output csv has a column named `GEOID10` which stores the census block group FIPS code
    - The columns in the output csv include the fields that will be used in the score and the names of those columns match the value expected by the score.
    - The `GEOID10` column contains all of the Census block groups
- Creates `tests/base/` directory to test the new methods added to `etl/base.py`

## Blockers or Questions

- [x] Several of these methods need to distinguish between the census ETL and the other dataset ETL classes, currently I've implemented this check by adding a key `is_dataset` to the `config.yaml` file. Is there a more elegant way to implement this check across class methods?
- [x] Are there other configuration variables we want to add to the `config.yaml` file for each dataset ETL? Or other class attributes that should be shared across dataset ETL classes?

## Additional Context 

A few of the features implemented in this PR are also designed to simplify the process of creating a "wizard" for new ETL classes, as described in #576, namely the pairing of `self._get_yaml_config()` with `self.__init__()` to dynamically populate a common set of class attributes that can be referenced by the other methods in the dataset classes.

[...]

PR #1088, opened 2021-12-27
Issue 1075: Add ETL tests to all data classes

lucasmbrown-usds:
#1075 . Work in progress by @saran-ahluwalia and myself to add tests to all ETL classes. 

Out of the punchlist [here](https://github.com/usds/justice40-tool/issues/1075#issuecomment-1001700157), so far we only have finished updating NRI. 

@esfoobar-usds @widal001  - let's talk about the overall approach. If this looks good to you, we can add these tests to every other ETL class. Let's get an approach we're happy with in one place before duplicating it elsewhere, though.


data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -107,7 +107,10 @@ def extract(
         # this can be accessed via super().extract()
         if source_url and extract_path:
             unzip_file_from_url(
-                source_url, self.TMP_PATH, extract_path, verify=verify
+                file_url=source_url,
```
lucasmbrown-usds, line 110: Just adding some named arguments because I love them.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/acs.csv
```diff
@@ -1,11 +0,0 @@
-GEOID10,POPULATION
```
lucasmbrown-usds, line 1: This file is no longer needed because we no longer convert from tracts to CBGs.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/input.csv
```diff
@@ -1,11 +1,11 @@
...
...,0.000003274959403,24.889691462507123,0.020362504191456,559.520769230545852,5.870692520235838,Very Low,6.746910814468500,Very Low,7.000000000000000,0.031969309000000,198555247.532617419958115,978.467889623433734,7436355961.138096809387207,7634911208.670714378356934,0.001559313957881,0.000003873415999,Very Low,9898.016764864929428,0.000121164102392,920.847178175504837,10818.863943040432787,23.658087226460832,Relatively High,20.211588413559333,Relatively Moderate,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000541106984497,0.000003737122925,No Rating,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,No Expected Annual Losses,0.000000000000000,No Rating,,,,,,,,,,,Insufficient Data,,,,,,,Insufficient Data,,Insufficient Data,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000085057095,0.000000105693624,0.000000000001782,No Rating,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,0.000000000000000,No Expected Annual Losses,0.000000000000000,No Rating,November 2021
```
lucasmbrown-usds, line 2: Changed two rows of input data as an example.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/transform.csv
```diff
@@ -1,56 +1,56 @@
...
....6070545485,Relatively Low,0.0000000000,0.0000653310,912658000.0000000000,8385.0000000000,63726000000.0000000000,147860.5647200878,64638805860.5647201538,0.0089662390,0.0000059784,0.0021079463,Very Low,534.6107152638,0.0000032750,24.8896914625,0.0203625042,559.5207692305,5.8706925202,Very Low,6.7469108145,Very Low,7.0000000000,0.0319693090,198555247.5326173902,978.4678896234,7436355961.1380958557,7634911208.6707134247,0.0015593140,0.0000038734,Very Low,9898.0167648649,0.0001211641,920.8471781755,10818.8639430404,23.6580872265,Relatively High,20.2115884136,Relatively Moderate,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0005411070,0.0000037371,No Rating,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000000,No Expected Annual Losses,0.0000000000,No Rating,,,,,,,,,,,Insufficient Data,,,,,,,Insufficient Data,,Insufficient Data,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000851,0.0000001057,0.0000000000,No Rating,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000000,0.0000000000,No Expected Annual Losses,0.0000000000,No Rating,November 2021,0.0000035067,0.0087782352,0.0000719506
```
lucasmbrown-usds: This is all changed because we stopped forcing a certain number of decimals in the output. All the numbers have stayed the same where we expected them to, though.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -10,8 +14,41 @@
     settings.APP_ROOT / "tests" / "sources" / "national_risk_index" / "data"
 )
 
+"""
+Set UPDATE_TEST_FIXTURES to `True` temporarily to allow you to quickly update the test
```
lucasmbrown-usds, line 18: We'll put this in a README somewhere too.

widal001:
@lucasmbrown-usds @saran-ahluwalia Thanks for pushing this work forward! I'll aim to take a look at this tonight and comment on this thread

widal001 (commented):
### Summary

@lucasmbrown-usds @saran-ahluwalia These look like solid improvements and the existing unit tests run successfully with the changes, so they should be good to merge. I added specific suggestions in inline comments and offered some broader considerations for next steps for this strategy below, but they shouldn't be blockers for this PR.

### Highlights

- Love the move to standardize unit tests across the ETL classes and these changes are a solid foundation to build on
- Automating the method to update the test fixtures for each ETL source is a smart way to save time and effort, if approached with the appropriate caution when populating those files
- Appreciate the introduction of a test for the `extract()` method that was an oversight in the initial testing suite that I wrote

### Considerations

- While I like the approach of automating the generation/updating of test fixtures, we'll just need to be careful we don't codify any bugs in the fixtures, since it seems like the ETL code being tested against those files is the same code used to generate them (at least in the transform step) 
- This is a bit outside the scope of the changes introduced by this PR, but something I flagged when submitting the first test suite for the NRI class is that the two different `conftest.py` (the one in `data_pipeline/tests/` and the one in `data_pipeline/ETL/tests/`) actually write tmp files to different directories. As a result, depending on which subset of the tests we're running, the location of the temp files created by tests can change. This doesn't result in any failures but can make it confusing if you're trying to manually inspect the output after executing the tests. I can explain why this is happening if it would be helpful, though it might be easier to explain in real-time instead of asynchronously. 
- It may be worth validating that the output test fixture adheres to the contract established for the output of each ETL source class. That was the initial goal behind the work done in #619 that I regrettably didn't get a chance to expand on. The `ExtractTransformLoad.validate_output()` method will likely need to be adjusted if we no longer use CBG as the common geographic unit for source data, but an updated version could be useful to ensure that the fixtures being used for each testing suite adhere to a standard contract that the score classes expect.

### Suggestions

- If the goal is to replicate a variation of this testing suite across each of the ETL source classes, it may be beneficial to create a `TestBaseETL` class with common code for each UUT in the ETL source class. That base test class can then be inherited by the respective test classes for each ETL source class per [this issue](https://stackoverflow.com/questions/52044490/replacement-for-test-case-inheritance-in-pytest)
- Abstracting a base class for ETL source tests might also be easier if we were to update the ETL source classes to utilize some of the standardized class attributes also introduced in #619, for example using the following items set in `ExtractTransformLoad.__init__()`. Admittedly, that probably isn't a requirement for getting unit testing in place and may be outside the scope of this immediate goal, even if it could help simplify how these tests are written.
```python
def __init__(self, config_path: Path) -> None:
        """Inits the class with instance specific variables"""

        # set by _get_yaml_config()
        self.NAME: str = None
        self.SOURCE_URL: str = None
        self.GEOID_COL: str = None
        self.GEO_LEVEL: str = None
        self.SCORE_COLS: list = None
        self.FIPS_CODES: pd.DataFrame = None
        self.OUTPUT_PATH: Path = None
        self.CENSUS_CSV: Path = None

        self._get_yaml_config(config_path)
```

lucasmbrown-usds:
In reponse to your excellent and thoughtful review, @widal001 -- 

> This is a bit outside the scope of the changes introduced by this PR, but something I flagged when submitting the first test suite for the NRI class is that the two different conftest.py (the one in data_pipeline/tests/ and the one in data_pipeline/ETL/tests/) actually write tmp files to different directories

Great catch. I created https://github.com/usds/justice40-tool/issues/1092 to track that. 

> It may be worth validating that the output test fixture adheres to the contract established for the output of each ETL source class.

I actually had never noticed `validate_output` on the base class, that's super cool. Question: do you think it's smarter to run that as part of the test, or as part of each ETL (e.g., make the ETL runner call `extract`, `transform`, `load`, and then `validate_output`)? Or maybe in both places! 

> TestBaseETL

Yeaaaa I really like that idea. That would clean this up a lot and it would also be nice to have explicit test contracts on the parent class that make it obvious that all child classes must pass these same tests, in addition to the child-specific tests. 

I think sadly the `_get_yaml_config` is a relic from a different vision at a different time -- we've moved to specifying all those parameters directly in the code now. But I do like the idea of forcing standard parameters across classes to simplify things and forcing checks like, does it have a tract ID if the `GEO_LEVEL` is `tract`, or a CBG ID if the `GEO_LEVEL` is `CBG`. 

Some of the things in those suggested parameters I've been considering as part of https://github.com/usds/justice40-tool/issues/1013 too. This is good motivation to start down that path! 

(more responses to come)

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -36,6 +73,54 @@ def test_init(self, mock_etl, mock_paths):
         assert etl.GEOID_FIELD_NAME == "GEOID10"
         assert etl.GEOID_TRACT_FIELD_NAME == "GEOID10_TRACT"
 
+    @mock.patch("data_pipeline.utils.requests")
+    def test_extract(self, requests_mock, mock_paths):
```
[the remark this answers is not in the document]
lucasmbrown-usds: Yeaaa good catch! To be honest I am much more familiar with this individualized way of doing a `@mock.patch` on the method, but it is out-of-sync with how the rest of the test fixtures work and I shouldl take the extra step to learn something new today and bring more consistency to the codebase. 

Thanks for flagging this, I will refactor!

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -10,8 +14,41 @@
...
+Then running `UPDATE_TEST_FIXTURES = True` will update all subsequent files.
+
+If you're confused by any of this, ask for help, it's confusing :).
+"""
+UPDATE_TEST_FIXTURES = False
+
 
 class TestNationalRiskIndexETL:
+    def test_update_test_fixtures(self, mock_paths, mock_etl):
+        """Assert that UPDATE_TEST_FIXTURES is False."""
```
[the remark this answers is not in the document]
lucasmbrown-usds: Overall these are really great flags, thanks. And I appreciate your care for detail and the extremely appropriate paranoia about what a future distracted developer might do with this! 

A few notes. 

Overall my philosophy here was to try to replicate the philosophy of [snapshot testing](https://blog.bitsrc.io/whats-snapshot-testing-and-can-we-use-it-for-frontend-f5a441cae27a) from the front-end world for backend ETL scripts. (Now that I'm talking about it, I realize that I 100% stole this idea from our friend @dfkoh !) Front-end snapshots will update with a single button press, and thus indeed it's true that the code under test will create the updated test output. 

With front-end snapshot testing, the approach tries to balance two needs: 

1. Alert the developer that their code has caused something to change in the site rendering, which may be intentional or unintentional. 
2. Make the tests easy enough to update that developers actually maintain good code coverage. 

In addition, by making the changes in the snapshot part of the codebase, any changes in outputs are subject to peer review. 

While updating the NRI tests for various small changes, I found myself getting fairly tired of how much work it takes to update the fixtures. (We have a similar challenge with https://github.com/usds/justice40-tool/issues/1048.) 

I'd love to have a way that easily updates the fixtures *after* I've carefully reviewed them to make sure that the outputs are indeed as expected. This also shows up in the `git diff` for a pull request. But as with front-end testing, to quote that link above, "At first many developers tend to blindly update the snapshots when they come across a failing test. However, blindly updating snapshots should never be done." 

I like the idea of making this a common method! And also maybe it shouldn't be a `test_` method at all, but instead a non-test method that gets run explicitly by the developer. Hmm.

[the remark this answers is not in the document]
vim-usds: @lucasmbrown-usds - strongly agree on this:

> "At first many developers tend to blindly update the snapshots when they come across a failing test. However, blindly updating snapshots should never be done."

Given how simple it is to update snapshots on the FE, under deadlines, I am guilty of this (on other projects). However reading failures make take a few extra minutes and worth the extra time to understand why it's failing rather than blindly updating all snapshots.

I also would say that updating ALL snapshots versus updating the one that is failing is also a good practice.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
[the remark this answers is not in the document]
widal001: @lucasmbrown-usds After reflecting on this further, I realized that if we're abstracting these tests to a base test class (`TestETLBase`) then there probably won't be much added benefit of promoting this segment of code to a pytest fixture since other ETL classes would simply use `super().test_extract()` instead of the `requests_mock` directly. I'm all for learning new stuff, haha, but it seems like you could de-scope this from the immediate refactor tasks you outlined below

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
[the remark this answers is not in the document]
widal001: This is such a cool strategy! Doesn't surprise me that snapshot tests for backend code is David's brainchild haha. I'll definitely have to adopt them as a part of my testing toolkit.

lucasmbrown-usds:
> * Update validate_output and make sure it runs either as part of the tests or as part of the ETL itself

FYI I'm going to punt this to https://github.com/usds/justice40-tool/issues/1186.

widal001:
@lucasmbrown-usds Sorry I missed this question earlier!

> I actually had never noticed `validate_output` on the base class, that's super cool. Question: do you think it's smarter to run that as part of the test, or as part of each ETL (e.g., make the ETL runner call `extract`, `transform`, `load`, and then `validate_output`)? Or maybe in both places!

Eventually this would be ideal to have as a part of the ETL following `transform` so it could be called by the ETL runner but I was anticipating excluding it for now to avoid breaking the existing ETL workflows that might not adhere to the contract conditions tested in `validate_output()`

As a bridge to updating the existing ETL classes, it would be great to include it in the tests. In particular I think it would be crucial to include in the refactored `update_test_fixtures()` because it could ensure that the fixtures we're testing the ETL against adhere to that contract. We can do a deeper dive at 2pm to walk through some ideas for incorporating `validate_output()` in the work you've done on that function

lucasmbrown-usds:
FYI @emma-nechamkin @widal001 -- I've implemented some of the code for updating test fixtures. 

I created an ExampleETL class used by the base ETL test class. All it does is unzip a CSV and multiply the column by 2. I updated a cell in the "input.zip" fixture (from 3.1 to 6.1), and then the following rows updated in the 3 "downstream" CSVs in this nice clean diff. 🎉 

Screen Shot 2022-01-27 at 4 50 33 PM

data/data-pipeline/data_pipeline/tests/base/config.yaml
```diff
@@ -1,10 +0,0 @@
-name: Template
```
lucasmbrown-usds, line 1: These files are all from an older approach to dataset schemas that's out of date.

data/data-pipeline/data_pipeline/tests/conftest.py
```diff
@@ -22,8 +22,9 @@ def copy_data_files(src: Path, dst: Path) -> None:
     """
     if not dst.exists():
         dst.parent.mkdir(parents=True, exist_ok=True)
-        copyfile(src, dst)
-        assert dst.exists()
+
+    copyfile(src, dst)
```
lucasmbrown-usds, line 26: This was a small but important bug! The data was only copied if the destination file didn't exist.

data/data-pipeline/data_pipeline/tests/conftest.py
```diff
@@ -34,24 +35,17 @@ def mock_paths(tmp_path_factory) -> tuple:
...
+
     # creates DATA_PATH and TMP_PATH directories in temp directory
     data_path = tmp_path_factory.mktemp("data", numbered=False)
     tmp_path = data_path / "tmp"
     tmp_path.mkdir()
-    return data_path, tmp_path
-
 
-@pytest.fixture(scope="session")
-def mock_census(mock_paths) -> Path:
```
lucasmbrown-usds, line 45: No longer needed.

data/data-pipeline/data_pipeline/tests/sources/example/etl.py
```diff
@@ -0,0 +1,58 @@
...
+        df: pd.DataFrame = pd.read_csv(
+            self.TMP_PATH / "input.csv",
+            dtype={self.GEOID_TRACT_FIELD_NAME: "string"},
+            na_values=["None"],
+            low_memory=False,
+        )
+
+        df[self.EXAMPLE_FIELD_NAME] = df[self.INPUT_FIELD_NAME] * 2
+
+        self.output_df = df
```
lucasmbrown-usds, line 58: `load` is now implemented on the base class, so nothing more needed here. 🎉 !

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -1,17 +1,58 @@
...
+class TestNationalRiskIndexETL(TestETL):
+    _ETL_CLASS = NationalRiskIndexETL
+    _DATA_DIRECTORY_FOR_TEST = (
+        settings.APP_ROOT / "tests" / "sources" / "national_risk_index" / "data"
+    )
+
+    def _setup_etl_instance_and_run_extract(
+        self, mock_etl, mock_paths
+    ):
+        with mock.patch('data_pipeline.utils.requests') as requests_mock:
```
lucasmbrown-usds, line 28: moved this from being a decorator to being a context manager so that the signature of `_setup_etl_instance_and_run_extract` is identical between the base and child classes. 

Also sadly @widal001 I don't think this makes sense to move to conftest.py, because we're monkeypatching `requests` in a very specific way that's going to return the NRI data, so I don't think it's useful on other classes. Do you agree?

data/data-pipeline/data_pipeline/tests/sources/test_etl_base.py
```diff
@@ -0,0 +1,244 @@
...
+        )
+        etl.transform()
+
+        assert etl.output_df is not None
+
+        # Assert it has some rows
+        assert etl.output_df.shape[0] > 0
+
+        # Check that it has all columns
+        for col in etl.COLUMNS_TO_KEEP:
```
emma-nechamkin, line 201: I think if there's no uniqueness test for the columns, it might make sense to check shape[1] with the idea that having every column and having every column precisely one time are not the same.

data/data-pipeline/data_pipeline/tests/sources/test_etl_base.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # setup - load expected output
+        expected_output = pd.read_csv(
+            filepath_or_buffer=expected_output_csv_path,
+            dtype={etl.GEOID_TRACT_FIELD_NAME: str},
+        )
+
+        # check that the `COLUMNS_TO_KEEP` are in the output
+        for col in etl.COLUMNS_TO_KEEP:
+            assert col in actual_output.columns, f"{col} is missing from output"
```
emma-nechamkin, line 241: Same point about uniqueness, feel free to disregard

emma-nechamkin (commented):
This is quite clear and I think comprehensive! Small nits only.

vim-usds:
Assumption here is that this is for the first ETL and then we will expand to others

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -26,15 +66,86 @@ def test_init(self, mock_etl, mock_paths):
...
+
+        # After running load, write the results as the "output.csv" in the test
+        # directory.
+        etl.load()
+        copy_data_files(
+            src=etl._get_output_file_path(),
+            dst=self._DATA_DIRECTORY_FOR_TEST / self._OUTPUT_CSV_FILE_NAME,
+        )
+
+    def test_extract(self, mock_etl, mock_paths):
```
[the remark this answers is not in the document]
lucasmbrown-usds: good call - apparently i hadn't!

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -76,23 +189,34 @@ def test_load(self, mock_etl):
...
+        expected_output = pd.read_csv(
+            filepath_or_buffer=expected_output_csv_path,
+            dtype={etl.GEOID_TRACT_FIELD_NAME: str},
+        )
 
         # validation
-        assert output_path.exists()
-        assert output.shape == (55, 5)
-        pd.testing.assert_frame_equal(output, expected, rtol=1e-10, atol=1e-10)
+        assert actual_output.shape == (55, 5)
```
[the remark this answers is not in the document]
lucasmbrown-usds: Yeah, good point. This is redundant. I'll remove. That also means we can probably move this to the base class!

data/data-pipeline/data_pipeline/tests/sources/test_etl_base.py
```diff
@@ -0,0 +1,244 @@
...
+        needs to explicitly define how to update the `input` fixture that comes after
+        the extract step.
+
+        Using this method to update fixtures  can be helpful if you expect the
+        results to change because you changed the logic of the ETL class and need to
+        quickly update the fixtures.
+
+        However, note a few things first:
+
+        1. Do *not* update these fixtures if you did not expect the ETL results to
```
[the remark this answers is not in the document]
lucasmbrown-usds: Right right. Let's have that conversation over slack or in person at some point -- I could be persuaded either the "one flag per fixture" or "one flag all fixtures" route.

data/data-pipeline/data_pipeline/tests/sources/test_etl_base.py
```diff
@@ -0,0 +1,244 @@
...
+        assert etl.DATA_PATH == data_path
+        assert etl.TMP_PATH == tmp_path
+
+        # Also make sure all parameters that need to be non-null are non-null
+        assert etl.NAME is not None
+        assert etl.LAST_UPDATED_YEAR is not None
+        assert etl.SOURCE_URL is not None
+        assert etl.GEO_LEVEL is not None
+        assert etl.COLUMNS_TO_KEEP is not None
+        assert len(etl.COLUMNS_TO_KEEP) > 0
```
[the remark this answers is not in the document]
lucasmbrown-usds: love it!

data/data-pipeline/data_pipeline/tests/sources/test_etl_base.py
lucasmbrown-usds: I added a uniqueness test on `COLUMNS_TO_KEEP` - let me know if you still think we should update this.

lucasmbrown-usds:
> Assumption here is that this is for the first ETL and then we will expand to others

Indeed. I created #1217 with a checklist (that we should try out and find out what it's missing!) for each other ETL class.

data/data-pipeline/data_pipeline/tests/sources/test_etl_base.py
emma-nechamkin: I think the uniqueness test is sufficient!

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -27,72 +36,68 @@ class ExtractTransformLoad:
...
     # TODO: investigate. Census says there are only 217,740 CBGs in the US. This might
     #  be from CBGs at different time periods.
     EXPECTED_MAX_CENSUS_BLOCK_GROUPS: int = 250000
 
+    # Eleven digits in a census tract ID.
+    EXPECTED_CENSUS_TRACTS_CHARACTER_LENGTH: int = 11
     # TODO: investigate. Census says there are only 74,134 tracts in the US,
     #  Puerto Rico, and island areas. This might be from tracts at different time
     #  periods. https://github.com/usds/justice40-tool/issues/964
     EXPECTED_MAX_CENSUS_TRACTS: int = 74160
```
emma-nechamkin, line 78: very annoying QQ -- wondering if we can set this in a constants file rather than in the class itself, with the idea that if the data changes, it will be less annoying to change? no strong feelings though.

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -116,63 +124,150 @@ def transform(self) -> None:
...
+                    raise ValueError(
+                        f"Too many rows: `{self.output_df.shape[0]}` rows in "
+                        f"output exceeds expectation of `{expected_rows}` "
+                        f"rows."
+                    )
+
+                geo_field_character_lengths = (
+                    self.output_df[geo_field].str.len().unique()
+                )
+
```
emma-nechamkin, line 193: I think using `nunique` would be clearer, but would contain less information?

Like, `assert self.output_df[geo_field].str.len().nunique() == 1` and then `len(self.output_df[geo_field].values[0]) == `expected` is more legible to me personally, but again, no strong preference here.

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -116,63 +124,150 @@ def transform(self) -> None:
...
+                elif (
+                    geo_field_character_lengths[0]
+                    != expected_geo_field_characters
+                ):
+                    raise ValueError(
+                        "Wrong character length: some of the census geography data "
+                        "has the wrong length."
+                    )
+
+                non_unique_geo_field_values = len(
```
emma-nechamkin, line 209: I think this would be clearer with `nunique` but it's a personal preference

```
self.output_df[geo_field].shape[0] == self.output_df[geo_field].nunique()
```

data/data-pipeline/data_pipeline/etl/runner.py
```diff
@@ -60,6 +60,9 @@ def etl_runner(dataset_to_run: str = None) -> None:
         # run load
         etl_instance.load()
 
+        # run validate
+        etl_instance.validate()
```
emma-nechamkin, line 64: 🥇 yay

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -26,73 +96,56 @@ def test_init(self, mock_etl, mock_paths):
...
+            == "https://hazards.fema.gov/nri/Content/StaticDocuments/DataDownload//NRI_Table_CensusTracts/NRI_Table_CensusTracts.zip"
+        )
+        assert etl.GEO_LEVEL == ValidGeoLevel.CENSUS_TRACT
+        assert etl.COLUMNS_TO_KEEP == [
+            etl.GEOID_TRACT_FIELD_NAME,
+            etl.RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME,
+            etl.EXPECTED_POPULATION_LOSS_RATE_FIELD_NAME,
+            etl.EXPECTED_AGRICULTURE_LOSS_RATE_FIELD_NAME,
+            etl.EXPECTED_BUILDING_LOSS_RATE_FIELD_NAME,
+        ]
```
emma-nechamkin, line 117: I am a little confused here from the perspective of updating / changing data -- like it would be really nice imo if all of this were housed in a single place outside of the test, with the idea that it could be shared by the etls themselves and also the tests -- but it's entirely possible i'm not understanding something here :)

emma-nechamkin (commented):
I don't have any substantive comments -- this looks great to me!

esfoobar-usds (approved):
This looks good to me. The only small nit is around clarity of when there should be pickles generated and when not, but let's see how it goes with the next tests we write.

🍏

lucasmbrown-usds:
> The only small nit is around clarity of when there should be pickles generated and when not

Yeah right now we're comparing only against the values of the dataframes, not full pickles which would also allow us to compare types (but also, pickles introduce some challenges around the version of pandas used to create the files). I think the main downside of switching to pickles would be that the diffs in the snapshot updates would no longer be human-readable and would make oversight of those changes a lot harder.

data/data-pipeline/data_pipeline/etl/base.py
lucasmbrown-usds: Makes sense! Updated.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -111,17 +114,7 @@ def extract(self) -> None:
...
-            / "dataset"
-            / "national_risk_index_2020"
-            / "usa.csv"
-        )
-        self.national_risk_index_df = pd.read_csv(
-            national_risk_index_csv,
-            dtype={self.GEOID_TRACT_FIELD_NAME: "string"},
-            low_memory=False,
-        )
+        self.national_risk_index_df = NationalRiskIndexETL.get_data_frame()
```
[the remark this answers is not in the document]
lucasmbrown-usds: I agree! I've been dreaming of this day 🤓

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
lucasmbrown-usds: Oh I see -- your idea would be like, maybe there's a configuration file or something where we specify this, and then we're loading that configuration from both the ETLs and the tests? 

That is fairly reasonable to me (and not far off from the ideas of the Big Refactor). At the same time, with unit tests, it's really good to actually avoid being DRY (see this [post](https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html), or the principle of having tests be ["WET"](https://www.heroku.com/podcasts/codeish/102-whether-to-repeat-yourself-dry-damp-or-wet) (write everything twice). 

So I appreciate that if someone needs to update the columns to keep, that the test will explicitly fail and they have to manually update it. This is can catch an accidental modification of COLUMNS_TO_KEEP that would get loaded into the test from the same config and pass. 

What do you think?

emma-nechamkin: That seems reasonable. It's slightly at odds with the "update all pickles!" flag but :shrug: probably in a good way. I'm convinced!

widal001 (commented):
## Summary

@lucasmbrown-usds This all looks solid! Most of the comments I left were for forward-looking considerations beyond the scope of this PR. 

The only immediate consideration before merging is to rename `TestETL.test_update_fixtures()` to `TestETL.update_fixtures()` to avoid accidentally updating the fixtures in the ETL test classes that inherit from when `test_update_fixtures()` isn't re-implemented correctly in those classes. This change would prevent pytest from executing that code unless a derivative test specifically implemented `test_update_fixtures()` and used a `super().update_fixtures()` to invoke that method.

## Other Thoughts/Suggestions

While these thoughts shouldn't be blockers for merging this PR I wanted to highlight some common themes from my comments and areas where I think they overlapped with some of @emma-nechamkin 's questions/suggestions:

- **Separate Config Files** The design pattern that we currently follow is to define a lot of attributes about the ETL source as class globals that need to be set in each ETL instance and potentially mocked out in testing. One general question I think would be worth investigating (especially alongside the broader ETL refactor we discussed) is opportunities to standardize the attributes we set across ETL classes and define them in either a localized config/constants file or in a central config/constants file. I know there was a desire to move away from `yaml` files so I'm sure there's a good justification for avoiding this approach, but I might have missed the explanation around that design choice.
- **Canonical CBG and Census Tract Dataset** A specialized case of the approach described above is how we've hard coded a lot of attributes about the official set of CBGs and Census Tracts in the Base ETL class (e.g. number of blocks/tracts, shape of the resulting field values, etc.) This is probably fine given that these values won't change often, but I do wonder if it would be valuable to establish a canonical set of CBGs and Census Tracts (instead of hard-coding attributes about those geographic units) and use that canonical set to validate the geographic outputs of the subsequent ETL sources

data/data-pipeline/data_pipeline/etl/base.py
lucasmbrown-usds: Good idea. Created this to track: https://github.com/usds/justice40-tool/issues/1245

data/data-pipeline/data_pipeline/tests/sources/test_etl_base.py
```diff
@@ -0,0 +1,549 @@
...
+
+    Note: every method that does *not* need to be reimplemented by child classes has
+    the test name pattern of `test_*_base`. All other tests need to be reimplemented.
+    """
+
+    # In every child test class, change this to the class of the ETL being tested.
+    _ETL_CLASS = ExampleETL
+
+    # In every child test class, change this to the directory of the text fixtures.
+    _DATA_DIRECTORY_FOR_TEST = (
```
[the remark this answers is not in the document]
lucasmbrown-usds: Looks like we can't use `__init__` (https://stackoverflow.com/questions/62460557/cannot-collect-test-class-testmain-because-it-has-a-init-constructor-from), but I implemented a different version of this. Good idea!

data/data-pipeline/data_pipeline/tests/sources/test_etl_base.py
```diff
@@ -0,0 +1,549 @@
...
+        elif etl.GEO_LEVEL == ValidGeoLevel.CENSUS_BLOCK_GROUP:
+            assert pd.api.types.is_string_dtype(
+                output_df[ExtractTransformLoad.GEOID_FIELD_NAME]
+            )
+
+        else:
+            raise NotImplementedError("This geo level not tested yet.")
+
+    # TODO: Add a flag to make this run only when pytest is run with an argument.
+    def test_update_test_fixtures(self, mock_etl, mock_paths):
```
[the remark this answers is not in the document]
lucasmbrown-usds: Hmm. I'm not sure I see what you're driving at here. The following lines: 

```
        if self._ETL_CLASS is not ExampleETL:
            raise NotImplementedError(
                "Update fixtures method not defined for this class."
            )
```

Will force all child classes to reimplement this. Does that seem sufficient to you?

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -116,63 +124,146 @@ def transform(self) -> None:
...
+                "`self.COLUMNS_TO_KEEP` must be specified."
+            )
 
-        remove_all_from_dir(self.TMP_PATH)
+        if self.output_df is None:
+            raise NotImplementedError(
+                "The `transform` step must set `self.output_df`."
+            )
+
+        for column_to_keep in self.COLUMNS_TO_KEEP:
```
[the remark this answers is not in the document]
emma-nechamkin: I'm kinda with @widal001 here and kind of not -- I think canonical census tracts would be a good way to ensure we are getting a complete set of data each time, but on the other hand, data can change and so too can census tracts and having a canonical set of census tracts could be a problem?

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -1,8 +1,10 @@
+import enum
+import pathlib
```
emma-nechamkin, line 2: a real nit -- since we import pathlib, can we refer to Path as pathlib.Path and then delete line 4? (I am sorry)

emma-nechamkin (commented):
i don't have much to add but this looks GREAT and I am looking forward to implementing it all!

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
lucasmbrown-usds: Right - and to my knowledge, very few of the sources have all of the same tracts, so we'd have to have some thoughtful logic about how to decide what exactly we're comparing against the canonical tracts.

data/data-pipeline/data_pipeline/etl/base.py
lucasmbrown-usds: Yessss eagle eyes! I love it. 

TBH I'm always in favor of importing top-level modules and not specific items from the module. It is a little chattier, but the code is more readable because you know where that thing came from.

PR #1089, opened 2021-12-29
Tree Equity Output: Change output from Geojson to CSV format for easier analysis

saran-ahluwalia:
Addresses #993

Steps to reproduce:  `poetry run python3 data_pipeline/application.py etl-run --dataset tree_equity_score`

Aggregation is at the block group - which means in the future we will need to perform an additional aggregation and reduction to align with Census tracts.


data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -7,6 +7,19 @@
 
 
 class TreeEquityScoreETL(ExtractTransformLoad):
+    """
```
lucasmbrown-usds, line 10: love these docstrings! 

(if you want, same comment as rounding PR: move a <= 1-line summary of this class to the first line.)

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
```diff
@@ -83,8 +96,42 @@ def transform(self) -> None:
...
+        )
+
     def load(self) -> None:
-        logger.info("Saving Tree Equity Score GeoJSON")
+        logger.info("Saving Tree Equity Score CSV")
         # write nationwide csv
         self.CSV_PATH.mkdir(parents=True, exist_ok=True)
-        self.df.to_file(self.CSV_PATH / "tes_conus.geojson", driver="GeoJSON")
+        self.df = self.df[
+            [
```
lucasmbrown-usds, line 111: Looks good! If you want, you can follow the conventions by moving this list to `self.COLUMNS_TO_KEEP` defined on the class – see `data/data-pipeline/data_pipeline/etl/sources/ejscreen/etl.py` for an example. 

But that's not required, this is fine too.

lucasmbrown-usds (approved):
One small comment, take it or leave it!

data/data-pipeline/data_pipeline/etl/sources/tree_equity_score/etl.py
lucasmbrown-usds: Looks like you decided not to change this -- that's fine!

lucasmbrown-usds (approved):
LGTM!

PR #1067, opened 2021-12-16
Add Maryland EJ Screen into data-pipeline's ETL and provide automated scoring and statistics outputs

saran-ahluwalia:
Addresses #239 

To replicate: 

1. `poetry run python3 data_pipeline/application.py etl-run --dataset mdejscreen`
2. `poetry run python3 data_pipeline/application.py score-run`
3. In the comparison tool Jupyter notebook run the entire notebook to generate scores for Maryland based on the 90th percentile

Outstanding questions:

1. Is the level of detail and comments all necessary and clear?
2. Is it required to include the output Maryland basic statistics for burdened tracts?
3. The threshold for prioritized communities that is used here is the 90th percentile. Is this the best threshold to use?  The literature from MD EJ Screen doesn't offer too much guidance.

Issue #239, opened 2021-06-28
As a contributor to J40, I want to compare the J40 score results for the state of Maryland to Maryland EJScreen data, so that I can understand how the J40 score compares.


BethMattern:
**Description**
The J40 team is working to develop the best possible score for identifying the most underserved and overburdened communities. In order to assess the quality of the score and build credibility in the EJ community, the J40 team build a tool to compare the output of the most recent version of the J40 score with other data sources that generate scores.

We need help adding more data sources for comparison to our tool. We would like to compare our Maryland results to the Maryland EJSCreen score, that can be found here: https://www.adaptationclearinghouse.org/resources/maryland-environmental-justice-screen-tool-md-ejscreen.html

**Tasks**
- [x] ETL Maryland EJScreen score
- [x] Modify comparison tool script to load Maryland EJScreen data and compare it against current score
- [x] Modify comparison tool script to add stats about the comparison with Maryland EJScreen with the current score to the report generated at the end of the script

**Definition of "Done"**
- [x] The J40 comparison tool can be run using the most recent version of our score
- [x] The comparison tool can access the Maryland EJScreen data
- [x] A report / data export can be generated that outputs the results of the comparison on a census tract level

saran-ahluwalia:
Data is supposed to be found here: https://github.com/gisumd/EJScreen.

saran-ahluwalia:
One thing that I was not able to access was the EJS raw data from the reference link that was stated in the document.

lucasmbrown-usds:
@saran-ahluwalia -- data now available at: 

```
(
            settings.AWS_JUSTICE40_DATASOURCES_URL
            + "/MD_EJScreen.zip"
        )
```

saran-ahluwalia:
Thanks!

saran-ahluwalia:
Regarding the EJScore in the MDEJscreen. This is the distribution of percentile scores in which each score - call it `N` - represents that census tract's respective overall score is higher than `N`% of the tracts in the state.

The aggregate `EJScore`, here, is the product of the empirical average of the two "Pollution Burden" scores and the empirical average of the "Population Characteristics" categories.

<img width="433" alt="Screen Shot 2021-12-16 at 7 48 41 AM" src="https://user-images.githubusercontent.com/94847739/146375104-e74f080e-7c11-4789-bfd9-b09548f3edcd.png">

saran-ahluwalia:
[...]

Issue #1062, opened 2021-12-14
Implement list of changes to export files


BethMattern:
Implement all the changes outlined in "Indicators for UI and Export" and "Export file changes"

switzersc-usds:
Will review with @esfoobar-usds today

PR #1093, opened 2021-12-31
Add Maryland EJSCREEN into data-pipeline's ETL and provide automated scoring and statistics outputs 

saran-ahluwalia:
Addresses #239 

To replicate: 

1. `poetry run python3 data_pipeline/application.py etl-run --dataset maryland_ejscreen`
2. `poetry run python3 data_pipeline/application.py score-run`
3. In the comparison tool Jupyter notebook run the entire notebook to generate scores for Maryland based on the 90th percentile

Outstanding questions:

1. Is the level of detail and comments all necessary and clear?
2. Is it required to include the output Maryland basic statistics for burdened tracts?
3. The threshold for prioritized communities that is used here is the 90th percentile. @lucasmbrown-usds Regarding this comment: https://github.com/usds/justice40-tool/pull/1067/files#r776754636. I have added some details in the README that clarify the rationale for choosing the 90th percentile for areas of concern. To my knowledge, there is not an explicit recommendation for which percentile determines burden.


data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,115 @@
...
+        combined_df = combined_df[combined_df["Census_Tra"] != 0]
+
+        # Set our class instance variable.
+        self.df = combined_df.copy()
+
+        # Rename
+        self.df.rename(
+            columns={
+                "Census_Tra": self.GEOID_TRACT_FIELD_NAME,
+                "EJScore": field_names.MARYLAND_PERCENTILE_FIELD_NAME,
```
lucasmbrown-usds, line 88: If I'm reading the documentation and data correctly, this is one of those very small mistakes that has massive implications for our data analysis. I'm pretty sure the EJScore is *not* a percentile – for instance, 

> The two "Pollution Burden" average scores are then averaged together and the result is multiplied by the average of the "Population Characteristics" categories to get the total EJ Score for each tract. 

This is somewhat obvious from looking at the data and the results -- only 2% of Maryland's tracts are > .90 on the EJScore. If it was a percentile, that would be 10% of Maryland's tracts by definition. 

If you agree with this analysis, could you update and re-run the statistical output? Thanks!

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/README.md
```diff
@@ -0,0 +1,24 @@
...
+According to the [documentation](https://p1.cgis.umd.edu/mdejscreen/help.html):
+
+There exist two data categories: Population Burden and Population Characteristics.
+
+There are two indicators within Population Burden: Exposure, and Socioeconomic. Within Population Characteristics, there exist two indicators: Sensitive, Environmental Effects. Each respective indicator contains several relevant covariates, and an averaged score. 
+
+The two "Pollution Burden" average scores are then averaged together and the result is multiplied by the average of the "Population Characteristics" categories to get the total EJ Score for each tract.
+
+For each indicator, the percentile is given. For example, the indicator value for "Asthma Emergency Discharges" with 0.9 is therefore in the 90th percentile, which means only 10% of tracts in Maryland have higher values. EJ Scores near 1 represent areas of the greatest environmental justice concern.
+
```
lucasmbrown-usds, line 16: nit: unnecessary extra line

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+A study of Bladensburg, MD - located in Prince George’s County - demonstrated the application of the MD EJSCREEN (Driver et al., 2019). According to the study, The Bladensburg population is primarily Black (62.7%) and Latinx (33.0%), with 20.1% of the community members living below the federal poverty line. Through an analysis, leveraging the Maryland EJSCREEN, Bladensburg with MD EJSCREEN, the researchers found that Bladensburg has an EJ score higher than 99% of the census tracts in Prince George’s County, indicating a higher prevalence of environmental hazards in the region.
```
lucasmbrown-usds, line 18: I'm not sure the demographics are relevant here but the poverty line data is.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/README.md
```diff
@@ -0,0 +1,24 @@
+# Maryland EJSCREEN
+
+The Maryland EJSCREEN application and tool can be found [here](https://p1.cgis.umd.edu/mdejscreen/).
+
+### Methodology Summary
```
lucasmbrown-usds, line 5: Super helpful summary!! Awesome. Setting a good standard for ETL READMEs.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __init__(self):
+        self.MARYLAND_EJSCREEN_URL = (
+            settings.AWS_JUSTICE40_DATASOURCES_URL + "/MD_EJScreen.zip"
+        )
+
+        self.SHAPE_FILES_PATH = self.TMP_PATH / "mdejscreen"
+        self.OUTPUT_CSV_PATH = self.DATA_PATH / "dataset" / "maryland_ejscreen"
+
+        self.COLUMNS_TO_KEEP = [
```
lucasmbrown-usds, line 27: You don't actually use this in `load` so it's not currently doing anything.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,115 @@
...
+            for df in dfs_list
+        ]
+        # pylint: disable=unsubscriptable-object
+        combined_df = gpd.GeoDataFrame(pd.concat(dfs_list, axis=1))
+
+        # Reset index so that we no longer have the tract as our index
+        combined_df = combined_df.reset_index()
+        # coerce into integer into
+        # pylint: disable=unsupported-assignment-operation, unsubscriptable-object
+        combined_df["Census_Tra"] = (combined_df["Census_Tra"]).astype(int)
```
lucasmbrown-usds, line 75: Nit: This is fine here because maryland's geoids start with "24", but for future reference if you were in a state like alabama (fips 01), converting to int would lose the leading 0 and make this geoid invalid.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Drop the 10 census tracts that are zero: please see here:
+        # https://github.com/usds/justice40-tool/issues/239#issuecomment-995821572
```
lucasmbrown-usds, line 78: When we're ready to send an email, let's contact the Maryland EJSCREEN authors about this.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,115 @@
...
+        combined_df = combined_df[combined_df["Census_Tra"] != 0]
+
+        # Set our class instance variable.
+        self.df = combined_df.copy()
```
lucasmbrown-usds, line 82: I'm curious: why are you doing this in the middle of the method, rather than at the end?

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,115 @@
...
+            columns={
+                "Census_Tra": self.GEOID_TRACT_FIELD_NAME,
+                "EJScore": field_names.MARYLAND_PERCENTILE_FIELD_NAME,
+            },
+            inplace=True,
+        )
+
+        # Baseline Comparisons with some quartiles and the 90th percentile.
+        # Interpretation: The score is greater than or equal to N% of the tracts in the state.
+        self.df[field_names.MARYLAND_EJSCREEN_TRACT_25_PERCENT_FIELD] = (
```
lucasmbrown-usds, line 95: Nice, I like you calculated at all these cutoffs.

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -214,6 +214,24 @@
 HOLC_GRADE_D_TRACT_75_PERCENT_FIELD: str = "Tract is >75% HOLC Grade D"
 
 
+# Maryland EJSCREEN Data.
```
lucasmbrown-usds, line 217: :tada:

lucasmbrown-usds (changes requested):
A few comments, thanks. Overall looking like good progress.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
saran-ahluwalia: Good point

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
saran-ahluwalia: Hi @lucasmbrown-usds - I am not clear nor do I completely agree with the statements outlined here. Citing [this](https://p1.cgis.umd.edu/mdejscreen/help.html) and below from the documentation, I think there may be a misunderstanding? I am happy to clarify and address your specific concerns. 

Screen Shot 2021-12-31 at 1 01 14 PM

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/README.md
saran-ahluwalia: Is there any action to take here?

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
lucasmbrown-usds: Or, alternatively, if we get the sense that Maryland is prioritizing *not* based on the percentile but based on the EJScore (raw) itself, that's fine. We can rename these fields to remove all references to percentiles (since they're not percentiles) choose an appropriate cutoff for the comparison tool. Maybe we re-run the comp tool with MARYLAND_EJSCREEN_TRACT_75_PERCENT_FIELD. Thanks!

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
saran-ahluwalia: I understand this - that is why I felt comfortable doing this at the time.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
saran-ahluwalia: As we discussed - when you return we have several things to raise with Dr. Wilson

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
saran-ahluwalia: A reflection of my thought process

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,128 @@
...
+                "Census_Tra": self.GEOID_TRACT_FIELD_NAME,
+                "EJScore": field_names.MARYLAND_EJSCREEN_SCORE_FIELD,
+            },
+            inplace=True,
+        )
+
+        self.df[field_names.MARYLAND_EJSCREEN_PERCENTILE_FIELD] = self.df[
+            field_names.MARYLAND_EJSCREEN_SCORE_FIELD
+        ].rank(
+            pct=True,
```
saran-ahluwalia, line 98: this computes the _ranked percentile_ of the EJ Score (which _is_ a percentile. Please see here for [details](https://github.com/usds/justice40-tool/issues/239#issuecomment-1003567593)

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,127 @@
...
+        self.df[field_names.MARYLAND_EJSCREEN_TRACT_90_PERCENT_FIELD] = (
+            self.df[field_names.MARYLAND_EJSCREEN_SCORE_FIELD] >= 0.90
+        )
+
+    def load(self) -> None:
+        logger.info("Saving Maryland EJSCREEN CSV")
+        # write maryland tracts to csv
+        self.OUTPUT_CSV_PATH.mkdir(parents=True, exist_ok=True)
+        self.df[self.COLUMNS_TO_KEEP].to_csv(
+            self.OUTPUT_CSV_PATH / "maryland_ejscreen.csv", index=False
```
lucasmbrown-usds, line 126: Nit: all our other classes write to `"usa.csv"`, you could possibly change to that or just `"maryland.csv`", but the EJSCREEN part will be obvious through the directory.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
lucasmbrown-usds: Can we change this? I'm not sure it makes a ton of sense to copy this in the middle of the method. It would make more sense to copy it either at the beginning or the end.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,127 @@
...
+            columns={
+                "Census_Tra": self.GEOID_TRACT_FIELD_NAME,
+                "EJScore": field_names.MARYLAND_EJSCREEN_SCORE_FIELD,
+            },
+            inplace=True,
+        )
+
+        # Interpretation: An EJ score (reported as a percentile)
+        # has a percentile rank of N for some N between 0 - 100"
+        self.df[field_names.MARYLAND_EJSCREEN_PERCENTILE_FIELD] = self.df[
```
lucasmbrown-usds, line 97: 👍

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -219,6 +219,33 @@
     "Michigan EJSCREEN Priority Community"
 )
 
+# Maryland EJSCREEN Data.
+MARYLAND_EJSCREEN_TRACT_25_PERCENT_FIELD: str = (
```
lucasmbrown-usds, line 223: Nit: a percent is a very different thing than a percentile, so can we call all of these `_PERCENTILE_FIELD` instead of `_PERCENT_FIELD`?

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -219,6 +219,33 @@
...
+    "Tract is >=50% all other Maryland Tracts"
+)
+MARYLAND_EJSCREEN_TRACT_75_PERCENT_FIELD: str = (
+    "Tract is >=75% all other Maryland Tracts"
+)
+MARYLAND_EJSCREEN_TRACT_90_PERCENT_FIELD: str = (
+    "Tract is >=90% all other Maryland Tracts"
+)
+
+MARYLAND_EJSCREEN_PERCENTILE_FIELD: str = (
```
lucasmbrown-usds, line 236: I would suggest replacing this with simply referencing `field_names.MARYLAND_EJSCREEN_SCORE_FIELD + field_names.PERCENTILE_FIELD_SUFFIX` everywhere, to follow the pattern used by other code and have a standardized field name.

lucasmbrown-usds (changes requested):
Overall looking good!

data/data-pipeline/data_pipeline/score/field_names.py
saran-ahluwalia: Good point - I will make the change here.

data/data-pipeline/data_pipeline/score/field_names.py
saran-ahluwalia: 👍

saran-ahluwalia: My only concern is that _percentile_ has a very different meaning than _percentile rank_ - as raised in #239. Is this not misleading?

lucasmbrown-usds: We use "percentile" to mean what you're calling "percentile rank" everywhere in this codebase, and it's very standard usage. 

From wikipedia: 

> In statistics, a k-th percentile, is a score below which a given percentage k of scores in its frequency distribution falls or a score at or below which a given percentage falls. For example, the 50th percentile is the score below which or at or below which 50% of the scores in the distribution may be found. 

The percentile is K. 

Let's go ahead with calling these percentiles, as we do everywhere else.

saran-ahluwalia: @lucasmbrown-usds : This definition is  aligned with what I shared and provided [here](https://github.com/usds/justice40-tool/issues/239#issuecomment-1003567593). This is the main area of concern. I note this in the codebase. Just because we use it everywhere else doesn't make this necessarily a correct statement. 

Quoting my reasoning from that link as well:

_A statement such as "a height of 173 cm has a percentile rank of 80" has a different meaning from a statement such as "the 80th percentile is 173 cm."_

_The first of these two statements say that 80% of the population had heights that were less than a height of 173 cm, while the second statement says that 80 percent of the population had a height of 173 cm or less. Statements involving nth percentiles, which are not employed in this case, are typically used for defining the cutoffs when the scores have been arranged numerically into 100 equal intervals._

_In summary, percentile refers to a single point; a percentile rank covers an entire interval. When we start with a given proportion or percentage, we compute a percentile whose value is some raw score; when we start with a given raw score or interval, we compute a percentile rank whose value is between 0 and 100._

saran-ahluwalia: The CDC SVI Index also references percentile ranks as well in their methodology. I am just not comfortable with changing this to something that does not reflect what was actually done to the data. 

I can just remove that percentile rank reference altogether and use the reported EJ Score (which is a percentile).

saran-ahluwalia: > We use "percentile" to mean what you're calling "percentile rank" everywhere in this codebase, and it's very standard usage.
> 
> From wikipedia:
> 
> > In statistics, a k-th percentile, is a score below which a given percentage k of scores in its frequency distribution falls or a score at or below which a given percentage falls. For example, the 50th percentile is the score below which or at or below which 50% of the scores in the distribution may be found.
> 
> The percentile is K.
> 
> Let's go ahead with calling these percentiles, as we do everywhere else.

The percentile is `K` as it pertains to what? A single entity or an interval? My answer in the issue addresses this.

In the same article that you cite, I should have added that there is a section in `Definitions` with the statement `There is no standard definition of percentile`.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
lucasmbrown-usds: Maybe make a note to this effect in the code?

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
lucasmbrown-usds: Why did you mark this as resolved? You haven't addressed the requested change. Thanks!

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -219,6 +219,29 @@
...
+    "Tract is >=90% all other Maryland Tracts"
+)
+
+MARYLAND_EJSCREEN_SCORE_FIELD: str = "Maryland Environmental Justice Score"
+
+# this references percentile ranks
+# please see here for interpretation: https://github.com/usds/justice40-tool/issues/239#issuecomment-1003567593
+# and here: https://github.com/usds/justice40-tool/issues/239#issuecomment-1003448074
+MARYLAND_EJSCREEN_BURDENED_THRESHOLD_FIELD: str = (
+    "Tract has a percentile rank above 75 for the EJ Score"
```
lucasmbrown-usds, line 242: As previously discussed, please call these percentiles and do not call them percentile rank. Also try to match the format you've used for similar field names throughout `field_name`.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
```diff
@@ -0,0 +1,139 @@
...
+                field_names.MARYLAND_EJSCREEN_SCORE_FIELD
+                + field_names.PERCENTILE_FIELD_SUFFIX
+            ]
+            > 0.75
+        )
+
+        # Baseline Comparisons with some quartiles and the 90th percent OF EJ Score
+        # Interpretation: The score is greater than or equal to N% of the tracts in the state.
+        # Please see the README for details on the EJScore interpretation
+        self.df[field_names.MARYLAND_EJSCREEN_TRACT_25_PERCENTILE_FIELD] = (
```
lucasmbrown-usds, line 120: I'm a little surprised that we're still talking about this, but these are NOT percentiles. MARYLAND_EJSCREEN_SCORE_FIELD is not a percentile so labeling these as percentiles is misleading. 

If you're worried about this code, you can also just drop these four fields and just keep `field_names.MARYLAND_EJSCREEN_BURDENED_THRESHOLD_FIELD`.

lucasmbrown-usds (changes requested):
A few changes requested.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
saran-ahluwalia: I will do as you have suggested. I am merely reflecting on what is in the papers and the documentation that was published by the research team.

data/data-pipeline/data_pipeline/score/field_names.py
saran-ahluwalia: I will remove this.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
lucasmbrown-usds: Bump. Same comment.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
lucasmbrown-usds: Same comment - why did you mark this as resolved? You haven't addressed the requested change.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
lucasmbrown-usds: Bump. Doesn't seem like you addressed this comment?

lucasmbrown-usds (changes requested):
I wonder if perhaps a commit didn't get pushed -- there are a few comments from the last round of review that you still haven't addressed.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
saran-ahluwalia: The way these comments come in and out depending on the file changes and the order in which files change often impedes readability. This was missed earlier for that reason.

data/data-pipeline/data_pipeline/etl/sources/maryland_ejscreen/etl.py
saran-ahluwalia: I was trying to reconcile this with the earlier comment here: https://github.com/usds/justice40-tool/pull/1093#discussion_r782537285

I have gone ahead and just removed all references in the source code. The reader will just have to review the README and the papers.

lucasmbrown-usds (approved):
LGTM. Could you run this ETL from scratch and run the comparison tool just to make sure everything works? Otherwise all good, thanks!

saran-ahluwalia:
I had to close this because of issues reconciling changes from #1062

PR #1099, opened 2022-01-05
Re-export requirements.txt to fix version errors

sverchdotgov:
The version of lxml in this file had a known vulnerability that got caught by the "safety" checker, but it is updated in the poetry files.

Regenerated using: https://github.com/usds/justice40-tool/tree/main/data/data-pipeline#miscellaneous


esfoobar-usds (approved):
🍏

Issue #651, opened 2021-09-09
Update Github Actions to use Poetry as the installer and discard the use of requirements.txt and package caching


esfoobar-usds:
**Is your feature request related to a problem? Please describe.**
Right now any new packages added to Poetry require the developer to run a one line command so that Poetry updates `requirements.txt`. This introduces issues for package management and potential issues for missing packages not synced.

**Describe the solution you'd like**
We need to replace the Poetry Github Action package to [Snok/Poetry](https://github.com/snok/install-poetry) and also use [actions/cache](https://github.com/actions/cache) to speed up the CI/CD pipeline.

**Describe alternatives you've considered**
None

**Additional context**
None

sverchdotgov:
@esfoobar-usds I closed https://github.com/usds/justice40-tool/issues/1100 as a duplicate because I think this is the real solution to that problem.

sverchdotgov:
I notice that the current dockerfile references requirements.txt: https://github.com/usds/justice40-tool/blob/main/data/data-pipeline/Dockerfile#L33

That might also explain why we had that pandas warning turn into an error recently, and only in the docker compose setup, since I recently updated requirements.txt.

@esfoobar-usds Are you okay with me using poetry in the docker setup instead of requirements.txt as part of making this change?

esfoobar-usds:
Yes, we can install and run poetry on Dockerfile.

[I have this example](https://github.com/fromzeroedu/quart-postgres-boilerplate/blob/main/Dockerfile#L17-L27) from one of my bootstrap projects.

Issue #1108, opened 2022-01-06
Docker compose fails to build the project end to end on "Rounding Decimals" error


sverchdotgov:
**Describe the bug**
The docker compose workflow described in the [quickstart](https://github.com/usds/justice40-tool/blob/main/QUICKSTART.md) doesn't work.

**To Reproduce**

```
$ docker-compose up --build
j40_data_pipeline_1 | 2022-01-06 08:47:25,574 [data_pipeline.etl.score.etl_score_post] INFO     Rounding Decimals
j40_data_pipeline_1 | Traceback (most recent call last):
j40_data_pipeline_1 |   File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
j40_data_pipeline_1 |     return _run_code(code, main_globals, None,
j40_data_pipeline_1 |   File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
j40_data_pipeline_1 |     exec(code, run_globals)
j40_data_pipeline_1 |   File "/data-pipeline/data_pipeline/application.py", line 283, in <module>
j40_data_pipeline_1 |     cli()
j40_data_pipeline_1 |   File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1128, in __call__
j40_data_pipeline_1 |     return self.main(*args, **kwargs)
j40_data_pipeline_1 |   File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1053, in main
j40_data_pipeline_1 |     rv = self.invoke(ctx)
j40_data_pipeline_1 |   File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1659, in invoke
j40_data_pipeline_1 |     return _process_result(sub_ctx.command.invoke(sub_ctx))
j40_data_pipeline_1 |   File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1395, in invoke
j40_data_pipeline_1 |     return ctx.invoke(self.callback, **ctx.params)
j40_data_pipeline_1 |   File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 754, in invoke
j40_data_pipeline_1 |     return __callback(*args, **kwargs)
j40_data_pipeline_1 |   File "/data-pipeline/data_pipeline/application.py", line 266, in data_full_run
j40_data_pipeline_1 |     score_post(data_source)
j40_data_pipeline_1 |   File "/data-pipeline/data_pipeline/etl/runner.py", line 102, in score_post
[...]

PR #1111, opened 2022-01-07
Remove requirements.txt as a dependency

sverchdotgov:
This converts both docker and tox to use poetry, eliminating usage of requirements.txt in both flows.

- In tox, uses the tox-poetry package which installs dependencies from the lockfile.
- In docker, uses https://stackoverflow.com/questions/53835198/integrating-python-poetry-with-docker as a reference.

Fixes: https://github.com/usds/justice40-tool/issues/651


sverchdotgov:
Marking this as ready for review. I think the checks are currently failing because of https://github.com/usds/justice40-tool/issues/1108.

sverchdotgov:
Hmm, unfortunately [liccheck](https://pypi.org/project/liccheck/) depends on the existence of requirements.txt. Maybe there's a way to generate it during the tox run.

sverchdotgov:
@esfoobar-usds @saran-ahluwalia I tested this build with the fixes for the "pandas assigning to a copy of a slice issue" and everything passed, so once that's merged in this should work.

sverchdotgov:
@esfoobar-usds @saran-ahluwalia Seems like this is passing now! Does this look good? The most unpleasant thing here is the hack to get liccheck to work since it looks at requirements.txt, but at least that will prevent issues with becoming out of sync because it generates it on demand.

data/data-pipeline/tox.ini
```diff
@@ -7,32 +7,17 @@ skip_missing_interpreters = true
...
+allowlist_externals = bash
 commands = safety check
-           liccheck
+           bash scripts/run-liccheck.sh
 
 [testenv:pytest]
 # Run tests
 deps = pytest
-commands = pytest
+commands = pytest --full-trace
```
esfoobar-usds, line 23: What is this flag?

esfoobar-usds (approved):
🍏  -- just one small question

data/data-pipeline/tox.ini
sverchdotgov: Making the output more verbose on failures: https://docs.pytest.org/en/latest/how-to/output.html#modifying-python-traceback-printing

I ran into https://github.com/usds/justice40-tool/issues/1108 while working on this, and the traceback helped.

Issue #599, opened 2021-09-02
As a j40 team member, I want my data pipeline tested so that I know when we generate new versions of the score or available processed data, we'll get good data in the expected format.


switzersc-usds:
**Description**
This user story is all about making sure we have adequate testing in place to consider our backend (data pipeline and score generation) production ready!

**Solution**
Generate score, score-post and score-geo on every merge to a PR where the etl/sources and etl/score folder has changes. 

**Describe alternatives you've considered**
We can not have tests and continually live in a state of anxiety and freakouts and fire-fighting when doing new data processing or score experimentation.

**Links to user research or other resources**

**Tasks**

**Definition of "Done"**

sverchdotgov:
Next step is that @esfoobar-usds and I sync on this to figure out holistically how we want to do end to end testing with docker compose.

Some notes so far:

- Tile generation is CPU heavy and that works, so we think compiles will work in github actions.
- This data is public, so rather than generating a small synthetic dataset, we can instead pull out a subset of the data (maybe just one state).
- We can run this just on merges to main, rather than every push to a PR.

esfoobar-usds:
Unfortunately I think we need to run on every merge. I also believe the data should come from S3 mini-fixtures and not from the data sources themselves. We can discuss this afternoon.

esfoobar-usds:
Also, no tile generation needed -- just the score and score-post.

sverchdotgov:
@esfoobar-usds Assigning this to you because I think you are taking over https://github.com/usds/justice40-tool/pull/1134

PR #1112, opened 2022-01-07
Add github action for testing generate score post

sverchdotgov:
This doesn't take very long, even with the full dataset, so it might be worth running this on every pull request commit.

Related to https://github.com/usds/justice40-tool/issues/599.


sverchdotgov:
Closing in favor of https://github.com/usds/justice40-tool/pull/1134

PR #1113, opened 2022-01-07
More obviously call out deploy time in quickstart

sverchdotgov:
Just to set expectation that this will take a while, so new users don't think they did something wrong.


esfoobar-usds (approved):
🍏

Issue #1114, opened 2022-01-10
Task: Make markdown link checker only run when markdown files change


sverchdotgov:
## Describe the task

Sometimes the markdown link checker fails on things unrelated to the PR (for example, a site was down, which caused our checks to fail, even though the links were correct).

We should only run this when markdown files change, to reduce friction on code changes.

## Acceptance Criteria

- [ ] Markdown link checker only runs when markdown files change.

PR #1117, opened 2022-01-10
Run markdown link checker only on markdown changes

sverchdotgov:
This makes sure that this action only runs if the pull request is on the main branch and is changing markdown files.

Fixes: https://github.com/usds/justice40-tool/issues/1114

### DIFF
--- a/.github/workflows/markdown-link-check.yml
+++ b/.github/workflows/markdown-link-check.yml
@@ -1,7 +1,5 @@
 name: Check Markdown links
 on:
-  push:
-    branches: [ main ]
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ main ]
--- a/.github/workflows/markdown-link-check.yml
+++ b/.github/workflows/markdown-link-check.yml
@@ -1,7 +1,14 @@
 name: Check Markdown links
 on:
   push:
+    branches: [ main ]
   pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+    # Run only if the pull request changes a markdown file:
+    # https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths
+    paths:
+      - '**.md'
 jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/markdown-link-check.yml
```diff
@@ -1,7 +1,14 @@
 name: Check Markdown links
 on:
   push:
+    branches: [ main ]
```
lucasmbrown-usds, line 4: What was your motivation for restricting to `main`? I would think we would want to run this checker *before* merging to main.

sverchdotgov: I copied this part from the codeql checks, which seem to run on every push to a PR against main (see below check).

I suppose this one doesn't matter because main is protected, but it does mean that we won't have github actions run on people's dev branches if they aren't part of a PR yet.

esfoobar-usds: Push to `main` doesn't make sense in our repos since pushing to `main` is blocked. You should just use the `pull_request` action.

sverchdotgov:
@esfoobar-usds Removed the "on.push" block.

.github/workflows/markdown-link-check.yml
```diff
@@ -1,7 +1,12 @@
 name: Check Markdown links
 on:
-  push:
   pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
```
lucasmbrown-usds, line 5: I still confused how we landed here. I thought I suggested that we want to check these *before* merging to main (so we don't introduce errors on main), and @esfoobar-usds suggested a very useful check that runs on pull requests? Did I miss some of the conversation? 

Thanks!

sverchdotgov: @lucasmbrown-usds I believe the configuration as it is now will run whenever someone creates a pull request that is intended to merge into main, so it will catch that case.

This just means if we create a PR against someone else's branch, these won't run.

See https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration

esfoobar-usds (approved):
🍏 approving for now, we'll review if it keeps getting errors

Issue #1121, opened 2022-01-11
Task: Fix node-forge dependabot alert


sverchdotgov:
## Describe the task
We have an alert for node-forge having a security issue. We should make sure this is either resolved or not in production.

## Acceptance Criteria

- [ ] The issue is fixed or confirmed to not affect production and closed in github.

## Additional context

![alert](https://user-images.githubusercontent.com/25348713/148986840-03e1509f-f772-48c9-b591-e485fd200a6b.png)

sverchdotgov:
This is dev only:

```
$ npm audit --prod
found 0 vulnerabilities
$ npm audit | grep node-forge
node-forge  <1.0.0
URL parsing in node-forge could lead to undesired behavior. - https://github.com/advisories/GHSA-gf8q-jrpm-jvxq
node_modules/node-forge
  Depends on vulnerable versions of node-forge
```

sverchdotgov:
Trying to figure out why this config didn't stop that: https://github.com/usds/justice40-tool/blob/main/.github/dependabot.yml

I found this issue: https://github.com/dependabot/dependabot-core/issues/2521

Looks like others are struggling to get dependabot to ignore dev dependencies for security issues.

sverchdotgov:
Made a PR for documenting this issue, I'll close the vulnerability on the repo now.

Note that this workflow should catch real prod issues: https://github.com/usds/justice40-tool/blob/main/.github/workflows/deploy_main.yml#L52

sverchdotgov:
So I think the dependabot config isn't working. See https://github.com/usds/justice40-tool/pull/1122

I think we'll continue to get dependabot alerts for non-production dependency issues, but I'm closing this issue out because it's more about this specific alert.

@esfoobar-usds FYI this was the context behind that pull request.

PR #1122, opened 2022-01-11
Document current dependabot issues in config file

sverchdotgov:
Fixes: https://github.com/usds/justice40-tool/issues/1121

It's unclear whether this file is doing what we want it to do. This is an open issue in the dependabot repo. Added a comment to document the current status of what's going on here.

If these keep coming up (which they probably will), we might either want to disable dependabot or document a quick workflow for verifying this is a dev only issue (https://github.com/usds/justice40-tool/issues/1121 might be that documentation).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,6 @@
+# It's unclear whether this configuration is supported for stopping dev only
+# security issues, see https://github.com/usds/justice40-tool/issues/1121 and
+# https://github.com/dependabot/dependabot-core/issues/2521.
 version: 2
 updates:
   - package-ecosystem: "npm"


esfoobar-usds (changes requested):
This is probably a left over from the previous iteration of the platform. I would just remote this file

sverchdotgov:
@esfoobar-usds Actually I added this file from the dependabot docs: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates

I just think there's an open issue where it doesn't support the options we want right now (see https://github.com/dependabot/dependabot-core/issues/2521).

We could remove it, but it seems like the most logical place to capture that context.

esfoobar-usds:
Got it. I just don't understand the context. Sine the repo is both Python and Javascript, isn't this flagging the repo as JavaScript only?

sverchdotgov:
@esfoobar-usds Actually we now have an example showing that this isn't affecting how dependabot checks python: https://github.com/usds/justice40-tool/security/dependabot/data/data-pipeline/poetry.lock/Pillow/open

So I think either this config doesn't work as the documentation says, or configuring javascript behavior leaves the python behavior alone. Either way, this seems like it won't affect python dependency checks.

esfoobar-usds:
Closing pending further research

PR #1126, opened 2022-01-11
add badge for end to end GH workflow

mchelen-gov:


esfoobar-usds (changes requested):
Thanks for the submission @mchelen-gov. We typically add what Github issue this PR addresses on the description with a text like "Fixes #{issue-number}". Can you please add that?

It would be nice to have badges for other GHA we care about, mainly `data-checks.yml` and `markdown-link-check.yml`.

Issue #1129, opened 2022-01-11
As a comparison tool user, I want to remove `other_census_tract_fields_to_keep` which doesn't do anything


lucasmbrown-usds:
**Description**
`other_census_tract_fields_to_keep` is part of some old code and is no longer used anywhere. This ticket removes it.

saran-ahluwalia:
I think we can close this @lucasmbrown-usds?

lucasmbrown-usds:
Closed in #1130

PR #1130, opened 2022-01-11
Issue 1129: remove deprecated field

lucasmbrown-usds:
#1129. other_census_tract_fields_to_keep is part of some old code and is no longer used anywhere. This PR removes it.


esfoobar-usds (approved):
🍏

Issue #1128, opened 2022-01-11
Error "Poetry could not find a pyproject.toml file" when running `docker-compose up`


mchelen-gov:
**Describe the bug**
Error "Poetry could not find a pyproject.toml file" when running `docker-compose up`

**To Reproduce**
1. Follow steps from [Quickstart](https://github.com/usds/justice40-tool/blob/main/QUICKSTART.md)
2. Run `docker-compose up`

**Expected behavior**
Command is successful

**Screenshots**
```
$ docker-compose up
Creating network "justice40-tool_default" with the default driver
Building score
[+] Building 259.6s (19/20)                                                                                                                                       
 => [internal] load build definition from Dockerfile                                                                                                         0.5s
 => => transferring dockerfile: 1.44kB                                                                                                                       0.0s
 => [internal] load .dockerignore                                                                                                                            0.4s
 => => transferring context: 2B                                                                                                                              0.0s
 => [internal] load metadata for docker.io/library/ubuntu:20.04                                                                                              0.8s
 => [auth] library/ubuntu:pull token for registry-1.docker.io                                                                                                0.0s
 => [ 1/15] FROM docker.io/library/ubuntu:20.04@sha256:b5a61709a9a44284d88fb12e5c48db0409cfad5b69d4ff8224077c57302df9cf                                      2.2s
 => => resolve docker.io/library/ubuntu:20.04@sha256:b5a61709a9a44284d88fb12e5c48db0409cfad5b69d4ff8224077c57302df9cf                                        0.1s
 => => sha256:b5a61709a9a44284d88fb12e5c48db0409cfad5b69d4ff8224077c57302df9cf 1.42kB / 1.42kB                                                               0.0s
 => => sha256:57df66b9fc9ce2947e434b4aa02dbe16f6685e20db0c170917d4a1962a5fe6a9 529B / 529B                                                                   0.0s
 => => sha256:d13c942271d66cb0954c3ba93e143cd253421fe0772b8bed32c4c0077a546d4d 1.46kB / 1.46kB                                                               0.0s
 => => sha256:ea362f368469f909a95f9a6e54ebe0121ce0a8e3c30583dd9c5fb35b14544dec 28.57MB / 28.57MB                                                             0.4s
 => => extracting sha256:ea362f368469f909a95f9a6e54ebe0121ce0a8e3c30583dd9c5fb35b14544dec                                                                    0.7s
 => [internal] load build context                                                                                                                            0.3s
[...]

PR #1131, opened 2022-01-11
Add pyproject.toml to fix docker compose build

sverchdotgov:
Even though we want to use locked dependencies, pyproject.toml is still required.

Fixes: https://github.com/usds/justice40-tool/issues/1128

### DIFF
--- a/data/data-pipeline/Dockerfile
+++ b/data/data-pipeline/Dockerfile
@@ -35,7 +35,7 @@ ENV PYTHONFAULTHANDLER=1 \
   POETRY_VERSION=1.1.12
 
 WORKDIR /data-pipeline
-COPY pyproject.toml poetry.lock /data-pipeline/
+COPY pyproject.toml /data-pipeline/
 
 RUN pip install "poetry==$POETRY_VERSION"
 RUN poetry config virtualenvs.create false \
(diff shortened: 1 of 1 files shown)


data/data-pipeline/Dockerfile
```diff
@@ -35,7 +35,7 @@ ENV PYTHONFAULTHANDLER=1 \
   POETRY_VERSION=1.1.12
 
 WORKDIR /data-pipeline
-COPY poetry.lock /data-pipeline/
+COPY pyproject.toml poetry.lock /data-pipeline/
```
esfoobar-usds, line 38: You don't need the `poetry.lock` here -- it will be generated by the `poetry install` below.

esfoobar-usds (changes requested):

data/data-pipeline/Dockerfile
sverchdotgov: I thought we wanted to install the locked version of the dependencies.

https://python-poetry.org/docs/cli/#install

> If there is a poetry.lock file in the current directory, it will use the exact versions from there instead of resolving them. 

If that's not true we can remove it, I just thought this was the intention.

esfoobar-usds (approved):
🍏

Issue #488, opened 2021-08-09
Create a Github Action for data source ETL so that upstream data issues do not impact score and tile generation.


switzersc-usds:
Updating this issue as a result of #946. We want to mitigate the effect of issues in upstream data sources on our data pipeline. 

Make a Github action that:

- Runs the etl for all the sources 
- Run score, score-post and score-geo

sverchdotgov:
This issue might help with performance of the tilefy job, because `aws s3 rm --recursive` is not using batch operations and is known to be slow, but this might eliminate the need for that step. See https://github.com/usds/justice40-tool/issues/485#issuecomment-1006791510

PR #1134, opened 2022-01-12
Adding github actions to test score and score post

sverchdotgov:
In service of https://github.com/usds/justice40-tool/issues/599

This does not test the etl jobs specific to the datasources. Instead, it takes the "datasets" directory created after all those run and uploads it to s3 in one github action, and downloads it in another. We believe this will take about 8 minutes to run end to end based on https://github.com/usds/justice40-tool/pull/1039#pullrequestreview-830528432.

### DIFF
--- a/.github/workflows/test-generate-score.yml
+++ b/.github/workflows/test-generate-score.yml
@@ -34,7 +34,7 @@ jobs:
           aws-region: us-east-1
       - name: Download Score Test Data From AWS
         run: |
-          aws s3 sync s3://justice40-data/data-pipeline-test-data/data/score/datasets/ ./data_pipeline/data/score/datasets/ --delete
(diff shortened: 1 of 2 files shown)


sverchdotgov:
@vim-usds This is the result of our discussion. Turns out that `application.py` already had targets that did each step separately, so this didn't even require changing the python code.

@esfoobar-usds This is the first pass, and it turns out when you split up the initial download step, the score and score post only take 5 minutes. That's a similar order of magnitude to codeql.

Given that, I actually have another proposal. What if instead of having a separate "upload fixture data to s3" step, we instead split the pipeline to production to have a "pull-upstream-data" step that does the initial etl on all the sources, and a "score and score post" step that downloads from s3. Then the "score and score post" test can just use the results of "pull-upstream-data" directly (as it is here). That would give us part of https://github.com/usds/justice40-tool/issues/488 for free.

PR #1135, opened 2022-01-12
Added pandas numeric conversion example

sverchdotgov:
Trying to find a sane way to convert the type of a column. It seems like creating a new Series object with a "dtype" argument does it and preserves NaNs.

Also added jupyterlab as a dev dependency. The best advice I ever got on pandas is to struggle with it in an ipython notebook until something makes sense. Sure enough, this took like 30 attempts, but the quick turnaround time helped and I think this approach works.


sverchdotgov:
@esfoobar-usds @saran-ahluwalia @vim-usds Does this resolve the original issue? https://github.com/usds/justice40-tool/blob/9849b465b7f3319041f4af3124f5a45475856e75/data/data-pipeline/notebooks/float-to-int-missing-data.ipynb

sverchdotgov:
Note that `pd.NA` is "experimental": https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html#experimental-na-scalar-to-denote-missing-values

I don't think there is otherwise anyway to mark "missing value" in an integer dataframe.

PR #1145, opened 2022-01-14
Fix territory shortcuts on mobile

### LINKED ISSUE #1060: Side panel covers state/territory shortcut buttons on map in mobile
**Describe the bug**
When viewing the map on mobile, you can only see the shortcut for "48", not for any of the states or territories. This makes it difficult to find and navigate to those states/territories.

**To Reproduce**
1. Open the site on a mobile device
2. Go to Explore the Tool
3. Try to select the Alaska shortcut
4. You can't!

**Expected behavior**
You should be able to see and click/tap on shortcuts on map.

**Screenshots**

 - OS: [e.g. iOS] - Android
 - Browser [e.g. chrome, safari] - Firefox

**Additional context**

vim-usds:
- remove zoom controls on mobile
- center search on mobile
- move up the territory shortcuts on mobile
- increase height of map on mobile in portrait mode
- update tests

closes #1060


vim-usds:
What was tried today?

1. console.log windowWidth inside the function
2. write out values of conditional and their types to the DOM
3. Place the conditional on another a random div's style.height prop

All failed on staging.

Next up, place height at 100% detect, control height via parent and control via CSS media query.

vim-usds:
This could be worth a try:
https://github.com/JimmyBeldone/gatsby-plugin-breakpoints

vim-usds:
## Staging strangeness
After adding [remote debug](https://developer.chrome.com/docs/devtools/remote-debugging/), it was determined that the reason mobile on staging wasn't rendering the map is because the height was set to 100%. 

Seems that the main issue on why the the map doesn't show up on staging on mobile has something to do with the conditional on the [height](https://github.com/usds/justice40-tool/blob/2c65215e1a9a8ef6859f25604317232e48406d8c/client/src/components/J40Map.tsx#L288) prop (API [here](https://visgl.github.io/react-map-gl/docs/api-reference/static-map#height)). _**Keeping in mind that this conditional [works on prod](https://github.com/usds/justice40-tool/blob/f4926376faf75faa50cfea46e682e2b131a4d725/client/src/components/J40Map.tsx#L274) for both mobile and desktop.**_

The 4 environments we will be looking at are
1. the staging URL on a mobile device will be referred to as staging-mobile or staging mobile
2. the staging URL on a desktop device will be referred to as staging-desktop or staging desktop
3. the prod URL on a mobile device will be referred to as prod-mobile or prod mobile
4. the prod URL on the desktop device will be referred to as prod-desktop or prod desktop

staging-desktop and prod-desktop seem to work fine. 

### Things I've tried:

#### `height={windowWidth < constants.USWDS_BREAKPOINTS.DESKTOP ? '60vh' : '100%'}`
- console log'd the `windowWidth` (both in the component and onLoad()) and it's shows `320` on staging mobile (using remote debug). 320 < 1024 right?
- height still renders as 100% on staging mobile 🤔 
- Maybe it's the constant? hard code the constant:

#### `height={windowWidth < 1024 ? '44vh' : '100%'}`
- console log'd the windowWidth (same as above)
- height still renders as 100% on staging mobile 🤔 
- Maybe the width isn't enough, maybe it needs device information?

#### `height={(isMobileMapState || (windowWidth < 1024)) ? '44vh' : '100%'}`
- isMobileState is the state variable of isMobile from react-device-detect
- height still renders as 100% on staging mobile 🤔 
- also isMobile doesn't update on width device changes, use windowWidth instead
- Let's swap the order and have '44vh' be the all-else case:

#### `height={windowWidth > 1024 ? '100%' : '44vh'}`
- height still renders as 100% on staging mobile! 🤔 
- Hail Mary :J remove the conditional:

#### `height={'55vh'}`
- map finally shows up! 🥳 
- Ok so something with the conditional, trying to swap the case and use 90% instead:

#### `height={windowWidth > 1024 ? '90%' : '44vh'}`
- height renders 90% on staging mobile 🤔 
- This indicates that even tho the console log of windowWidth is 320, it is still passing the conditional as true.
- How about adding the isMobile as an AND?

#### `height={!isMobile && (windowWidth > 1024) ? '90%' : '44vh'}`
- height renders 90% on staging mobile 🤔 
- console log'd windowWidth = 320, and isMobile = true
- So False && False somehow evaluates to the true case?
- Maybe a function instead of a conditional?
 
 ```
const getMapHeight = () => {
// Todo: change to 100% and move to constants:
    if (isMobile) {
      return isMobile ? ‘55vh’ : ’90%’;
    } else if (windowWidth) {
      return windowWidth > constants.USWDS_BREAKPOINTS.DESKTOP ? ‘90%’ : ‘55vh’;
} else {
      return ‘55vh’;
    }
  };
  ```
- height renders 90% on staging mobile 🤔 
- adding a breakpoint to the function shows that isMobile is true! 🤯 
- when changing the width, the else if clause is triggered and windowWidth = 320, yet it still renders 90%

Screen Shot 2022-01-19 at 8 20 22 PM

- maybe this has to do the height prop of the map component, let's try a regular div

#### Set the height via a conditional on regular 'ol div

```jsx
<div style={{height: `${windowWidth > constants.USWDS_BREAKPOINTS.DESKTOP ? '88.5%' : '55vh'}`}}>
</div>
```
Again even thought the conditional should render to false, the height is set to `88.5`

#### Place all variables in the DOM

```jsx
<div style={{height: `${windowWidth > constants.USWDS_BREAKPOINTS.DESKTOP ? '88.5%' : '55vh'}`}}>
 <pre>
   {`
      windowWidth: ${windowWidth},  \\ 320
      constant.desktop: ${constants.USWDS_BREAKPOINTS.DESKTOP},  \\1024
      boolean test: ${windowWidth > constants.USWDS_BREAKPOINTS.DESKTOP} \\ false
      typeof windowWidth: ${typeof windowWidth}, \\ number
      typeof constant.desktop: ${typeof constants.USWDS_BREAKPOINTS.DESKTOP}, \\ number
   `}
  </pre>
</div>
```

All values seem to evaluate as expected, including the boolean test. However when placing it as a value to the height prop the opposite case is selected (expecting 55vh but it selects 88.5)


#### Open a new PR with no changes
This [PR ](https://github.com/usds/justice40-tool/pull/1150/files)simply adds a comment to the frontend code base.

In this, staging-mobile does not render correctly, the height is still set to 100%.

#### Investigate hydration differences between staging and prod
This still needs more investigation. Since staging and prod both run a production build, the chances of differences seem slim.

#### Infrastructure differences
Production has a CDN that points to an S3 origin, while staging points directly to S3. Could this be a factor? 🤷‍♂️ 

### Conclusions
  
1. When we set the height to a value without a conditional, the height renders correctly on staging-mobile and on prod-mobile.
2. When we set the height via a conditional, the height renders correctly on prod-mobile but not on staging-mobile.
2. When we try and conditionally specify the value based on `windowWidth` or `isMobile` the condition chooses the 'desktop' case on staging-mobile
3. Conditionals and functions behave the same

### Questions

1. Have we tried enough scenarios on stage that it's now OK to merge this PR to main, see if the conditional works on prod (main) and then explore the infrastructure differences between staging and prod?
2. Are there other breakpoints that could be added to troubleshoot this
3. Are there other [selectors](https://github.com/duskload/react-device-detect/blob/master/docs/selectors.md#selectors) that can be tried (other than isMobile) or other [hooks](https://github.com/streamich/react-use) other than `useWindowSize` (which is what is used to get `windowWidth`?

esfoobar-usds (approved):
🍏

PR #1155, opened 2022-01-18
Allow open source map with new API key

vim-usds:
- create new getOSBaseMap function
- set secret on GitHub actions
- read secret into J40Map.tsx
- add secret to gatsby-config
- remove secret from .env.*

closes #1155


esfoobar-usds (approved):
🍏

Issue #1140, opened 2022-01-13
Loss rate fields from FEMA need to be rounded to 6 decimal places in excel download file


BethMattern:
The loss rate fields from NRI have values that are really, really small since they represent the % of __ value at risk from climate disasters in the census tract (e.g., 0.000404786 is an example value for Expected building loss rate (Natural Hazards Risk Index ). if we cut off at 2 decimals, the entire column is full of zeros. 

For the following fields we should round to 6 decimal points in the excel download file:
Expected agricultural loss rate (Natural Hazards Risk Index)
Expected building loss rate (Natural Hazards Risk Index)
Expected population loss rate (Natural Hazards Risk Index)

PR #1170, opened 2022-01-21
Issue 1140 loss rate rounding

emma-nechamkin:
#1140 This is a draft. Some outstanding Qs: (1) is there a better, less case-by-case way to specify rounding procedures for columns in the downloadable data that you guys had been thinking of implementing?, (2) are there downstream impacts from changing the loss rate columns from integers to floats?, and (3) is the loss rate rounding procedure right (rate * 100, rounded to 4 decimal places)?


data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -259,12 +259,23 @@ def _create_downloadable_data(
...
+                    number_of_decimals=constants.TILES_FEMA_ROUND_NUM_DECIMALS,
+                )
+
             else:
                 # Round all other floats.
                 df[column] = floor_series(
                     series=df[column].astype(float64),
                     number_of_decimals=constants.TILES_ROUND_NUM_DECIMALS,
                 )
+                logger.info(df[column].dtype)
```
lucasmbrown-usds, line 278: Did you mean to remove this after using it for sanity-checking?

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -69,12 +69,15 @@
...
     "Energy burden",
-    "loss rate",
     "greater than or equal to 18 years",
     field_names.PERCENTILE_FIELD_SUFFIX,
 ]
-
 TILES_ROUND_NUM_DECIMALS = 2
+
+# FEMA rounding suffix
+FEMA_SUFFIX = "loss rate"
```
lucasmbrown-usds, line 78: I'm torn on this - on one hand it's consistent with how we handle `PERCENT_PREFIXES_SUFFIXES` above and so I appreciate the consistency. On the other it's a little nerve racking to have too much fancy logic based on string matching, for instance someone could add an "Energy loss rate" field and be surprised to see it rounded to 4 digits. So I'm tempted to suggest that we directly list and import the `[field_names.EXPECTED_POPULATION_LOSS_RATE_LOW_INCOME_FIELD, ...]` and check them exactly. 

Hopefully the Big Refactor we're talking about implementing will clean this up, where each field has a schema that defines its own float decimals. 

I could be convinced either way here (to stick with string matching or do exact matches on the field names). What do you think?

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -259,12 +259,23 @@ def _create_downloadable_data(
                     pd.to_numeric(df_100, errors="coerce")
                 ).astype("Int64")
                 df[column] = df_int
+
+            elif constants.FEMA_SUFFIX in column:
+                # Convert loss rates by multiplying by 100 (they are percents)
+                # and then rounding appropriately.
+                df_100 = df[column] * 100
```
lucasmbrown-usds, line 266: Looks great!

lucasmbrown-usds (changes requested):
Overall looks great! Left one big question up to you.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
emma-nechamkin: yes, oof! sorry about that!

data/data-pipeline/data_pipeline/etl/score/constants.py
emma-nechamkin: Hmm... I don't have strong preferences here. I think clearer is better generally,  so I'll update this to list out the columns exactly, and I'll keep the `elif`

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -74,8 +74,12 @@
 ]
 TILES_ROUND_NUM_DECIMALS = 2
 
-# FEMA rounding suffix
-FEMA_SUFFIX = "loss rate"
+# FEMA rounding columns
+FEMA_ROUND_NUM_COLUMNS = [
+    "Expected agricultural loss rate (Natural Hazards Risk Index)",
```
lucasmbrown-usds, line 79: Oh, I'd actually vote we reference these as `field_names.EXPECTED_BUILDING_LOSS_RATE_FIELD` instead of as strings.

lucasmbrown-usds (changes requested):

data/data-pipeline/data_pipeline/etl/score/constants.py
emma-nechamkin: ohhh, yeah that makes more sense. Updated.

lucasmbrown-usds (approved):
LGTM!

esfoobar-usds (approved):
🍏

PR #1182, opened 2022-01-26
Hot fix for Score M

esfoobar-usds:


data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -184,17 +184,9 @@ def _create_score_data(
...
-        ]
-
-        # subtract data sets
-        # this follows the XOR pattern outlined here:
-        # https://stackoverflow.com/a/37313953
-        de_duplicated_df = pd.concat(
-            [merged_df, null_tract_df, null_tract_df]
-        ).drop_duplicates(keep=False)
+        de_duplicated_df = merged_df.dropna(
+            subset=[field_names.SCORE_M_COMMUNITIES]
```
lucasmbrown-usds, line 188: In a PR that @emma-nechamkin and @vim-usds and I were working on, we created a single constant at the top of the file, like `DISADVANTAGED_COMMUNITIES_FIELD = field_names.SCORE_M_COMMUNITIES` because this gets used in a couple places, and it's not obvious where to update it. 

It's also used in this line:

```
 downloadable_df_copy = downloadable_df.rename(
            columns={
                field_names.SCORE_M_COMMUNITIES: "Identified as disadvantaged (v0.1)"
            },
            inplace=False,
        )
```

What do you think? If you don't want to make that change now, we could also make it later.

lucasmbrown-usds (approved):
One suggestion for consideration, otherwise LGTM!

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
esfoobar-usds: Done!

Issue #1180, opened 2022-01-26
Update variable names and definitions used by Definition M to be accurate in backend and frontend


lucasmbrown-usds:
**Description**
When implementing Definition M in https://github.com/usds/justice40-tool/pull/1151, I (Lucas) took a few shortcuts. 

In order to quickly finish Definition M, I continued to use fields like the following: 

```
UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD = (
    f"Greater than or equal to the {PERCENTILE}th percentile for unemployment"
    " and has low HS education"
)
``` 

However, this is inaccurate now, because it should now be something like: 

```
UNEMPLOYMENT_LOW_HS_LOW_HIGHER_ED_FIELD = (
    f"Greater than or equal to the {PERCENTILE}th percentile for unemployment"
    " and has low HS education and low higher ed attendance"
)
``` 

Note the change in both the variable name AND the field string value.

We need to keep the "old" field names/values so they can continue to be used for Score L, but then update these for score M. 

I believe this involves touching all of the following parts of the code: 

Backend:
- [x] `field_names`: add new fields to capture both low_hs AND low_higher_ed
- [x] `score_m`: replace old fields with new fields 
- [x] `DOWNLOADABLE_SCORE_COLUMNS`: replace old fields with new fields
- [x] `TILES_SCORE_COLUMNS`: replace old fields with new fields

Frontend components of this work covered in #1183 and static content covered in #1184.

PR #1185, opened 2022-01-26
(Work In Progress) Updating backend fields to meet Score M definitions

emma-nechamkin:
#1180 updating backend variables to match Score M definitions. Work in progress!


emma-nechamkin:
(I made a git mistake and am going to open a clean pr instead!)

PR #1188, opened 2022-01-27
Update file with branch naming convention

vim-usds:


lucasmbrown-usds (approved):
Love it. Thanks for updating!

Issue #1189, opened 2022-01-27
Make sure workforce development fields for Island areas appear in the UI sidebar


lucasmbrown-usds:
**Description**
Right now, the sidebar panel in the island areas does *not* show any data from the workforce development category even if that data exists for the tract one of those territories. 

This is because that data for the territories is written to a differently named column: eg., in data/data-`pipeline/data_pipeline/score/score_m.py`, the two fields are `field_names.ISLAND_AREAS_UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD` versus `field_names.UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD`. 

`field_names.UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD` is what gets written to the tiles and interpreted by the UI to show up in the sidebar. 

**Solution**
TBD.

**Describe alternatives you've considered**
Possible approaches: 

1. In the backend, create a single field that combines both, e.g., field_names.ISLAND_AREAS_UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD and field_names.UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD into a single field that can be shown by the UI. 

2. In the backend, write both fields to the tiles, and in the frontend, write some logic that uses one of the two fields depending on which one is present (or whether or not you're in the island areas).

emma-nechamkin:
PROPOSED SOLUTION: For the front end to best produce sidebars that vary based on what type of data is being shown -- three source variable will specify whether a file pertains to Islands (USVI, GU, AS, MP), Puerto Rico (PR), or the rest of the US.

lucasmbrown-usds:
A couple other small notes from that discussion: 

1. Put total number of possible thresholds as a column in the tiles data (in US: 21, in PR: 5, in Island Areas: 4)
2. Add island areas workforce development percentile columns to tiles

PR #1213, opened 2022-01-31
Adding island area indicators to the tiles

emma-nechamkin:
Adding thresholds, user experience variable, and island workforce percentiles to tiles. #1189


data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -168,6 +179,7 @@
     field_names.ISLAND_AREAS_LOW_HS_EDUCATION_FIELD: "IALHE",
 }
 
+
```
emma-nechamkin, line 182: Have not yet confirmed all workforce columns are here

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -153,16 +152,14 @@ def _aggregate_to_tracts(
         self, block_group_df: gpd.GeoDataFrame
     ) -> gpd.GeoDataFrame:
         # The tract identifier is the first 11 digits of the GEOID
-        block_group_df["tract"] = block_group_df.apply(
-            lambda row: row[self.GEOID_FIELD_NAME][0:11], axis=1
-        )
+        block_group_df["tract"] = block_group_df[self.GEOID_FIELD_NAME].str[:11]
```
emma-nechamkin, line 155: I think this is marginally faster

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -1,6 +1,7 @@
 import math
 import pandas as pd
 import geopandas as gpd
+import numpy as np
```
emma-nechamkin, line 4: A few notes on this file in places I haven't changed:

1. In line 73, there's a `use_cols` param to the gpd read file call, but I don't think this argument can be used. Can I delete this? 
2. In line 86, a geoid column is renamed in place. I'm wondering if there's a reason? I'm also wondering why the merge can't take place on indexes (like casting the GEOID Tract as an index and then merging with `left_index=True, right_index=True`. I'm just curious!
3. In line 148 (or thereabouts), the rounding procedure takes a series. I'm wondering why it's not just an integer?

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -222,12 +225,38 @@ def _create_tile_data(
             axis=0,
         )
 
+        logger.info("Adding fields for island areas and Puerto Rico")
```
emma-nechamkin, line 228: This is kind of unwieldy, but it should be reasonably quick and clear? Any thoughts here? The other Q I have is this leaves the tiles with two "not short" names -- Region and Thresholds. Am wondering if this is a problem?

emma-nechamkin: There's also a tradeoff here between creating these columns here, once we've narrowed the set of columns in the df, and creating the columns earlier such that these field names can be used in the dictionary. I happen to think (personally) that this should happen as late in the pipeline as possible, but I don't feel super strongly, so overrule me away!

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -74,6 +74,17 @@
 ]
 TILES_ROUND_NUM_DECIMALS = 2
 
+TILES_ISLAND_AREAS = [
+    "American Samoa",
+    "Guam",
+    "Mariana Islands",
+    "US Virgin Islands",
+]
+TILES_ISLAND_FIELD_COUNT = 4
```
lucasmbrown-usds, line 83: Nit: "Field" is a bit vague, maybe call it "Thresholds" to match the language we use in the UI?

lucasmbrown-usds: also feel free to tell me to buzz off at any point with all my OCD nits ;)

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
lucasmbrown-usds: Would love for @esfoobar-usds to weigh in as well!

1. Seems like you're right so you can probably drop it. 

2. Actually as part of `pandas-vet` compliance we'll be moving away from all uses of `inplace=True` (I was guilty of a lot of these, though not specifically this one, before realizing it was frowned upon). So you can make that change. 

2b. Merging on indices would improve speed and it sounds fine to me as long as you reset the index (or otherwise write the output with index=True) to make sure downstream the data stays the same.

3. I'm not super familiar with this code but I *think* what is happening is that it's only rounding the `self.TARGET_SCORE_RENAME_TO` column. Which could also perhaps happen in a more direct / obvious fashion.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
lucasmbrown-usds: Sounds good!

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -153,16 +152,14 @@ def _aggregate_to_tracts(
...
-        )
+        block_group_df["tract"] = block_group_df[self.GEOID_FIELD_NAME].str[:11]
         state_tracts = block_group_df.dissolve(by="tract", aggfunc="mean")
         return state_tracts
 
     def _create_buckets_from_tracts(
         self, state_tracts: gpd.GeoDataFrame, num_buckets: int
     ) -> gpd.GeoDataFrame:
-        # assign tracts to buckets by D_SCORE
+        # assign tracts to buckets by SCORE
```
lucasmbrown-usds, line 162: nice catch

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -153,16 +152,14 @@ def _aggregate_to_tracts(
...
         state_tracts.sort_values(self.TARGET_SCORE_RENAME_TO, inplace=True)
         SCORE_bucket = []
```
lucasmbrown-usds, line 164: Can we invoke the campsite rule (leave it better than you found it) and fix the capitalization on this? I'd expect it to be `score_bucket`.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -222,12 +225,38 @@ def _create_tile_data(
...
+        score_tiles[field_names.REGION_FIELD] = np.where(
+            score_tiles[field_names.STATE_FIELD].isin(
```
lucasmbrown-usds, line 230: I get nervous about matching on strings -- could we use the FIPS codes instead? Or wdyt? I suppose the downside with matching on FIPS codes is that they're less human readable to find errors, but at the same time, they're more constant and predictable than the string names (which could change capitalization, punctuation, etc).

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -222,12 +225,38 @@ def _create_tile_data(
...
+                constants.TILES_PUERTO_RICO_AREAS
+            ),
+            "PR",
```
lucasmbrown-usds, line 233: nit: i'm generally averse to acronyms, i feel like the time saved typing is always lost in the time spent trying to read and decipher them. Would you be comfortable expanding these to full names (in whatever case you think is appropriate)? 

second nitty nit: I am unnaturally terrified of strings being used multiple times in code, due to the chance of fat-fingering mistakes when trying to reference the same field by a string name. When using a constant for field names instead, the compiler will discover that the field name is misspecified before the code ever gets committed. I know I'm a little more paranoid than is justified, though. Are you comfortable making constants for these and reusing them?

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -8,6 +8,8 @@
 GEOID_TRACT_FIELD = "GEOID10_TRACT"
 STATE_FIELD = "State/Territory"
 COUNTY_FIELD = "County Name"
+REGION_FIELD = "Region"
```
lucasmbrown-usds, line 11: Nit: Region actually has a specific census meaning that we use in the comparison tool (https://www2.census.gov/geo/pdfs/maps-data/maps/reference/us_regdiv.pdf), is there another name for this? Nothing obvious comes to mind for me. Could even make it very UI-specific (since that's what this is) and call it something like USER_INTERFACE_EXPERIENCE (I don't love that, but you get the idea).

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
lucasmbrown-usds: Discussion moved to slack on the second comment.

lucasmbrown-usds (commented):
Overall looking good! Nice work diving into this!

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
emma-nechamkin: yes to both of these

data/data-pipeline/data_pipeline/score/field_names.py
emma-nechamkin: yes! this makes sense

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
emma-nechamkin: 1 and 2, 2b sound good. For 3, I meant something along the lines of df[column].round(2), which should accomplish the same thing (I believe!) as df[column].round(pd.Series([2,2,2,2,2,2...]) and would be easier to read?

data/data-pipeline/data_pipeline/etl/score/constants.py
emma-nechamkin: OCD nits are generally helpful imo (with the idea that hopefully each PR has fewer stylistic issues), so this sounds good

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
emma-nechamkin: this sounds good, FIPS code works

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -74,6 +74,25 @@
 ]
 TILES_ROUND_NUM_DECIMALS = 2
 
+# Controlling Tile user experience columns
```
emma-nechamkin, line 77: I made the choice to move all of this into constants, rather than field names. I think this makes more sense because none of this information is used in scoring and I thought it was clearer this way. Happy to revert if you think the field names belong better with other field names :)

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -74,6 +74,25 @@
...
+# Note that the FIPS code is a string
+# The FIPS codes listed are:
+# 60: American Samoa, 66: Guam, 69: N. Mariana Islands, 78: US Virgin Islands
+TILES_ISLAND_AREA_FIPS_CODES = ["60", "66", "69", "78"]
+TILES_PUERTO_RICO_FIPS_CODE = ["72"]
+
+# Constant to reflect UI Experience version
+# "Nation" referring to 50 states and DC is from Census
+USER_INTERFACE_EXPERIENCE_FIELD_NAME = "UI Experience"
+NATION_USER_EXPERIENCE = "Nation"
```
emma-nechamkin, line 92: This is to reflect the language Lucas got from Census. Let me know if you think it is clear :)

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -291,6 +325,12 @@
     field_names.HIGH_SCHOOL_ED_FIELD,
     field_names.COMBINED_UNEMPLOYMENT_2010,
     field_names.COMBINED_POVERTY_LESS_THAN_100_FPL_FIELD_2010,
+    # field_names.LOW_CENSUS_DECENNIAL_AREA_MEDIAN_INCOME_PERCENT_FIELD_2009
```
emma-nechamkin, line 328: To delete -- this is actually in combined fields.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -83,26 +82,33 @@ def extract(self) -> None:
 
     def transform(self) -> None:
         # Rename GEOID10_TRACT to GEOID10 on score to allow merging with Census GeoJSON
-        self.score_usa_df.rename(
-            columns={self.TRACT_SHORT_FIELD: self.GEOID_FIELD_NAME},
-            inplace=True,
-        )
+        # self.score_usa_df.rename(
```
emma-nechamkin, line 85: temporarily changing this to make it a little faster, happy to revert.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -70,7 +70,6 @@ def extract(self) -> None:
         self.geojson_usa_df = gpd.read_file(
             self.CENSUS_USA_GEOJSON,
             dtype={self.GEOID_FIELD_NAME: "string"},
-            usecols=[self.GEOID_FIELD_NAME, self.GEOMETRY_FIELD_NAME],
```
emma-nechamkin, line 73: I removed this line because it doesn't do anything.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
emma-nechamkin: Okay so I thought a little more about this and (for now). 

Yesterday, you mentioned that your instinct was to have this upstream, for the following reasons:
__pros:__
- makes this data available in the “all the things” file, keeping the spirt of that file having “all the things”
- researchers using the “big” file would be able to run analyses using the total number of thresholds
- we might want the total # of thresholds in the download xslx eventually

__cons:__
- might be a confusing field to see for people looking at the file
- weird one-off fixes should generally be targeted at specific areas of the codebase, not perpetuated through the whole thing

Then we talked a little bit about splitting this up -- having a threshold count available upstream and the rest down by the tiles. The issue there is the logic gets split up and it could be more confusing. 

I'm leaning towards keeping this where it is. I think deeming this a UI-only change makes sense because all of the information contained in this step is contained in other columns. What I mean is that as a data scientist, I would prefer documentation (even an additional table) that explains how the thresholds were computed, and NaNs in the actual file with the thresholds when the data for that threshold is missing. This could just be weird quirks about me, though -- but I think researchers who wanted to do analysis that included number of thresholds would already have that information. 

This is just my gut, though --  ultimately it's your call

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -83,26 +82,29 @@ def extract(self) -> None:
...
+        # )
 
         logger.info("Pruning Census GeoJSON")
         fields = [self.GEOID_FIELD_NAME, self.GEOMETRY_FIELD_NAME]
         self.geojson_usa_df = self.geojson_usa_df[fields]
-
         logger.info("Merging and compressing score CSV with USA GeoJSON")
-        self.geojson_score_usa_high = self.score_usa_df.merge(
-            self.geojson_usa_df, on=self.GEOID_FIELD_NAME, how="left"
+        self.geojson_score_usa_high = (
```
emma-nechamkin, line 94: I checked only that the shape of the df was consistent, I didn't check the merge. If we decide to go forward with this I will check it in more detail.

data/data-pipeline/data_pipeline/score/score_m.py
```diff
@@ -148,7 +148,6 @@ def _climate_factor(self) -> bool:
             field_names.EXPECTED_POPULATION_LOSS_RATE_LOW_INCOME_FIELD,
             field_names.EXPECTED_AGRICULTURE_LOSS_RATE_LOW_INCOME_FIELD,
             field_names.EXPECTED_BUILDING_LOSS_RATE_LOW_INCOME_FIELD,
-            # field_names.EXTREME_HEAT_MEDIAN_HOUSE_VALUE_LOW_INCOME_FIELD,
```
emma-nechamkin, line 151: I got rid of this for clean up

emma-nechamkin:
As a general note -- I haven't yet updated the testing fixtures and will do so once we are aligned on where everything goes and what everything is named, so that I only have to do it once. Let me know if this is an issue!

emma-nechamkin:
And one other super nitpicky and annoying thought, feel free to disregard: 

I think this bit of code in `etl_score` could be sped up a little with `nunique`:

__ORIGINAL__
```
        non_unique_tract_values = len(
            df_to_check[self.GEOID_TRACT_FIELD_NAME]
        ) - len(df_to_check[self.GEOID_TRACT_FIELD_NAME].unique())

        if non_unique_tract_values > 0:
            raise ValueError(
                f"There are {non_unique_tract_values} duplicate tract IDs in {dataframe_descriptor}"
            )
```
__PROPOSED__
```
(
assert df_to_check[self.GEOID_TRACT_FIELD_NAME].nunique() == df_to_check[self.GEOID_TRACT_FIELD_NAME].shape[0], 
 f"There are {non_unique_tract_values} duplicate tract IDs in {dataframe_descriptor}"
)
```

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
lucasmbrown-usds: Oh I see. If that's how that works (I haven't checked documentation), that sounds more readable to me so I'd endorse it!

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
lucasmbrown-usds: I endorse your proposal, your reasoning makes a lot of sense to me! (And though I have longer tenure, my vote counts the same as yours 😄 .) 

Let's do that for now. We can always refactor later if we learn more and change our minds.

data/data-pipeline/data_pipeline/etl/score/constants.py
lucasmbrown-usds: I think having the constants (like the threshold count, fips codes, etc) here makes a lot of sense. I'm torn on moving the field names here. That feels somewhat unintuitive since all the other field names are there. But I think it's fine for now? We're clearly planning some big moves around Fields and Models soon so can reconsider all this then.

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -166,6 +185,13 @@
     field_names.ISLAND_AREAS_POVERTY_LOW_HS_EDUCATION_FIELD: "ISPLHSE",
     field_names.ISLAND_AREAS_LOW_MEDIAN_INCOME_LOW_HS_EDUCATION_FIELD: "IALMILHSE",
     field_names.ISLAND_AREAS_LOW_HS_EDUCATION_FIELD: "IALHE",
+    # Percentiles for Island areas' workforce columns
+    field_names.LOW_CENSUS_DECENNIAL_AREA_MEDIAN_INCOME_PERCENT_FIELD_2009
+    + field_names.PERCENTILE_FIELD_SUFFIX: "IAULHSE_PFS",
+    field_names.CENSUS_DECENNIAL_POVERTY_LESS_THAN_100_FPL_FIELD_2009
+    + field_names.PERCENTILE_FIELD_SUFFIX: "ISPLHSE_PFS",
+    field_names.CENSUS_DECENNIAL_UNEMPLOYMENT_FIELD_2009
+    + field_names.PERCENTILE_FIELD_SUFFIX: "IALMILHSE_PFS",
```
lucasmbrown-usds, line 194: lol these short names are getting to be a mouthful

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
lucasmbrown-usds: Oh I see. This is no longer necessary because you join directly on the indices below. Makes sense to me, I think you can go ahead and delete this line then.

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -74,6 +74,25 @@
...
+TILES_PUERTO_RICO_THRESHOLD_COUNT = 5
+TILES_NATION_THRESHOLD_COUNT = 21
+
+# Note that the FIPS code is a string
+# The FIPS codes listed are:
+# 60: American Samoa, 66: Guam, 69: N. Mariana Islands, 78: US Virgin Islands
+TILES_ISLAND_AREA_FIPS_CODES = ["60", "66", "69", "78"]
+TILES_PUERTO_RICO_FIPS_CODE = ["72"]
+
+# Constant to reflect UI Experience version
```
lucasmbrown-usds, line 89: nit: maybe a slightly more elaborate comment so that a developer unfamiliar with the codebase can understand? E.g., the comment you have, but also something like "This variable will be used by the front-end to show a different side panel. The different side panels are necessary because the nation, Puerto Rico, and island areas all have different fields to display, respectively." 

Something like that.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -222,6 +224,27 @@ def _create_tile_data(
             axis=0,
         )
 
+        logger.info("Adding fields for island areas and Puerto Rico")
+        fips_code_series = score_tiles[field_names.GEOID_TRACT_FIELD].str[:2]
+        score_tiles[constants.USER_INTERFACE_EXPERIENCE_FIELD_NAME] = np.where(
```
lucasmbrown-usds, line 229: Could you maybe add some more comments explaining why we're doing all this? Thanks. If I were a new developer in the codebase, I'd be pretty confused.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -222,6 +224,27 @@ def _create_tile_data(
...
+                constants.NATION_USER_EXPERIENCE,
+            ),
+        )
+        score_tiles[constants.THRESHOLD_COUNT_TO_SHOW_FIELD_NAME] = score_tiles[
+            constants.USER_INTERFACE_EXPERIENCE_FIELD_NAME
+        ].map(
+            {
+                constants.PUERTO_RICO_USER_EXPERIENCE: constants.TILES_PUERTO_RICO_THRESHOLD_COUNT,
+                constants.ISLAND_AREAS_USER_EXPERIENCE: constants.TILES_ISLAND_AREAS_THRESHOLD_COUNT,
+                constants.NATION_USER_EXPERIENCE: constants.TILES_NATION_THRESHOLD_COUNT,
```
lucasmbrown-usds, line 244: we love to see a nice clean set of constants

lucasmbrown-usds (commented):
Overall looking great! Nice work. Sorry about the bizarre island areas percentiles issue we discussed on Zoom. Thanks for cleaning up my campsites :)

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
esfoobar-usds: I'm good with all your suggestions @emma-nechamkin!

data/data-pipeline/data_pipeline/etl/score/constants.py
emma-nechamkin: They are consistent with the others, which I think is maybe a good thing, but yah they aren't so short anymore!

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -199,6 +231,16 @@
...
     field_names.UNEMPLOYMENT_FIELD + field_names.PERCENTILE_FIELD_SUFFIX,
+    # Percentiles for Island areas' workforce columns
+    # To be clear: the island areas pull from 2009 census. PR does not.
+    field_names.LOW_CENSUS_DECENNIAL_AREA_MEDIAN_INCOME_PERCENT_FIELD_2009
+    + field_names.PERCENTILE_FIELD_SUFFIX,
+    field_names.CENSUS_DECENNIAL_POVERTY_LESS_THAN_100_FPL_FIELD_2009
+    + field_names.ISLAND_AREAS_PERCENTILE_ADJUSTMENT_FIELD
+    + field_names.PERCENTILE_FIELD_SUFFIX,
+    field_names.CENSUS_DECENNIAL_UNEMPLOYMENT_FIELD_2009
+    + field_names.ISLAND_AREAS_PERCENTILE_ADJUSTMENT_FIELD
```
emma-nechamkin, line 242: This is an additional field to create a column that is just percentiles based on the 2009 data for the Island areas. It's not the cleanest solution generally speaking, but I can't think of any "good way"?

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -82,27 +81,24 @@ def extract(self) -> None:
...
-
         logger.info("Pruning Census GeoJSON")
         fields = [self.GEOID_FIELD_NAME, self.GEOMETRY_FIELD_NAME]
         self.geojson_usa_df = self.geojson_usa_df[fields]
-
         logger.info("Merging and compressing score CSV with USA GeoJSON")
-        self.geojson_score_usa_high = self.score_usa_df.merge(
-            self.geojson_usa_df, on=self.GEOID_FIELD_NAME, how="left"
+        self.geojson_score_usa_high = (
+            self.score_usa_df.set_index(self.TRACT_SHORT_FIELD)
```
emma-nechamkin, line 89: Still have to confirm this join performs as expected

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -146,31 +142,30 @@ def transform(self) -> None:
         )
 
         # round to 2 decimals
-        decimals = pd.Series([2], index=[self.TARGET_SCORE_RENAME_TO])
-        self.geojson_score_usa_low = self.geojson_score_usa_low.round(decimals)
+        self.geojson_score_usa_low = self.geojson_score_usa_low.round(
```
emma-nechamkin, line 145: Still have to confirm this works as intended.

data/data-pipeline/data_pipeline/score/score_m.py
```diff
@@ -56,6 +56,22 @@ def _combine_island_areas_with_states_and_set_thresholds(
             [column_from_island_areas, column_from_decennial_census]
         ].mean(axis=1, skipna=True)
 
+        # Create a percentile field for use in the Islands / PR visualization
+        # TODO: move this code
+        # In the code below, percentiles are constructed based on the combined column
+        # of census and island data, but only reported for the island areas (where there
+        # is no other comprehensive percentile information)
+        return_series_name = (
```
emma-nechamkin, line 64: So a true nit comment -- I think if we have to do this, it makes sense to collapse this part of the code:

```
        # Calculate the percentile threshold raw value.
        raw_threshold = np.nanquantile(
            a=df[combined_column_name], q=threshold_cutoff_for_island_areas
        )

        logger.info(
            f"For combined field `{combined_column_name}`, "
            f"the {threshold_cutoff_for_island_areas*100:.0f} percentile cutoff is a "
            f"raw value of {raw_threshold:.3f}."
        )

        threshold_column_name = (
            f"{column_from_island_areas} exceeds "
            f"{threshold_cutoff_for_island_areas*100:.0f}th percentile"
        )

        df[threshold_column_name] = (
            df[column_from_island_areas] >= raw_threshold
        )
```

From here, I *think* we could simply do: 
```
        threshold_column_name = (
            f"{column_from_island_areas} exceeds "
            f"{threshold_cutoff_for_island_areas*100:.0f}th percentile"
        )

        df[threshold_column_name] = (
            df[return_series_name] >= threshold_cutoff_for_island_areas
        )
```

And we would not have to:
(1) calculate the threshold directly
(2) use the threshold to calculate which fields get included
(3) only treat the island areas -- the default here for NaN is `false` I think (`NaN` equals nothing, I'm pretty sure)

I'm not sure this actually matters,  though, especially given the refactor on the horizon. But, just a thought :)

lucasmbrown-usds: Switching to just using the 2nd code block in your comment makes a lot of sense me. I assume this code would come after you set `return_series_name`, right? 

Btw, this is not the best process, but in our absence of tests (sigh) you can compare the before and after score files using the notebook `data/data-pipeline/data_pipeline/ipython/compare_two_score_files_for_differences.ipynb`. Just give it two paths to different score files and it'll give you a pretty good readout on whether there are any differences.

data/data-pipeline/data_pipeline/score/score_m.py
```diff
@@ -623,6 +638,8 @@ def _workforce_factor(self) -> bool:
         ]
 
         # First, combine unemployment.
+        # This will include an adjusted percentile column for the island areas
```
lucasmbrown-usds, line 641: Nice, very helpful.

data/data-pipeline/data_pipeline/score/score_m.py
lucasmbrown-usds: If you don't want to re-run `main` to get an old score file, you can get a `usa.csv` output from `"https://justice40-data.s3.amazonaws.com/data-pipeline"`, which is defined in `data/data-pipeline/settings.toml` and follows the same directory structure as your score outputs locally.

lucasmbrown-usds (approved):
LGTM. Based on your comments, it sounds like there's enough slightly nervewracking changes here that you may want to run the two score files comparison I suggest before we merge!

data/data-pipeline/data_pipeline/score/score_m.py
emma-nechamkin: ty ty! will do

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -294,4 +336,12 @@
     field_names.ISLAND_AREAS_UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD,
     field_names.ISLAND_AREAS_POVERTY_LOW_HS_EDUCATION_FIELD,
     field_names.ISLAND_AREAS_LOW_MEDIAN_INCOME_LOW_HS_EDUCATION_FIELD,
+    field_names.LOW_CENSUS_DECENNIAL_AREA_MEDIAN_INCOME_PERCENT_FIELD_2009
```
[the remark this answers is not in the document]
emma-nechamkin: Oof, disregard this comment -- I was comparing apples to oranges. (The booleans are ALL full across the datasets 😵‍💫 and the percentiles are all not)

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -163,9 +186,20 @@
     field_names.FPL_200_AND_COLLEGE_ATTENDANCE_SERIES: "FPL200S",
     field_names.THRESHOLD_COUNT: "TC",
     field_names.ISLAND_AREAS_UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD: "IAULHSE",
-    field_names.ISLAND_AREAS_POVERTY_LOW_HS_EDUCATION_FIELD: "ISPLHSE",
+    field_names.ISLAND_AREAS_POVERTY_LOW_HS_EDUCATION_FIELD: "IAPLHSE",
```
emma-nechamkin, line 189: I think this tile short name actually had a typo

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -163,9 +186,20 @@
...
+    field_names.LOW_CENSUS_DECENNIAL_AREA_MEDIAN_INCOME_PERCENT_FIELD_2009
+    + field_names.PERCENTILE_FIELD_SUFFIX: "IALMILHSE_PFS",
+    field_names.CENSUS_DECENNIAL_POVERTY_LESS_THAN_100_FPL_FIELD_2009
+    + field_names.ISLAND_AREAS_PERCENTILE_ADJUSTMENT_FIELD
+    + field_names.PERCENTILE_FIELD_SUFFIX: "IAPLHSE_PFS",
+    field_names.CENSUS_DECENNIAL_UNEMPLOYMENT_FIELD_2009
+    + field_names.ISLAND_AREAS_PERCENTILE_ADJUSTMENT_FIELD
+    + field_names.PERCENTILE_FIELD_SUFFIX: "IAULHSE_PFS",
+    # Percentage of HS Degree completion for Islands
+    field_names.CENSUS_DECENNIAL_HIGH_SCHOOL_ED_FIELD_2009: "IAHSEF",
```
emma-nechamkin, line 202: I added this -- I don't think I was passing the educational attainment rate for islands before and it seems like the front end needs that information too!

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -199,6 +233,18 @@
...
+    field_names.LOW_CENSUS_DECENNIAL_AREA_MEDIAN_INCOME_PERCENT_FIELD_2009
+    + field_names.PERCENTILE_FIELD_SUFFIX,
+    field_names.CENSUS_DECENNIAL_POVERTY_LESS_THAN_100_FPL_FIELD_2009
+    + field_names.ISLAND_AREAS_PERCENTILE_ADJUSTMENT_FIELD
+    + field_names.PERCENTILE_FIELD_SUFFIX,
+    field_names.CENSUS_DECENNIAL_UNEMPLOYMENT_FIELD_2009
+    + field_names.ISLAND_AREAS_PERCENTILE_ADJUSTMENT_FIELD
+    + field_names.PERCENTILE_FIELD_SUFFIX,
+    # Island areas HS degree attainment rate
+    field_names.CENSUS_DECENNIAL_HIGH_SCHOOL_ED_FIELD_2009,
```
emma-nechamkin, line 247: added

emma-nechamkin:
I was getting an error in `etl_score_geo` so I rolled back all changes I added to that file... I think I am still getting an error though?

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -81,40 +82,27 @@ def extract(self) -> None:
...
-        logger.info(
-            f"{geojson_score_usa_high.equals(self.geojson_score_usa_high)}"
-        )
-        assert 1 == 2
+
         self.geojson_score_usa_high = gpd.GeoDataFrame(
             self.geojson_score_usa_high, crs="EPSG:4326"
         )
 
+        logger.info(f"Columns: {self.geojson_score_usa_high.columns}")
```
lucasmbrown-usds, line 104: intentional to keep? i think it could be helpful, maybe just a more descriptive comment.

emma-nechamkin: yikes oops,  def not intentional (I pulled this down from main, see comment below, will remove this line :))

data/data-pipeline/data_pipeline/etl/score/constants.py
lucasmbrown-usds: If we're updating it, let's make sure to flag it for @vim-usds on the front-end dictionary.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
emma-nechamkin: (fwiw i would like to merge this into main and then separately update this code, because the tests take forever to pass? @lucasmbrown-usds)

lucasmbrown-usds: That's fine!

PR #1234, opened 2022-02-03
Update boolean for HS degree indicator to LHE

### LINKED ISSUE #1148: High school degree achievement rate is not blue in the side panel when it should be
The high school degree achievement rate should be blue if it is between the values of 0 and 89. 

Example tract where it should be blue and isn't:
https://d29zfl8cj7y1zf.cloudfront.net/en/cejst#14.65/24.55716/-81.78308

vim-usds:
# Purpose

Update the boolean value that determines if the HS degree achievement indicator on the side panel is blue or not

closes #1148


esfoobar-usds:
@emma-nechamkin I'd love this field to be a boolean instead of "0" or "1" as other fields.

KameronKerger:
can confirm. is blue.

PR #1235, opened 2022-02-03
Update FE to Def M

### LINKED ISSUE #1183: Front end updates to variables used by definition M
**Description**

When implementing Definition M, I (Lucas) took a few shortcuts.

In order to quickly finish Definition M, I continued to use fields like the following:

```
UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD = (
    f"Greater than or equal to the {PERCENTILE}th percentile for unemployment"
    " and has low HS education"
)
```

However, this is inaccurate now, because it should now be something like:

```
UNEMPLOYMENT_LOW_HS_LOW_HIGHER_ED_FIELD = (
    f"Greater than or equal to the {PERCENTILE}th percentile for unemployment"
    " and has low HS education and low higher ed attendance"
)
```

Note the change in both the variable name AND the field string value.

We need to keep the "old" field names/values so they can continue to be used for Score L, but then update these for score M.

I believe this involves touching all of the following parts of the code on the front end:

- [ ] Update text descriptions of data on methodology page
- [ ] Update the reverse dictionary translating short names on tiles back to full names (maybe not necessary if we keep the abbreviations in TILES_SCORE_COLUMNS the same)
- [ ] Update full names of variables in side bar
- [ ] Update text descriptions of variables in side bar

vim-usds:
# Purpose
closes #1183 

## Make copy updates

- update first column of meth page, formula intro, ID,AND, & remove THEN
Screen Shot 2022-02-02 at 8 42 44 PM

- remove Low income box under download box
- update `identify as disadv.` in each category
- modify AND clause in each category
Screen Shot 2022-02-02 at 8 44 55 PM

- add dataset Higher ed enrollment
Screen Shot 2022-02-02 at 8 48 23 PM

## 4 side panel description updates
Screen Shot 2022-02-02 at 9 16 53 PM
Screen Shot 2022-02-02 at 9 17 11 PM
Screen Shot 2022-02-02 at 9 17 37 PM
Screen Shot 2022-02-02 at 9 17 55 PM

## No new backend fields were updated, were they suppose to be?
I may have misunderstood, however at first I believed the backed was a blocker on this. All changes so far have only been copy changes. Was I to update some new backend fields as well?


KameronKerger:
looks good @vim-usds! there is one **my bad typo** in the AND sentence in the methodology formula. there should be an "of" after the word set and before the word income. I've fixed living copy now.
Screen Shot 2022-02-03 at 7 49 35 AM

KameronKerger:
typo good to go!

esfoobar-usds (approved):
🍏

Issue #1003, opened 2021-12-07
Remove USVI and Guam territories from data and include/show on map American Samoa and Mariana Islands


BethMattern:
There are issues with the underlying data for USVI and Guam, so we do not want to include these two island territories in the data / on the map. 

[UPDATE January 11]
- We want score generation to run for all territories and written to the full `usa.csv` per @lucasmbrown-usds 
- We want to strip these territories right before tile generation, probably somewhere on score_post

vim-usds:
@jorge and @shelby making this a data ticket as there's another one for the FE work

lucasmbrown-usds:
@esfoobar-usds Would be helpful to only remove these at the "last minute" before we write the downloadable Excel file and the tiles.

esfoobar-usds:
@lucasmbrown-usds what about the download packet? Should all territories be written on the CSV/Excel, or exclude these from there as well?

lucasmbrown-usds:
All territories should be written to the huge `usa.csv` but only Virgin Islands, American Samoa, and Puerto Rico to the downloadable Excel file/CSV or to the tiles.

lucasmbrown-usds:
From Beth: the problem is with Guam (is on the map, shouldn't be) and American Samoa (is not on the map, should be).

vim-usds:
- updated local fips to latest fips
- re-running full pipeline...

- [ ] does full csv have all territory data, make a note of ones that are missing
- [ ] does tile csv have the ones we want (meaning, not GU, but yes to AS)
- [ ] is the SP_FPS value accurate in the ones we want?

vim-usds:
@jorge, @lucas and @beth,

Here are the codes for each of the territories:

## Territory codes from FIPS file from S3:
```
60, American Samoa,AS,American Samoa,American Samoa
66, Guam,GU,Guam,Guam
69, Northern Mariana Islands,MP,Northern Mariana Islands,Northern Mariana Islands
72, Puerto Rico,PR,Puerto Rico,Puerto Rico
78, Virgin Islands,VI,Virgin Islands,Virgin Islands
```

## CSV Files
The ✅  indicates if the data is in the CSV file
The ❌ indicates if the data is NOT in the CSV file

### CSV full
60, American Samoa,AS,American Samoa,American Samoa ✅ 
66, Guam,GU,Guam,Guam ✅ 
69, Northern Mariana Islands,MP,Northern Mariana Islands,Northern Mariana Islands ✅ 
[...]

PR #1248, opened 2022-02-09
Remove USVI and Guam territories from data and include/show on map American Samoa and Mariana Islands

emma-nechamkin:
#1003

Not the most elegant solution, but I get the sense this will be relatively temporary? 
Screen Shot 2022-02-08 at 7 25 04 PM

Only USVI and Guam FIPS codes (66, 78) have been removed.

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/score/constants.py
+++ b/data/data-pipeline/data_pipeline/etl/score/constants.py
@@ -63,6 +63,10 @@ SCORE_DOWNLOADABLE_ZIP_FILE_PATH = (
 # Column subsets
 CENSUS_COUNTIES_COLUMNS = ["USPS", "GEOID", "NAME"]
 
+# Drop USVI and Guam flag for tiles
+DROP_GUAM_AND_USVI_FROM_TILES = True
+DROP_FIPS_CODES = ["66", "78"]
+
 # Percent prefixes for rounding
 PERCENT_PREFIXES_SUFFIXES = [
     "Percent",
--- a/data/data-pipeline/data_pipeline/etl/score/constants.py
+++ b/data/data-pipeline/data_pipeline/etl/score/constants.py
@@ -63,8 +63,7 @@ SCORE_DOWNLOADABLE_ZIP_FILE_PATH = (
 # Column subsets
 CENSUS_COUNTIES_COLUMNS = ["USPS", "GEOID", "NAME"]
 
-# Drop USVI and Guam flag for tiles
-DROP_GUAM_AND_USVI_FROM_TILES = True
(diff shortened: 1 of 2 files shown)


data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -63,6 +63,10 @@
 # Column subsets
 CENSUS_COUNTIES_COLUMNS = ["USPS", "GEOID", "NAME"]
 
+# Drop USVI and Guam flag for tiles
+DROP_GUAM_AND_USVI_FROM_TILES = True
```
esfoobar-usds, line 67: I'm not sure we need this bool. I like and think that `DROP_FIPS_CODE` is enough to know we're dropping some FIPS. Then on `score_post`, instead of checking the bool, you can check the length of the list and if it's greater than 0, go into the if block.

esfoobar-usds (approved):
A small nit, let me know what you think @emma-nechamkin 

🍏

data/data-pipeline/data_pipeline/etl/score/constants.py
lucasmbrown-usds: That's also a good idea -- I could go either way.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -207,6 +209,22 @@ def _create_tile_data(
             tiles_score_column_titles
         ].copy()
 
+        if drop_guam_usvi:
+            # Currently, we do not want USVI or Guam on the map, so this will drop all
+            # rows with the FIPS codes (first two digits of the census tract)
+            logger.info("Dropping USVI and Guam from tile data")
+            tracts_to_drop = []
+            for fips_code in constants.DROP_FIPS_CODES:
```
lucasmbrown-usds, line 217: If we go with Jorge's suggestion, if `constants.DROP_FIPS_CODES` is empty then this will do nothing, so don't even really need an additional check.

lucasmbrown-usds (approved):
LGTM!

data/data-pipeline/data_pipeline/etl/score/constants.py
emma-nechamkin: I like that!

esfoobar-usds:
Regarding work in progress: I really don't like using the title of the PR for it. Github already has the concept of a "draft", which you can mark your PR as a draft, which you can find under the reviewers tab. Once the PR is ready you can mark as ready for review on the same link:

image

/cc @emma-nechamkin @lucasmbrown-usds @vim-usds

Issue #1038, opened 2021-12-10
As a data developer I want to run ETL processes in parallel so that I can complete jobs more quickly


saran-ahluwalia:
(Updated 2/8:)

All ETL classes can be run independently; none have any dependencies on each other. When we run application.py etl-run with no dataset specified, it downloads every dataset in DATASET_LIST from data/data-pipeline/data_pipeline/etl/constants.py.

We could add a very simple parallelization to etl-run to run all the items in DATASET_LIST in parallel. Most of the time is spent on I/O for these ETLs, waiting for the data to download and then writing the results to disk.

This would be fairly straightforward to setup a parallelization pool to run them independently. However, right now, each ETL uses the same `TMP` directory and then deletes all the files in the `TMP` directory at the end of the run. Run concurrently, this means sometimes deleting data needed by one of the other ETLs. To get around this, it would be easy to namespace the `TMP` directories using either the name of the class or a hash or something.

lucasmbrown-usds:
Thanks for creating this. It's close but not quite. 

Some more details: 

All ETL classes can be run independently; none have any dependencies on each other. When we run `application.py etl-run` with no dataset specified, it downloads every dataset in `DATASET_LIST` from `data/data-pipeline/data_pipeline/etl/constants.py`. 

We could add a very simple parallelization to `etl-run` to run all the items in `DATASET_LIST` in parallel. Most of the time is spent on I/O for these ETLs, waiting for the data to download and then writing the results to disk. 

Score must be run after all of the ETLs are complete. Right now this process generates several scores (Score A, Score B, ... Score L) and these could potentially be run in parallel but this would take a little extra effort to recombine the results (since they're all currently written to self.df). We would need to double-check that none of the scores depend on a previous score, too.

saran-ahluwalia:
@lucasmbrown-usds Understood, I think I have an idea of how to approach this and will open a draft PR.

Just to confirm, we would like to parallelize all instances we should scope this issue to the ETL first and validate that this provides performance improvements in IO-bound tasks.

It would be more opportune to then explore scoring all datasets in parallel. What are your thoughts on splitting this issue into two? One for ETL and one for scoring?

sverchdotgov:
I kind of went on a tangent in https://github.com/usds/justice40-tool/pull/1039#pullrequestreview-830528432 as part of trying to understand the pipeline. Meeting with @saran-ahluwalia tomorrow about the PR, and I'm thinking there might be some other unrelated perf issues to file.

[...]

PR #1253, opened 2022-02-10
Run ETL processes in parallel 

lucasmbrown-usds:
#1038 . 

It seemed like waiting [70 minutes for tests to pass](https://github.com/usds/justice40-tool/runs/5131694186?check_suite_focus=true) on Github actions was really going to slow us down, so I implemented this. On my local, it runs in 4 minutes (!!!).

Looks like the GHA servers have 2 CPUs, and either way, I implemented this as threads (not processes) so it can run multiple threads concurrently on the same CPU. So this should speed up local as well as the build server. 

I had to modify all the ETLs that were using `TMP_PATH` in order to use a new method `get_tmp_path` that uses the class name as a directory for the temporary files. This accomplishes two things: 

1. Make sure there's no collisions between files with the same name from multiple ETLs being written to the temp directory (not super common, but there are still some common names like "data.zip" or "usa.csv" or something like that). 
2. Make sure that at the end of each individual ETL process, when we clean up the results of the ETL, it only cleans up the temporary files from that specific ETL and doesn't clobber other active ETL processes.


data/data-pipeline/data_pipeline/etl/runner.py
```diff
@@ -1,13 +1,18 @@
...
 from data_pipeline.etl.score.etl_score_post import PostScoreETL
+from data_pipeline.utils import get_module_logger
 
 from . import constants
 
+logger = get_module_logger(__name__)
 
-def get_datasets_to_run(dataset_to_run: str):
+
+def _get_datasets_to_run(dataset_to_run: str) -> typing.List[dict]:
```
lucasmbrown-usds, line 15: Renaming this to reflect that it's a private method.

data/data-pipeline/data_pipeline/ipython/scoring_comparison.ipynb
```diff
@@ -102,7 +102,9 @@
     "# Create the state ID by taking the first two digits of the FIPS CODE of the tract.\n",
```
lucasmbrown-usds, line 102: Just `black` running here. We should figure out why it's chaotically going back and forth.

data/data-pipeline/data_pipeline/etl/constants.py
```diff
@@ -1,8 +1,13 @@
 DATASET_LIST = [
```
lucasmbrown-usds, line 1: Reordering this so that the longest-running ETLs start first. This will speed up parallelization.

esfoobar-usds (approved):
This is such a game changer!

Now let's see how it runs on GHA.

🍏

esfoobar-usds:
@lucasmbrown-usds as a final ask: is there a way to add more logger outputs on NRI, which takes the most time to run?

PR #1264, opened 2022-02-11
Issue 1050: Install and run pre-commit autoformatters

lucasmbrown-usds:
This will run a variety of autoformatters on every commit. To work, it needs to be installed by every developer on their local. 

This helps make sure we don't check code into git without the proper autoformatters run. It'll also help us pass `pylint` and `flake8` more frequently.

To get up and running, see the new section in the README.md: 

> 
> ### Configuring pre-commit hooks
> 
> To promote consistent code style and quality, we use git pre-commit hooks to automatically lint and reformat our
> code before every commit we make to the codebase. Pre-commit hooks are defined in
> [`.pre-commit-config.yaml`](../.pre-commit-config.yaml).
> 
> 1.  First, install [`pre-commit`](https://pre-commit.com/) globally:
> 
>         $ brew install pre-commit
> 
> 2.  Run `pre-commit install` to install the specific git hooks used in this repository.
> 
> Now, any time you commit code to the repository, the hooks will run on all modified files automatically. If you wish,
> you can force a re-run on all files with `pre-commit run --all-files`.
> 

BTW, @esfoobar-usds I don't actually know how to update the table on contents in readme.md?


lucasmbrown-usds:
@vim-usds -- these autoformatters slightly updated the snapshot test results. It looks all cosmetic in a way that would be ignored by the actual front-end presentation (just cleaning up whitespace) -- could you take a look? https://github.com/usds/justice40-tool/runs/5157737698?check_suite_focus=true 

If you agree I can update the snapshot test results.

vim-usds:
@lucasmbrown-usds  - during the staging deployment we can see that the tests are failing [here](https://github.com/usds/justice40-tool/runs/5157737698?check_suite_focus=true).

It seems that this pre-commit hook is updating the snapshots which is mostly likely breaking the FE snapshots. We have prettier installed on the FE to standardize the formatting the files on the FE (which runs on save).
Is there a way to remove the client folder from this pre-commit hook?

Issue #1222, opened 2022-02-02
determine why island territories in tiles csv are missing state codes


emma-nechamkin:
in the tiles csv, the territories (Guam, American Samoa, Mariana Islands, and USVI) appear as `NaN` for their state codes (`SF`)

lucasmbrown-usds:
@emma-nechamkin -- do you mind making the title a little more descriptive here? Thanks!

PR #1272, opened 2022-02-15
Updating tiles csv to include state code

emma-nechamkin:
#1222 

Currently, the tiles.csv file has nulls in place of state abbreviations. We would like to retain state abbreviations in the csv.


emma-nechamkin:
Note that the fixtures here need updating -- the column order has changed.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -54,7 +57,9 @@ def _extract_counties(self, county_path: Path) -> pd.DataFrame:
     def _extract_states(self, state_path: Path) -> pd.DataFrame:
         logger.info("Reading States CSV")
         return pd.read_csv(
-            state_path, dtype={"fips": "string", "state_abbreviation": "string"}
+            state_path,
```
emma-nechamkin, line 60: i think we can just read in the columns that we want to read, rather than dropping columns in place.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -159,19 +163,25 @@ def _create_score_data(
...
-        county_state_merged = counties_df.merge(
-            states_df, on="State Abbreviation", how="left"
-        )
 
-        # merge state + county with score
-        score_county_state_merged = score_df.merge(
-            county_state_merged,
+        logger.info("Merging county info with score info")
+        score_county_merged = score_df.merge(
+            # We drop state abbreviation so we don't get it twice
```
emma-nechamkin, line 169: The crucial difference between this and the original code is the order of the joins. In the original code, we merged state and county, and then merged from state and county to score based on GEOID. Island territories, however, do not have GEOIDs. To allow them to retain the state information, we merge counties first and then merge states. This is slightly annoying from the perspective of an n_tracts join two times.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -159,19 +163,25 @@ def _create_score_data(
...
             on="GEOID",  # GEOID is the county ID
             how="left",
         )
 
+        logger.info("Merging state info with county-score info")
+        # Here, we need to join on a separate key, since there's no
+        # GEOID for the territories.
+        score_county_merged["State Code"] = score_county_merged[
+            self.GEOID_TRACT_FIELD_NAME
+        ].str[:2]
```
emma-nechamkin, line 180: We have to create this for the join. If we adopt this method we should probably change the section of code that drops fips codes to use the State Code information rather than `GEOID10_TRACT.str.startswith`

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
lucasmbrown-usds: Nice! You're an excellent campsite cleaner. I really appreciate this about your work.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
lucasmbrown-usds: I'm sorry, I'm not following this. "Island territories, however, do not have GEOIDs" - ? Do you mean they don't have counties?

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -159,19 +163,25 @@ def _create_score_data(
...
+        score_county_state_merged = score_county_merged.merge(
+            states_df, left_on="State Code", right_on="State Code", how="left"
```
lucasmbrown-usds, line 182: nitty nit: you know i get nervous whenever i see string field names floating around. maybe make these module-level constants (i.e., at the top of the file) or class-level constants (i.e., at the start of the class)? Probably class-level makes a little more sense.

lucasmbrown-usds (approved):
Overall looks good. I don't quite understand the logic you're trying to convey - a little more information would be helpful (on the part where I commented). 

For future reference, we could also refactor this code so that the FIPS / State or Territory / County info gets created as an ETL process and joined in `etl_score`, rather than added in `post` like this. That would be a bit more consistent and automatically parallelized. 

Small note: a little overview on the PR when you submit it can be helpful to get a sense of the approach, challenges, areas where you especially want feedback, etc.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
emma-nechamkin: yes! that's exactly what I mean. Because they don't have counties, in the `state_county_merged_df`, there are nulls for `GEOID`. Thus, when we join on `GEOID`, we drop the island area information. Does this make more sense? I'll update the comment inline.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
emma-nechamkin: `For future reference, we could also refactor this code so that the FIPS / State or Territory / County info gets created as an ETL process and joined in etl_score, rather than added in post like this. That would be a bit more consistent and automatically parallelized.` -> this makes a toooon of sense to me. Wondering if we can do this in The Big Refactor?

emma-nechamkin:
this fails here, but the test is identical to my local run of main and also it works when i run it locally? o.O?

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
lucasmbrown-usds: Totally agree! And after we have tests would be good.

esfoobar-usds (approved):
🍏

Issue #1061, opened 2021-12-14
As a user who often uses ArcGIS, I want the CEJST data download to be available as ESRI shapefiles.


lucasmbrown-usds:
**Description**
Convert the CEJST data from geojson into ESRI shapefiles so that users who primarily use ArcGIS can easily load the CEJST data into their own tool.

Could either use `ogr2ogr` (I think this is more consistent with our current approach) https://gis.stackexchange.com/a/73026 or `geopandas` (https://gis.stackexchange.com/a/397726).

lucasmbrown-usds:
FYI an example use of `ogr2ogr` in the codebase is `data/data-pipeline/data_pipeline/etl/sources/census/etl.py:108`, and there's some `tippecanoe` usage in `data/data-pipeline/data_pipeline/tile/generate.py`.

PR #1275, opened 2022-02-15
Creating shapefiles for ArcGIS users

emma-nechamkin:
#1061 

Convert the CEJST data from geojson into ESRI shapefiles so that users who primarily use ArcGIS can easily load the CEJST data into their own tool.

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
+++ b/data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
@@ -27,6 +27,8 @@ class GeoScoreETL(ExtractTransformLoad):
         self.SCORE_LOW_GEOJSON = self.SCORE_GEOJSON_PATH / "usa-low.json"
         self.SCORE_HIGH_GEOJSON = self.SCORE_GEOJSON_PATH / "usa-high.json"
 
+        self.SCORE_SHP_PATH = self.DATA_PATH / "score" / "shapefile"
+        self.SCORE_SHP_FILE = self.SCORE_SHP_PATH / "usa.shp"
         self.SCORE_CSV_PATH = self.DATA_PATH / "score" / "csv"
         self.TILE_SCORE_CSV = self.SCORE_CSV_PATH / "tiles" / "usa.csv"
 
@@ -94,6 +96,7 @@ class GeoScoreETL(ExtractTransformLoad):
         fields = [self.GEOID_FIELD_NAME, self.GEOMETRY_FIELD_NAME]
         self.geojson_usa_df = self.geojson_usa_df[fields]
 
+        # TODO update this join
         logger.info("Merging and compressing score CSV with USA GeoJSON")
         self.geojson_score_usa_high = self.score_usa_df.merge(
             self.geojson_usa_df, on=self.GEOID_FIELD_NAME, how="left"
@@ -102,8 +105,6 @@ class GeoScoreETL(ExtractTransformLoad):
         self.geojson_score_usa_high = gpd.GeoDataFrame(
             self.geojson_score_usa_high, crs="EPSG:4326"
         )
-
-        logger.info(f"Columns: {self.geojson_score_usa_high.columns}")
 
         usa_simplified = self.geojson_score_usa_high[
             [
@@ -148,8 +149,9 @@ class GeoScoreETL(ExtractTransformLoad):
         )
 
         # round to 2 decimals
-        decimals = pd.Series([2], index=[self.TARGET_SCORE_RENAME_TO])
-        self.geojson_score_usa_low = self.geojson_score_usa_low.round(decimals)
+        self.geojson_score_usa_low = self.geojson_score_usa_low.round(
+            {self.TARGET_SCORE_RENAME_TO: 2}
+        )
 
     def _aggregate_to_tracts(
         self, block_group_df: gpd.GeoDataFrame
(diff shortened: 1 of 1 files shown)


emma-nechamkin:
QQ here @lucasmbrown-usds @esfoobar-usds 

I think if we use `gpd` to do this, we can do it in the load step in parallel as: 
```
def write_esri_shapefile():
            logger.info("Producing ESRI shapefiles")
            self.geojson_score_usa_high.to_file(
                self.SCORE_SHP_FILE
            )
            logger.info("Completed writing shapefile")

        with concurrent.futures.ThreadPoolExecutor() as executor:
            futures = {
                executor.submit(task)
                for task in [
                    write_high_to_file,
                    write_low_to_file,
                    write_esri_shapefile,
                ]
            }
```

This is consistent with how other files are written in the `load` step, vs the code we would have to write for `ogr2ogr`, after the parallelization:

```
            logger.info("Producing ESRI shapefiles")
            cmd = [
                 "ogr2ogr",
                 "-f",
                 "GeoJSON",
                 str(self.SCORE_HIGH_GEOJSON),
                 str(self.SCORE_SHP_FILE),
             ]
             subprocess.run(cmd, check=True)
```

this would be in the `etl_score_geo` step. 

What seems reasonable to you / what do you guys think?

emma-nechamkin:
(also this is not ready for review and most notably all column names have to be truncated to 10 characters, because ESRI shapefiles have column length limits)

emma-nechamkin:
@esfoobar-usds and I chatted about using `gpd` instead of `ogr2ogr`

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -27,6 +28,10 @@ def __init__(self, data_source: str = None):
         self.SCORE_LOW_GEOJSON = self.SCORE_GEOJSON_PATH / "usa-low.json"
         self.SCORE_HIGH_GEOJSON = self.SCORE_GEOJSON_PATH / "usa-high.json"
 
+        self.SCORE_SHP_PATH = self.DATA_PATH / "score" / "shapefile"
```
emma-nechamkin, line 31: This is new. I tried to mirror the existing code, but let me know if you think this needs its own script.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -221,10 +226,46 @@ def write_low_to_file():
             )
             logger.info("Completed writing usa-low")
 
+        def write_esri_shapefile():
```
emma-nechamkin, line 229: this is pretty gross. if you guys have a suggestion for how to streamline it, i am ALL EARS.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -221,10 +226,46 @@ def write_low_to_file():
...
+            renaming_map = {}
+
+            # allows us to quickly rename / describe columns
+            reversed_tiles = {
+                short: long
+                for long, short in constants.TILES_SCORE_COLUMNS.items()
+            }
+            for column in self.geojson_score_usa_high.columns:
+                # take first 10 characters, max due to ESRI constraints
+                new_col = column[:10]
```
emma-nechamkin, line 248: we need to do this because ESRI shp can only have 10 character long column names -- so the idea here would be to create columns that are only 10 or fewer characters, and then also print out a codebook for ease of use -- since the end user might not understand `IALHSEF` as a field. So that's what this does.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -221,10 +226,46 @@ def write_low_to_file():
...
+                codebook[new_col] = reversed_tiles.get(column, column)
+                if new_col != column:
+                    renaming_map[column] = new_col
+            pd.Series(codebook).reset_index().rename(
```
emma-nechamkin, line 252: this prints out the new codebook to a csv

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -221,10 +226,46 @@ def write_low_to_file():
...
+                # kept as strings because no downstream impacts
+                columns={0: "column", "index": "meaning"}
+            ).to_csv(self.SCORE_SHP_CODE_CSV, index=False)
+            self.geojson_score_usa_high.rename(columns=renaming_map).to_file(
```
emma-nechamkin, line 256: this renames the columns that we need to rename -- it's just a few. It's kinda ugly but I do not want to deep copy the df and I also do not want to modify it in any way because of the parallelization.

emma-nechamkin:
Also note that in the final shapefiles, these columns are included: 
```
UI Experie,UI Experience
Thresholds,Thresholds
```
I figured this might be helpful because people are mapping, but also happy to get rid of them.

emma-nechamkin:
(also getting this tox error: ```FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/justice40-tool/justice40-tool/data/data-pipeline/data_pipeline/data/tmp/downloaded.zip'``` ??)

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
esfoobar-usds: I like it!

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
esfoobar-usds: Interesting

esfoobar-usds (approved):
This looks good to me! 🍏 

One thing we'd need to add is a link to the new ESRI file from the website @vim-usds @KameronKerger 

It's be also cool to reach out to the folks at ArcGIS to have this data as a ready to go table for users to play with.

PR #1291, opened 2022-02-17
Updating threshold counter to skip some fips codes

emma-nechamkin:
i havent checked this, but it should screen PR out

### DIFF
--- a/data/data-pipeline/data_pipeline/score/score_m.py
+++ b/data/data-pipeline/data_pipeline/score/score_m.py
@@ -113,15 +113,29 @@ class ScoreM(Score):
         )
 
     def _increment_total_eligibility_exceeded(
-        self, columns_for_subset: list
+        self, columns_for_subset: list, skip_fips: tuple = ()
     ) -> None:
         """
         Increments the total eligible factors for a given tract
-        """
 
-        self.df[field_names.THRESHOLD_COUNT] += self.df[columns_for_subset].sum(
-            axis=1, skipna=True
-        )
+        The new skip_fips argument specifies which (if any) fips codes to
+        skip over for incrementing.
+        This allows us to essentially skip data we think is of limited veracity,
+        without overriding any values in the data.
+        THIS IS A TEMPORARY FIX.
+        """
+        if skip_fips:
+            self.df[field_names.THRESHOLD_COUNT] += np.where(
+                self.df[field_names.GEOID_TRACT_FIELD].str.startswith(
+                    skip_fips
+                ),
+                0,
+                self.df[columns_for_subset].sum(axis=1, skipna=True),
+            )
+        else:
+            self.df[field_names.THRESHOLD_COUNT] += self.df[
+                columns_for_subset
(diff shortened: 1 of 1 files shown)


data/data-pipeline/data_pipeline/score/score_m.py
```diff
@@ -182,7 +196,9 @@ def _climate_factor(self) -> bool:
             & self.df[field_names.FPL_200_AND_COLLEGE_ATTENDANCE_SERIES]
         )
 
-        self._increment_total_eligibility_exceeded(climate_eligibility_columns)
+        self._increment_total_eligibility_exceeded(
+            climate_eligibility_columns, skip_fips=("72")
```
emma-nechamkin, line 200: i'm sorry it's a string not a constant. all of them could in theory have different fips codes, etc. not sure where to put this.

data/data-pipeline/data_pipeline/score/score_m.py
```diff
@@ -113,15 +113,29 @@ def _create_low_income_and_low_college_attendance_threshold(
...
-        """
 
-        self.df[field_names.THRESHOLD_COUNT] += self.df[columns_for_subset].sum(
-            axis=1, skipna=True
-        )
+        The new skip_fips argument specifies which (if any) fips codes to
+        skip over for incrementing.
+        This allows us to essentially skip data we think is of limited veracity,
+        without overriding any values in the data.
+        THIS IS A TEMPORARY FIX.
```
lucasmbrown-usds, line 125: Nit: let's make the comments a little more publicly presentable, thanks!

emma-nechamkin:
Also just as an fyi -- i checked this briefly by looking at the max thresholds exceeded per fips and then looking at places where this varied vis a vis s3 (i was running with the updated ag, so there were a few places). But SPOT CHECKED this looks fine.

esfoobar-usds (approved):
🍏 

Usually if I see something repeated over and over (like `skip_fips=("72")`) I think there's some code smell -- like maybe setting the `72` in just one point of the code, but I don't think it's a huge deal at this point

data/data-pipeline/data_pipeline/score/score_m.py
lucasmbrown-usds: That's fine for now. Eventually it could be something like skip_fips=constants.PR_FIPS_CODE or something. Not today.

data/data-pipeline/data_pipeline/score/score_m.py
```diff
@@ -113,15 +113,29 @@ def _create_low_income_and_low_college_attendance_threshold(
         )
 
     def _increment_total_eligibility_exceeded(
-        self, columns_for_subset: list
+        self, columns_for_subset: list, skip_fips: tuple = ()
```
lucasmbrown-usds, line 116: nit though: why is this a tuple? is that because str.startswith takes a tuple?

lucasmbrown-usds (commented):
Approved.

PR #1304, opened 2022-02-17
Adding PR tracts to AMI

emma-nechamkin:
This pulls down a list of tracts from S3 in puerto rico to adjust for them being missing from geocorr. This is NOT a permanent fix.


esfoobar-usds (approved):
🍏

PR #1340, opened 2022-02-25
Fix broken YML links and add PR template

vim-usds:
I was trying to track the experience of a new person trying to make a contribution. As I tried to do so the build failed due to links failing. I believe these links are failing because we may have updated master directly. I suggest we avoid that.

This PR will
- add a PR template to help OS contributors
- correct broken MD links
- make all YML files follow the same capital case naming scheme


esfoobar-usds (approved):
🍏

PR #1342, opened 2022-02-25
WY tracts are not showing up until zoom >7

### LINKED ISSUE #1320: DACs don't show up for WY until you zoom > 7
It appears to not show any disadv. communities for zoom < 7, and then shows them for zoom > 7. This is the transition point from “low zoom” to “high zoom” tiles.

This doesn't seem to be the case for other states.

emma-nechamkin:
Passing this off to @vim-usds 

closes #1320


data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -92,26 +93,27 @@ def transform(self) -> None:
 
         logger.info("Pruning Census GeoJSON")
         fields = [self.GEOID_FIELD_NAME, self.GEOMETRY_FIELD_NAME]
-        self.geojson_usa_df = self.geojson_usa_df[fields]
 
         logger.info("Merging and compressing score CSV with USA GeoJSON")
-        self.geojson_score_usa_high = self.score_usa_df.merge(
-            self.geojson_usa_df, on=self.GEOID_FIELD_NAME, how="left"
+        self.geojson_score_usa_high = self.score_usa_df.set_index(
```
emma-nechamkin, line 98: I changed this slightly so it would run ever so slightly faster

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -92,26 +93,27 @@ def transform(self) -> None:
...
-        logger.info(f"Columns: {self.geojson_score_usa_high.columns}")
-
         usa_simplified = self.geojson_score_usa_high[
             [
-                self.GEOID_FIELD_NAME,
                 self.TARGET_SCORE_SHORT_FIELD,
                 self.GEOMETRY_FIELD_NAME,
             ]
-        ].reset_index(drop=True)
+        ].reset_index(drop=False)
```
emma-nechamkin, line 116: because this now merges on index, I don't want to drop the index

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -125,7 +127,11 @@ def transform(self) -> None:
 
         usa_tracts = gpd.GeoDataFrame(
             usa_tracts,
-            columns=[self.TARGET_SCORE_RENAME_TO, self.GEOMETRY_FIELD_NAME],
+            columns=[
+                self.TARGET_SCORE_RENAME_TO,
+                self.GEOMETRY_FIELD_NAME,
+                self.GEOID_FIELD_NAME,
```
emma-nechamkin, line 133: I think I left this in from testing yikes

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -155,24 +161,24 @@ def _aggregate_to_tracts(
         self, block_group_df: gpd.GeoDataFrame
     ) -> gpd.GeoDataFrame:
         # The tract identifier is the first 11 digits of the GEOID
-        block_group_df["tract"] = block_group_df.apply(
-            lambda row: row[self.GEOID_FIELD_NAME][0:11], axis=1
-        )
+        block_group_df["tract"] = block_group_df[self.GEOID_FIELD_NAME].str[:11]
```
emma-nechamkin, line 164: this is faster...

but more importantly, I'm not sure why we are doing this step. 

I believe `aggregate_to_tracts` is a holdover from when you were using block groups. here, we already have a dataframe at the census tract level, and so we very much can still `dissolve` it from a "can this be done?" but I think we don't have to

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -155,24 +161,24 @@ def _aggregate_to_tracts(
...
     def _create_buckets_from_tracts(
         self, state_tracts: gpd.GeoDataFrame, num_buckets: int
     ) -> gpd.GeoDataFrame:
-        # assign tracts to buckets by D_SCORE
+        # assign tracts to buckets by score
         state_tracts.sort_values(self.TARGET_SCORE_RENAME_TO, inplace=True)
-        SCORE_bucket = []
+        score_bucket = []
+
+        # while use_buckets <= 0
```
emma-nechamkin, line 175: this is a holdover from testing that I did not delete and I am sorry!

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -155,24 +161,24 @@ def _aggregate_to_tracts(
...
-        )
+        block_group_df["tract"] = block_group_df[self.GEOID_FIELD_NAME].str[:11]
         state_tracts = block_group_df.dissolve(by="tract", aggfunc="mean")
         return state_tracts
 
     def _create_buckets_from_tracts(
         self, state_tracts: gpd.GeoDataFrame, num_buckets: int
     ) -> gpd.GeoDataFrame:
-        # assign tracts to buckets by D_SCORE
+        # assign tracts to buckets by score
```
emma-nechamkin, line 171: here's where i think we need to do some thinking. 

right now, WY has score buckets that get absorbed into 0. This is where the buckets are assigned. It might make sense to ensure that all buckets are homogeneous, so either all 0 or all 1.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -155,24 +161,24 @@ def _aggregate_to_tracts(
...
+        score_bucket = []
+
+        # while use_buckets <= 0
         bucket_size = math.ceil(
             len(state_tracts.index) / self.NUMBER_OF_BUCKETS
         )
         for i in range(len(state_tracts.index)):
-            SCORE_bucket.extend([math.floor(i / bucket_size)])
-        state_tracts[f"{self.TARGET_SCORE_RENAME_TO}_bucket"] = SCORE_bucket
+            score_bucket.extend([math.floor(i / bucket_size)])
```
emma-nechamkin, line 180: This is a formatting change

vim-usds:
@emma-nechamkin - thanks so much for taking a look at this! Also I know you jumped into the bug time on Friday to talk about this but I was so distracted by intl talk that I missed this until now. Can we touch base next on this! TY!!!

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
vim-usds: @esfoobar-usds - any thoughts here?

esfoobar-usds: It's indeed a holdover from when we were census-block based and can be removed.

emma-nechamkin:
Okay, so I looked at this a little bit more, and I do not think my initial hunch was correct.

In the grab below, I've printed out a temporary file after the aggregate buckets step. Then, I filter to just WY. What you see is that WY has two buckets in it, 4 and 7. All tracts in 4 are 0s and all tracts in 7 are 1s for score. So, the bucketing appears to be working. 

Screen Shot 2022-03-02 at 10 47 17 PM

If you notice, bucket 6 is mixed. So, I added something that will increment the bucket number to be slightly bigger and reduce the number of "wrong" scores per bucket... but who knows if this is worth it. 

The larger point is that I think the process of making the geojsons is correct (the bucketing at least). It's pretty annoying to load geojsons on a map locally, so it's possible I am misunderstanding... but I am wondering if it could be something on the front end?

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
emma-nechamkin: So... I think this initial thought was wrong. I did implement code to make the buckets more homogeneous, but that is not what was going on with WY.

emma-nechamkin: (There COULD be something going on with the compression in `_breakup_multipolygons` but I'm not sure?)

emma-nechamkin:
(and with more buckets, the aggregation before compression looks like this...)
Screen Shot 2022-03-02 at 11 51 46 PM

emma-nechamkin:
Okay, I think this is going to work: 
Screen Shot 2022-03-04 at 5 29 44 PM

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
esfoobar-usds: Looks good!

esfoobar-usds (approved):
🍏

PR #1252, opened 2022-02-09
YAML Config for Downloadable Assets

esfoobar-usds:
Fixes #1107, or at least gets us prepared for it, but also related to #1215.

Happy to get some initial feedback. The territories tab is not part of the initial version, but this PR already writes the downloadable CSV completely driven by the YAML config file.

PR #1349, opened 2022-03-01
Cleaning up quick code

emma-nechamkin:
Taking a minute to clean up / resubmit for review some of the quick and dirty changes from launch.


data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -298,33 +298,19 @@ def _add_percentiles_to_df(
...
                 f"{output_column_name_root}"
                 f"{field_names.PERCENTILE_FIELD_SUFFIX}"
-            ] = np.where(
-                df[field_names.AGRICULTURAL_VALUE_BOOL_FIELD].isna(),
-                np.nan,
-                df[field_names.GEOID_TRACT_FIELD]
-                .map(temporary_ranking)
-                .fillna(0),
+            ] = (
+                df.where(df[field_names.AGRICULTURAL_VALUE_BOOL_FIELD] == True)[
```
emma-nechamkin, line 307: this should be much cleaner. asserted that it was equal to the existing code.

data/data-pipeline/data_pipeline/score/score_m.py
```diff
@@ -195,7 +196,8 @@ def _climate_factor(self) -> bool:
         )
 
         self._increment_total_eligibility_exceeded(
-            climate_eligibility_columns, skip_fips=("72")
+            climate_eligibility_columns,
```
emma-nechamkin, line 199: this should get rid of the bad code smell!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
emma-nechamkin: `== True` is gross but the bool has `NaN`s that we need to keep, so it's necessary unfortunately.

data/data-pipeline/pyproject.toml
```diff
@@ -39,6 +39,7 @@ tox = "^3.24.0"
 pytest-mock = "^3.6.1"
 tox-poetry = "^0.4.1"
 pandas-vet = "^0.2.2"
+nb-black = "^1.0.7"
```
emma-nechamkin, line 42: added this from the command line (`$ poetry add -D nb_black`)

esfoobar-usds (changes requested):
PR is not passing.

You can patch this commit or make these one line change: https://github.com/usds/justice40-tool/pull/1252/commits/64d614bca00eb3989dc764e75abedf7662dc1e11

esfoobar-usds (approved):
🍏

Issue #1263, opened 2022-02-10
As a data developer, I want to replace our custom data snapshots approach with `pytest-snapshot`


lucasmbrown-usds:
**Description**
https://pypi.org/project/pytest-snapshot/ provides a better solution than our home-grown snapshot tests because it does not require as much additional custom logic (because you can simple wrap the assertion as well as the snapshot writing into a single line). We should look into updating this custom logic to use this framework. 

At the same time, some of our custom tests use fixtures as the *inputs* to the test, not the outputs, and `pytest-snapshot` may not be as well suited for that. 

Investigate the pros and cons of the upgrade and implement the desired approach before we roll out unit testing to the other sources as part of #1217.

lucasmbrown-usds:
Just do for NRI for now.

widal001:
@emma-nechamkin and I were thinking about using this issue to serve as a scratch pad for our evaluation of `pytest-snapshot` to see if it meets the needs currently addressed by [`TestETL.test_update_test_fixtures()`](https://github.com/usds/justice40-tool/blob/521c61dff307212a91ad225bb5a32c9e88548746/data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py#L504). If those needs are met, it may also require some refactoring of the tests overall to adapt them to use `pytest-snapshot`

Below I've outlined my current understanding of the current use case and highlighted some key questions we'll want to answer about `pytest-snapshot` that may inform if/how we use it to replace the current snapshot update mechanism:

### Use Case

Our use case for updating test fixtures has the following characteristics/needs:

- **Manual Updates of Input Data:** I want to manually create/edit an input file that mocks the data source for a particular datasource, so that I have control over the seed data that populates the snapshots created at each step of the ETL process I'm testing
- **Sequential Updates of Snapshots**: When I need to change the fixtures, I want to automatically update each snapshot in the ETL process using the output of the previous step as the input to the subsequent step (and the manually supplied input for the first step), so that I can trust the changes to the input will propagate across steps and I don't have to manually provide input data for each step. 
- **Explicit Execution:** I _don't_ want ETL snapshots to be updated unless I explicitly pass a certain flag to `pytest` via the CLI during the execution of tests, so that I don't accidentally update snapshots when I run the tests to validate the ETL outputs against those snapshots.
[...]

PR #1355, opened 2022-03-02
Updating testing to include pytest-snapshot

emma-nechamkin:
#1263


emma-nechamkin:
Right now, this runs end to end including with snapshots getting updated. Here are the commands:

```
Emmas-MBP-2:j40 emmausds$ poetry run pytest data_pipeline/tests/sources/national_risk_index/test_etl.py --snapshot-update
/usr/local/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
=========================================== test session starts ===========================================
platform darwin -- Python 3.9.10, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/emmausds/Desktop/justice40-tool/data/data-pipeline, configfile: pytest.ini
plugins: mock-3.7.0, snapshot-0.8.1
collected 25 items                                                                                        

data_pipeline/tests/sources/national_risk_index/test_etl.py ..........s..........s...               [100%]

====================================== 23 passed, 2 skipped in 0.60s ======================================

Emmas-MBP-2:j40 emmausds$ poetry run pytest data_pipeline/tests/sources/national_risk_index/test_etl.py -v
/usr/local/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
=========================================== test session starts ===========================================
platform darwin -- Python 3.9.10, pytest-6.2.5, py-1.11.0, pluggy-1.0.0 -- /Users/emmausds/Library/Caches/pypoetry/virtualenvs/data-pipeline-mN7A7TGn-py3.9/bin/python
cachedir: .pytest_cache
rootdir: /Users/emmausds/Desktop/justice40-tool/data/data-pipeline, configfile: pytest.ini
plugins: mock-3.7.0, snapshot-0.8.1
collected 25 items                                                                                        

data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_existence_of_test_fixtures_base PASSED [  4%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_init_base PASSED         [  8%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_get_output_file_path_base PASSED [ 12%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_fixtures_contain_shared_tract_ids_base PASSED [ 16%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_transform_sets_output_df_base PASSED [ 20%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_transform_base PASSED    [ 24%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_load_base PASSED         [ 28%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_validate_base PASSED     [ 32%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_full_etl_base PASSED     [ 36%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_get_data_frame_base PASSED [ 40%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_update_test_fixtures SKIPPED [ 44%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_existence_of_test_fixtures_base PASSED [ 48%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_init_base PASSED [ 52%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_get_output_file_path_base PASSED [ 56%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_fixtures_contain_shared_tract_ids_base PASSED [ 60%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_transform_sets_output_df_base PASSED [ 64%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_transform_base PASSED [ 68%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_load_base PASSED [ 72%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_validate_base PASSED [ 76%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_full_etl_base PASSED [ 80%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_get_data_frame_base PASSED [ 84%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_update_test_fixtures SKIPPED [ 88%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_init PASSED [ 92%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_get_output_file_path PASSED [ 96%]
data_pipeline/tests/sources/national_risk_index/test_etl.py::TestNationalRiskIndexETL::test_extract PASSED [100%]

====================================== 23 passed, 2 skipped in 0.60s ======================================
```

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -30,6 +29,14 @@ class TestETL:
     _TRANSFORM_CSV_FILE_NAME = "transform.csv"
     _OUTPUT_CSV_FILE_NAME = "output.csv"
 
+    # This *does* need to be updated in the child class. It specifies where the "sample data" is
```
emma-nechamkin, line 32: This is really inelegant. One annoying thing about these tests is you have to "set them up" with sample data. I thought it would be nice if we could only put the sample data in one place -- the data folder for the test -- and then update the "sample data" for the transform step using the same `--snapshot-update` flag. A key issue here, though, is that to do this, we have to read the CSV in and pandas will change some types (not values). Do we think this is a problem / do you have any thoughts @widal001?

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -263,25 +246,49 @@ def test_load_base(self, mock_etl):
...
+        snapshot.snapshot_dir = self._DATA_DIRECTORY_FOR_TEST
+        snapshot.assert_match(
+            actual_output.to_csv(index=False),
+            self._OUTPUT_CSV_FILE_NAME,
         )
 
-        # setup - load expected output
-        expected_output = pd.read_csv(
-            filepath_or_buffer=expected_output_csv_path,
+    def test_sample_data_exists(self, snapshot):
```
emma-nechamkin, line 264: this is the inelegant test described above

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -151,13 +126,27 @@ def test_extract(self, mock_etl, mock_paths):
...
             tmp_path / "NationalRiskIndexETL" / "NRI_Table_CensusTracts.csv"
         )
+
         assert extracted_file_path.is_file()
 
-        input_csv_path = (
-            self._DATA_DIRECTORY_FOR_TEST / self._INPUT_CSV_FILE_NAME
-        )
+    def test_extract(self, snapshot, mock_etl, mock_paths):
+        """Tests the extract method. I think this works
```
emma-nechamkin, line 133: I need to update these header comments and will do so! Note to myself.

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
emma-nechamkin: One other note -- this is because the snapshot-update requires string/bytes as a comparison.

emma-nechamkin:
This runs for me locally:

```
Emmas-MBP-2:j40 emmausds$ poetry run pytest data_pipeline/tests/sources/national_risk_index/test_etl.py --snapshot-update
/usr/local/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
=========================================== test session starts ===========================================
platform darwin -- Python 3.9.10, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/emmausds/Desktop/justice40-tool/data/data-pipeline, configfile: pytest.ini
plugins: mock-3.7.0, snapshot-0.8.1
collected 26 items                                                                                        

data_pipeline/tests/sources/national_risk_index/test_etl.py ..........................              [100%]

=========================================== 26 passed in 0.70s ============================================
Emmas-MBP-2:j40 emmausds$ poetry run pytest data_pipeline/tests/sources/national_risk_index/test_etl.py
/usr/local/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
=========================================== test session starts ===========================================
platform darwin -- Python 3.9.10, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /Users/emmausds/Desktop/justice40-tool/data/data-pipeline, configfile: pytest.ini
plugins: mock-3.7.0, snapshot-0.8.1
collected 26 items                                                                                        

data_pipeline/tests/sources/national_risk_index/test_etl.py ..........................              [100%]

=========================================== 26 passed in 0.59s ============================================
```

To run a single test, try:
```
$ Emmas-MBP-2:j40 emmausds$ poetry run pytest data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_transform_base
```

And to run a single snapshot update, try:
```
Emmas-MBP-2:j40 emmausds$ poetry run pytest data_pipeline/tests/sources/national_risk_index/test_etl.py::TestETL::test_transform_base --snapshot-update
```

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -137,11 +112,11 @@ def test_get_output_file_path(self, mock_etl, mock_paths):
...
         assert output_file_path == expected_output_file_path
 
-    def test_extract(self, mock_etl, mock_paths):
-        """Ensure the extract results are working as appropriate."""
+    def test_extract_path(self, mock_etl, mock_paths):
+        """Ensure the extract is happening where appropriate."""
+        # this is returning true for no reason?
         tmp_path = mock_paths[1]
 
-        # ETL instance returned by this method is never used, so we assign it to `_`.
```
widal001, line 144: I found this comment kind of helpful, was there a reason we wanted to take it out?

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -151,13 +126,27 @@ def test_extract(self, mock_etl, mock_paths):
...
+    def test_extract(self, snapshot, mock_etl, mock_paths):
+        """Tests the extract method. I think this works
 
-        # Make sure extracted file is equal to the input fixture:
-        assert filecmp.cmp(
-            f1=extracted_file_path, f2=input_csv_path, shallow=False
+        To update: $ poetry run pytest
+            data_pipeline/tests/sources/national_risk_index/test_etl.py::test
+            --snapshot-update
+        """
```
widal001, line 138: Appreciate the guidance here on how to update the test fixtures for _just_ this ETL class

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -137,11 +112,11 @@ def test_get_output_file_path(self, mock_etl, mock_paths):
         )
         assert output_file_path == expected_output_file_path
 
-    def test_extract(self, mock_etl, mock_paths):
-        """Ensure the extract results are working as appropriate."""
+    def test_extract_path(self, mock_etl, mock_paths):
+        """Ensure the extract is happening where appropriate."""
+        # this is returning true for no reason?
```
widal001, line 117: This might have been what you meant by your comment in the code, but do we still need `test_extract_path()` now that you've implemented `test_extract()`?

widal001 (commented):
@emma-nechamkin It seems like you've made great progress! I do think there could be some value in having us connect live on the question you raised about how to handle copying/validating the initial input data. I had a set of related questions that I added as comments to the files. 

I want to make sure I fully understand your current approach with these changes and understand how you envisioned other developers would make changes manual changes to the sample data and propagate those changes through the various snapshots. 

My instinct would be to try to change the values in the `data/input.csv` file in the given test directory, but changing that file didn't seem to make any of the tests fail. I'll follow up over slack to see if we can find a time to connect, though, since it's possible that I'm simply missing something! 

In general, this is awesome!! Love seeing so many lines coming out of the codebase and having them replaced with a more established solution.

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
emma-nechamkin: @widal001 this was an intentional design choice, but a very silly one! In our existing workflow, the sample data "lives" somewhere else. I also think -- but might be misremembering -- that this is necessary for when we load / transform the "actual" data. I like your idea a lot, though, and think we should implement it!

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
emma-nechamkin: added it back!

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
emma-nechamkin: That is what I meant! but on reflection, I *think* we actually still need this because we are testing the "native" path if that makes sense?

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -263,25 +246,49 @@ def test_load_base(self, mock_etl):
...
+        """This will test that the sample data exists where it's supposed to as it's supposed to
+
+        The caveat here is that types will change
+        """
+        snapshot.snapshot_dir = self._SAMPLE_DATA_PATH
```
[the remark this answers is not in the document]
emma-nechamkin: We talked about this, so I'm just documenting this here... we need data both in the tmp for the testing of the "actual" pipeline and also in the data directory for snapshot testing.

emma-nechamkin:
@widal001 and I had a really great chat about this. We settled on a few outstanding questions:

1. Right now, were someone to change the input.csv data, this would not propagate "up" unless they zipped the data. We worry this isn't human-readable or intuitive and wonder if we should make a "seed_data.csv" file that a user can modify by hand, that then updates a snapshot that is a zip that then gets unzipped? The advantage here is that it allows a user to change data manually. 
2. If we don't end up making this change, should we add a louder fail when the input.csv is changed?

emma-nechamkin:
@widal001 so I *think* pytest snapshot can check bytes against a file, but not zip files. Here's the error I cannot seem to get around:

```
_____________________________________________________________________________ TestETL.test_seed_data_has_propogated_to_input _____________________________________________________________________________

self = <data_pipeline.tests.sources.example.test_etl.TestETL object at 0x11b0f55e0>, snapshot = <pytest_snapshot.plugin.Snapshot object at 0x11b0cf6d0>, mock_etl = None
mock_paths = (PosixPath('/Users/emmausds/Desktop/justice40-tool/data/data-pipeline/data_pipeline/data/tmp/tests/pytest-of-emmausds/...mausds/Desktop/justice40-tool/data/data-pipeline/data_pipeline/data/tmp/tests/pytest-of-emmausds/pytest-168/data/tmp'))

    def test_seed_data_has_propogated_to_input(
        self, snapshot, mock_etl, mock_paths
    ):
        """Uses seed snapshot to update zip file of data
    
        This is necessary because we don't want to have to manually change a
        zip file every time we change the input data. This is an *excellent*
        design suggestion from Billy.
        """
        # Read in the seed data
        tmp_df = pd.read_csv(
            self._DATA_DIRECTORY_FOR_TEST / self._SEED_DATA_CSV_FILE_NAME
        )
    
        # Update the "zip" files of the input data
>       snapshot.assert_match(
            tmp_df.to_csv(index=False),
            self._INPUT_DATA_ZIP_FILE_NAME,
        )

data_pipeline/tests/sources/example/test_etl.py:295: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../Library/Caches/pypoetry/virtualenvs/data-pipeline-mN7A7TGn-py3.9/lib/python3.9/site-packages/pytest_snapshot/plugin.py:183: in assert_match
    expected_value = decode(encoded_expected_value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

data = b'PK\x03\x04\x14\x00\x00\x00\x08\x00H\xb3iT\x9f\xdc\x9d\x1f\xbf\x00\x00\x00a\x01\x00\x00\x05\x00\x00\x00inputM\xd09N\x...00\x00\x00\x80\x01\x00\x00\x00\x00inputPK\x05\x06\x00\x00\x00\x00\x01\x00\x01\x003\x00\x00\x00\xe2\x00\x00\x00\x00\x00'

    def _file_decode(data: bytes) -> str:
        """
        Returns the string that would be read from a file using ``path.read_text(string)``.
        See universal newlines documentation.
        """
>       return data.decode().replace('\r\n', '\n').replace('\r', '\n')
E       UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb3 in position 11: invalid start byte

../../../../Library/Caches/pypoetry/virtualenvs/data-pipeline-mN7A7TGn-py3.9/lib/python3.9/site-packages/pytest_snapshot/plugin.py:66: UnicodeDecodeError
```

And here's the code that generates it: 
```
    def _verify_compression(self, filepath):
        """Helper class to compress output of snapshot-update.

        Because we now want to be able to update the seed data and have those changes propogate
        forward, we have to verify that our "zip" files are actually zip files, since
        snapshot-update will not compress the file (there must be a better way!)
        """
        if not zipfile.is_zipfile(filepath):
            re_compress_df = pd.read_csv(filepath, compression=None)
            re_compress_df.to_csv(filepath, index=False, compression="zip")

    def test_seed_data_has_propogated_to_input(
        self, snapshot, mock_etl, mock_paths
    ):
        """Uses seed snapshot to update zip file of data

        This is necessary because we don't want to have to manually change a
        zip file every time we change the input data. This is an *excellent*
        design suggestion from Billy.
        """
        # Read in the seed data
        tmp_df = pd.read_csv(
            self._DATA_DIRECTORY_FOR_TEST / self._SEED_DATA_CSV_FILE_NAME
        )

        # Update the "zip" files of the input data
        snapshot.assert_match(
            tmp_df.to_csv(index=False),
            self._INPUT_DATA_ZIP_FILE_NAME,
        )

        # Make sure it is actually zipped
        self._verify_compression(
            self._DATA_DIRECTORY_FOR_TEST / self._INPUT_DATA_ZIP_FILE_NAME
        )
```

My first tip-off should have been that the snapshot update could NOT produce a zip file, but alas 🤷 .

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -87,7 +99,7 @@ def _setup_etl_instance_and_run_extract(
         This must be re-implemented in every child class.
 
         This method can be used by multiple tests that need to run the same fixtures
-        that need these same mocks, and by `test_update_test_fixtures`.
+        that need these same mocks.
```
emma-nechamkin, line 102: I'm getting a pylance error at 91: 

```
type() call should not be used in type annotation
  Use Type[T] instead
```

I think to fix this, you do something like (https://docs.python.org/3/library/typing.html):
```
from typing import TypeVar, Generic
from logging import Logger

T = TypeVar('T')
```
And then refer to it as `Generic[T]`, but I want to see if tox fails :)

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
emma-nechamkin: We talked about this and I think this works!

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
emma-nechamkin: I moved this into the header for the class

emma-nechamkin:
Anyway, @widal001 -- so what I think this means is that there are two options:

(1) Either we keep the testing as is, with the slightly annoying "I need to zip changes" 
(2) We have a `seed_data.csv` file and we assume that ANY TIME that file is changed (that is, it doesn't match the unzipped input file), we automatically update the snapshot for the seed file and then rezip the input data based on it. I might be overthinking, but I think it's a lot of gymnastics to get this to work because we cannot compare to the zip. 

Because (2) would (I think!) have to be automatic (we can't check the zip file with `snapshot-update`), I'm inclined to leave this as is, unfortunately. I'm worried that we will accidentally bury other errors if every time we have a difference between the seed data and the true data we kick off a new zip file. 

I'm interested also if you see a way out!

emma-nechamkin:
One last thought! The final option I considered here was to have a "mirror" -- like you change seed data and it compares to a snapshot of the output of the zipped file that would be created when the test is run. Then, when you instantiate the object or whatever, the first thing that happens is that mirror gets zipped into input.zip like files. But this seems so similar to just zipping the file on your own, but it's much more opaque. 

Anyway I think the Pareto frontier is just full of mediocre solutions and would appreciate any insight!

data/data-pipeline/data_pipeline/tests/sources/example/data/input.csv
```diff
@@ -1,16 +1,16 @@
...
-15009030100,13.60946983
-15009030201,13.73235164
-15001021402,14.73305116
-15001021800,16.60833857
-15009030402,16.002535
-15009030800,15.34818251
-15003010201,14.58788769
-15007040604,14.27704917
\ No newline at end of file
+,GEOID10_TRACT,Input Field 1
```
lucasmbrown-usds, line 1: Hmm - are we sure we want (1) the index and (2) to drop leading zeroes (i.e., "06007040300" has now gone to "6007040300" here) here?

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/data/output.csv
```diff
@@ -1,56 +1,56 @@
...
-15003001904,21.1782783212,0.0000003731,0.0000000000,0.0004239595,False
-15003002202,16.8452811998,0.0000003007,0.0000000000,0.0003313697,False
-15003002600,21.1208678596,0.0000000125,0.0000000000,0.0000024799,False
-15003002800,14.0640371335,0.0000000147,0.0000000000,0.0000053072,False
-15003003000,15.9122177661,0.0000000173,0.0000000000,0.0000088501,False
-15003003101,16.7794100011,0.0000000378,0.0000000000,0.0000269712,False
-15003003407,12.1430194003,0.0000000124,0.0000000000,0.0000023264,False
-15003003501,15.9378488267,0.0000000124,0.0000000000,0.0000023264,False
-15003004600,15.6406606485,0.0000000123,0.0000004599,0.0000021180,True
+06001020100,18.6199401875,3.5067e-06,0.0031812618,6.6152e-05,True
```
lucasmbrown-usds, line 2: i don't think the output should use scientific notation, right? we've gone out of our way in the past to make sure they write them out with full digits.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -137,11 +119,11 @@ def test_get_output_file_path(self, mock_etl, mock_paths):
...
```
lucasmbrown-usds, line 124: What does this comment `# this is returning true for no reason?` mean?

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -30,6 +34,14 @@ class TestETL:
     _TRANSFORM_CSV_FILE_NAME = "transform.csv"
     _OUTPUT_CSV_FILE_NAME = "output.csv"
 
+    # This *does* need to be updated in the child class. It specifies where the "sample data" is
+    # so that we do not have to manually copy the "sample data" when we run the tests.
+    _SAMPLE_DATA_PATH = (
+        pathlib.Path(__file__).parents[3] / "data" / "tmp" / "ExampleETL"
```
lucasmbrown-usds, line 40: We could also just make a unique snapshot directory per test that is within the same directory as the test file?

data/data-pipeline/data_pipeline/tests/sources/example/data/input.csv
emma-nechamkin: these are both super good points, yikes

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
emma-nechamkin: yikes, this is a comment i left in from when i was implementing, removed.

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
emma-nechamkin: Sure. I did not change the overall structure of the tests from the homegrown snapshot tests. I don't have strong feelings here and updated this -- but if you have more specific thoughts of where you'd like these to be saved, please let me know :)

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -17,19 +16,30 @@
...
+            data_pipeline/tests/sources/national_risk_index/test_etl.py::TestClassNameETL::<testname>
+            --snapshot-update
     """
 
     # In every child test class, change this to the class of the ETL being tested.
     _ETL_CLASS = ExampleETL
 
     # The following constants do not need to be updated in child class.
-    _INPUT_CSV_FILE_NAME = "input.csv"
+    _EXTRACT_CSV_FILE_NAME = "extract.csv"
```
emma-nechamkin, line 32: I've left a snapshot for this in right now (@esfoobar-usds) but renamed it so it's less confusing.

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -198,9 +186,68 @@ def test_fixtures_contain_shared_tract_ids_base(self, mock_etl, mock_paths):
         else:
             raise NotImplementedError("This geo level not tested yet.")
 
+    def test_sample_data_exists(self):
```
emma-nechamkin, line 189: the sample data only exists in the `data` folder and there is no secondary `tmp` folder. The `tmp` paths below are the temp files.

esfoobar-usds:
@emma-nechamkin looks like there's two outstanding issues:
- A lint issue
- The poetry lock is out of date

I'm checking this branch out locally to see if I can fix these.

esfoobar-usds (approved):
Awesome work @widal001 and @emma-nechamkin -- this is ready to merge!

🍏

Issue #1368, opened 2022-03-03
Add new Public Engagement opportunities page and button to the site 


BethMattern:
Tasks

- [x] create new page with component
- [x] see if USWDS component can support mock
- [x] create new button component (may want to refactor existing component) to support both button types
- [x] add button to each page
- [x] add test
- [x] test on mobile
- [ ] testing on staging
- [ ] testing on prod

Bugs:
- [x] p descendants error
- [x] Collection doesn't collapse on mobile
- [x] button doesn't swap on mobile 

Improvements
- [ ] page refreshes rather than linking, this might have to do with using the USWDS component

Deviations from mock?
- [ ] NEW is left aligned?
- [ ] Registration link doesn't wrap on mobile and happen to wrap on desktop because of the hyphen in the registration link. This would look better on mobile if we could make [Regstration Link](https://zoom.com) a link

KameronKerger:
![Engagement_calls_v2_info](https://user-images.githubusercontent.com/77702996/156639805-47bbee30-f6f4-4580-9bd5-947da1edac6b.png)
![Engagement_calls_header_v2](https://user-images.githubusercontent.com/77702996/156639812-4b3db3ef-85ba-4feb-aa7c-2af9ccf6dabc.png)

KameronKerger:
![march_9](https://user-images.githubusercontent.com/77702996/156650051-6c1d7f0b-552f-492b-ae5f-d8daa0dcc81b.svg)
![march_10](https://user-images.githubusercontent.com/77702996/156650054-1b728589-9125-490f
![march_9](https://user-images.githubusercontent.com/77702996/156650122-b4d20bc0-5ec2-4156-8b71-fb569e84ff6e.png)
![march_10](https://user-images.githubusercontent.com/77702996/156650125-ee0f948d-4cbe-4f9b-ae1d-6e365b191816.png)
![march_15](https://user-images.githubusercontent.com/77702996/156650128-4e138304-15d3-458c-92b4-cec401e17ae7.png)
-a75a-5e6757066cab.svg)
![march_15](https://user-images.githubusercontent.com/77702996/156650060-c73d4206-f064-4461-8970-c6b194bd353c.svg)

KameronKerger:
![march_9](https://user-images.githubusercontent.com/77702996/156650190-aa8caa1e-a284-4487-82c8-5eddf2e36d1e.png)
[...]

PR #1372, opened 2022-03-04
Public Engagement Page

vim-usds:
## Purpose
Add a new page to the site to allow for public engagement via events with zoom links

Closes: #1368 

## Deviations from the mock
1. the NEW tag is left aligned. When can change it, but seems better and also looks good on mobile.
2. Registration link doesn't wrap on mobile. The first link happens to wrap around on desktop because of the hyphen in the registration link. This would look better on mobile if we could make [Regstration Link](https://zoom.com/) a link to the zoom link rather than have the actual link on the page. Is there a reason we're placing the link explicitly?

## QA 
- [x] How does it look on desktop?
- [x] How does it look on mobile?
- [x] Do all the links work?
- [x] Are the links correct?

## Fast follow improvements
1. clicking the button refreshes the app rather than the fast-cache linking. Need to investigate more on why = 3 pointer
2. add a "past events" section to the page that automatically detects the event's date and determines if it's expired = 3 pointer
3. refactor EVENT object to have EVENT_DATES define all dates and the all month / day instances should call that object. This will help in the event a date changes. Right now it's hard-coded. (will be done within one of the above).

What a day! 😅


KameronKerger:
looks good. I'll do a proper mobile test in a bit. one issue so far is that the header on the page is en espano
Screen Shot 2022-03-04 at 7 47 09 AM
l

esfoobar-usds (approved):
🍏

PR #1382, opened 2022-03-07
Download Packet Changes Round 3

### LINKED ISSUE #1215: Download Packet Changes Round 3
@KameronKerger requested the following changes to the download packet:

- [x] GEOID10_TRACT should be Census tract ID
- [ ] Can the alignment of all the fields be consistent? headers centered, words centered, numerical values right aligned
- [x] Can we change "Total threshold criteria exceeded" to "Total number of at or above thresholds criteria" so it is consistent with the site content?
- [ ] Can we change low college attendance to high college enrollment? I'll bring this up during score M discussions today but "attendance" makes it seem like people aren't going to school... this should be done already?
- [ ] I think the FEMA data is just "NationalRisk Index" and not "Natural Hazards Risk Index"
- [ ] We should spell out the Superfund acronym since we do the others. it's National Priorities List
- [ ] For asthma we can probably remove "Current" so it just starts with "Asthma"
- [ ] Same for "Diagnosed diabetes" so just "Diabetes" (this will make it more consistent with the site
- [ ] Overall can we make the columns a little smaller? Do if it is at like 30 make it more of a 20

katherinedm-usds:
Fixes #1215

### DIFF
--- a/data/data-pipeline/data_pipeline/content/config/csv.yml
+++ b/data/data-pipeline/data_pipeline/content/config/csv.yml
@@ -153,10 +153,10 @@ fields:
     label: Greater than or equal to the 90th percentile for asthma, is low income, and has a low percent of higher ed students?
     format: bool
   - score_name: Current asthma among adults aged greater than or equal to 18 years (percentile)
-    label: Current asthma among adults aged greater than or equal to 18 years (percentile)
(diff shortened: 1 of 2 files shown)


esfoobar-usds:
Legacy changes.

PR #1386, opened 2022-03-08
Reducing Docker start up and adding ArcGIS URL

### LINKED ISSUE #1362: For Dockerfile CMD generate map tiles using S3 score
We want to make the initial local install as easy and quick as possible, so we want to make the Dockerfile CMD to be similar to the GHA combine and tilefy which downloads the latest score from S3, avoiding the network timeout issues some users have experienced when doing the `docker-compose up`.

esfoobar-usds:
Fixes #1362.

This PR reduces the brand-new Docker setup for users from more than an hour to around 30 minutes and also avoids issues with external data sources rate limiting or networking issues, relying fully on our S3-based score.

The PR also adds the ArcGIS shape files URL.

### DIFF
--- a/data/data-pipeline/README.md
+++ b/data/data-pipeline/README.md
@@ -83,7 +83,7 @@ Once we have all the data from the previous stages, we convert it to tiles to ma
 
 #### 5. Shapefiles
 
-If you want to use the shapefiles in mapping applications, you [can download them here](https://justice40-data.s3.amazonaws.com/data-pipeline/data/score/shapefile/).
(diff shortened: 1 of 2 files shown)


vim-usds (approved):
Looks great!

crvanpollard:
@esfoobar-usds It looks like you are missing the .dbf file type for users to work with the ArcGIS shapefile. Could you please provide the .dbf along with the .shp and .shx file types.
Thanks for such an important tool and datasets.
feel free to reach out - Chris Pollard (cpollard@dvrpc.org)

esfoobar-usds:
Thanks so much @crvanpollard. We are going to create another ticket that will instead create a zip file with all the ArcGIS files. You can track it [here](https://github.com/usds/justice40-tool/issues/1388).

PR #1387, opened 2022-03-09
Updating higher education to be reversed

emma-nechamkin:
__Summary__ In this PR, we create a new variable so that the % college students is expressed as % not college students. This means that the front end can display % not college students. 

__Backend changes__ On the back end, this has introduced a new variable:
```
COLLEGE_NON_ATTENDANCE_FIELD = (
    "Percent of population not currently enrolled in college or graduate school"
)
```

This is defined as `(1-College Attendance)`. We don't need to add anything else new from the perspective of calculating the score, but I did make a change: 

I changed this calculation to be strictly less than:
```
self.df[field_names.COLLEGE_ATTENDANCE_LESS_THAN_20_FIELD] = (
            self.df[field_names.COLLEGE_ATTENDANCE_FIELD]
            < self.MAX_COLLEGE_ATTENDANCE_THRESHOLD
        )
```
Because we are moving this variable to be in the same direction as all other variables, we change this to be < rather than <=. This translates to "80% or more of residents are not college students", rather than "Strictly greater than 80% of residents are not college students." There are two tracts that are impacted by this (that is, they have exactly 20% college students) -- neither of these has been a DAC under any score.

I also tried to update the yaml file for the downloadable data. 

__Front end changes__ We can still use the same boolean variable (`CA_LT20`), which is logically equivalent to having greater than 80% non college students. The new variable to pipe in is `NCA` and it represents the percentage of non college students in the tract. I have not removed `CA`, the current variable, from the data (at least until we get the greenlight to reverse this threshold).


data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -443,6 +443,7 @@ def _prepare_initial_df(self) -> pd.DataFrame:
             field_names.UNEMPLOYMENT_FIELD,
             field_names.MEDIAN_HOUSE_VALUE_FIELD,
             field_names.COLLEGE_ATTENDANCE_FIELD,
+            field_names.COLLEGE_NON_ATTENDANCE_FIELD,
```
esfoobar-usds, line 446: I hope some day we can get out of this pattern of the floats!  😔

esfoobar-usds (approved):
🍏

Issue #1389, opened 2022-03-09
Update BE signals to support new side panel


vim-usds:
The new side panel design allows us to separate all indicators. For example, Low income is just low income and not low income and higher ed enrollment. In the previous UI design, these signals were combined. This ticket will separate these combined signals to allow each indicator to solely be based on weather or not it exceeds the threshold.

PR #1393, opened 2022-03-09
Adding booleans for FE to display 

emma-nechamkin:
Added a series of indicators to the tile data (and score data) that represent whether each individual criterion is exceeded. These are documented on the shared drive. 

Even though my branch name is 1311, this is actually #1389.


data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -166,32 +165,31 @@
...
-    field_names.LOW_LIFE_EXPECTANCY_LOW_INCOME_LOW_HIGHER_ED_FIELD: "LLELI",
-    field_names.LINGUISTIC_ISOLATION_LOW_HS_LOW_HIGHER_ED_FIELD: "LILHSE",
-    field_names.POVERTY_LOW_HS_LOW_HIGHER_ED_FIELD: "PLHSE",
-    field_names.LOW_MEDIAN_INCOME_LOW_HS_LOW_HIGHER_ED_FIELD: "LMILHSE",
-    field_names.UNEMPLOYMENT_LOW_HS_LOW_HIGHER_ED_FIELD: "ULHSE",
-    field_names.THRESHOLD_COUNT: "TC",
-    field_names.ISLAND_AREAS_UNEMPLOYMENT_LOW_HS_EDUCATION_FIELD: "IAULHSE",
-    field_names.ISLAND_AREAS_POVERTY_LOW_HS_EDUCATION_FIELD: "IAPLHSE",
-    field_names.ISLAND_AREAS_LOW_MEDIAN_INCOME_LOW_HS_EDUCATION_FIELD: "IALMILHSE",
-    field_names.ISLAND_AREAS_LOW_HS_EDUCATION_FIELD: "IALHE",
```
emma-nechamkin, line 194: I moved this to after the field definitions, I did not delete it

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -564,6 +564,63 @@
     f"Low high school education{ISLAND_AREAS_SUFFIX}"
 )
 
+# Booleans for percentile exceeded thresholds
```
emma-nechamkin, line 567: These are all new field names for these booleans. 

Field names is currently pretty clunky but I think we can wait to the refactor to streamline it :)

esfoobar-usds (approved):
LGTM! 🍏

vim-usds:
@emma-nechamkin - testing locally looks good! 

Next up is testing on staging. Once that's done, I'll open it up to general QA testing. Once that passes, we can merge. 🎈 

Thanks for the super fast turn around on this!

emma-nechamkin:
As I mentioned in stand up -- I deliberately didn't add these columns to the downloadable data. On the one hand, I'm worried that single thresholds absent context might be not user friendly ("this tract meets criterion x, why isn't it a DAC?") and I feel that including the columns isn't super important to analysis on the thresholds. On the other hand, I worry that a user might want to replicate what he or she sees on the tool. In that case, having the composite booleans would be useful. Right now, I've leaned towards the former (we show percentiles already), but if you have strong feelings either way -- just let me know!

vim-usds (approved):
LGTM! Nice work!

PR #1399, opened 2022-03-11
Enabling FormatJS linter rules to catch i18n issues

NatHillardUSDS:
See https://formatjs.io/docs/tooling/linter/ for more info


vim-usds:
Adding this in another branch

PR #1409, opened 2022-03-15
Quick fix: updating snapshots to have more sigfigs

emma-nechamkin:
I probably should have done this in a hot fix. 10 places after decimal for all snapshots.


data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -229,7 +229,7 @@ def test_extract_produces_valid_data(self, snapshot, mock_etl, mock_paths):
         )
         snapshot.snapshot_dir = self._DATA_DIRECTORY_FOR_TEST
         snapshot.assert_match(
-            tmp_df.to_csv(index=False, float_format="%.5f"),
+            tmp_df.to_csv(index=False, float_format="%.10f"),
```
lucasmbrown-usds, line 232: Nit: maybe just make this a constant, since we re-use it so often?

lucasmbrown-usds (approved):
Soooo nice to see how smoothly these update! I love it.

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
emma-nechamkin: Sure!

PR #1410, opened 2022-03-15
S3 Parallel Upload and Deletions

### LINKED ISSUE #1246: Redownload FIPS file on every full etl-run
We've had some issues with the caching of the FIPS state zip file. To make sure this is avoided, we should ensure the file is cleared and downloaded every time the etl-run is initiated.

### LINKED ISSUE #1319: As an app developer, I want the uploads to S3 and deletions of S3 files to be parallelized, so I can reduce the app deploy times by up to 20 minutes.
**Description**
The parts of the full score and tiles deployment process that take the longest are: 

1. Deleting all previous tiles from S3 
2. Uploading all new tiles to S3 

This takes a long amount of time (20 minutes?). 

There are a few well-known solutions to addressing this: https://stackoverflow.com/a/53336858/1569221.  In that example they took upload times from 10 minutes to 18 seconds. There might be some other known approaches. 

Also as part of this PR, will add invalidation of backend cloudfront

esfoobar-usds:
Fixes #1319
Fixes #1246


vim-usds (approved):
Now I understand that cache hit! So it checks if a poetry is available locally and if so, use it rather than installing it! Nice!

esfoobar-usds:
> Now I understand that cache hit! So it checks if a poetry is available locally and if so, use it rather than installing it! Nice!

Yes! I can (and want) go over it in detail because it's a nice caching pattern that can be used for anything where packages are involved, [including node](https://stackoverflow.com/questions/67136614/cache-node-modules-in-github-actions).

PR #1416, opened 2022-03-16
Add intl pre-commit hook and update GHA

### LINKED ISSUE #1373: Add a note to a github action that translations are needed when en.json changes
The current plan is to 

1. Use a pre-commit hook to detect changes on the data / copy folder and if so run the en.json extract command
2. In the Github action check for this file change and flag a message that translations are needed

vim-usds:
# Purpose
- closes #1373 
- will add Husky lib
- will add pre-commit script that will run `intl:extract`, detect any `warnings` on extraction and abort commit
- update deploy staging yml to show message that en.json was modified


esfoobar-usds (approved):
🍏

Issue #1327, opened 2022-02-23
Create marshmallow schema for YAML files to standardize and better control content configuration


esfoobar-usds:
Right now there is no schema for the content/config YAML files and we are using constants to establish some of it. It would be better to define a schema for the YAML and add validation tests.

PR #1469, opened 2022-03-22
Creating a data dictionary for the download packet

emma-nechamkin:
Goal is to automatically generate a codebook for the data.


data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,179 @@
+"""
+Quick script to generate a data dictionary for the download packet.
```
emma-nechamkin, line 2: i haven't thought much about where this script goes.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,179 @@
...
+LOSS_RATE_STRING = "loss rate"
+LOW_STRING = "Low "
+ISLAND_STRING = "island areas"
+
+PERCENTILE_EXPLANATION = "All percentiles are floored (rounded down to the nearest percentile). For example, 89.7th percentile is rounded down to 89 for this field."
+LOW_PERCENTILE_EXPLANATION = "This percentile is reversed, meaning the lowest raw numbers become the highest percentiles."
+ISLAND_AREAS_EXPLANATION = "Because not all data is available for the Nation, Puerto Rico, and the Island Territories, this uses different underlying data for the islands."
+
+
+OUTPUT_PATH = (
```
emma-nechamkin, line 56: This seems right to me, in that it's close to the downloadable stuff and can be zipped. What do you guys think?

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,179 @@
+"""
+Quick script to generate a data dictionary for the download packet.
+
+A few notes:
+(1) This will ONLY crosswalk the downloadable CSV and the downloadable EXCEL file.
```
emma-nechamkin, line 5: If we want to make a shp dictionary (beyond -- "here's the column and what it means in plain English") we will need to do it separately since the shapefiles have more fields.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
esfoobar-usds: Yup, this is a good destination for it, although we might need to add it to the zip files list in score post

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+def _get_calculation_notes(column_name):
+    """produces calculation notes"""
+    calculation_notes = []
+    if PERCENTILE_STRING in column_name:
+        calculation_notes += [PERCENTILE_EXPLANATION]
+    if LOW_STRING in column_name:
+        calculation_notes += [LOW_PERCENTILE_EXPLANATION]
+    if ISLAND_STRING in column_name:
+        calculation_notes += [ISLAND_AREAS_EXPLANATION]
+    return " ".join(calculation_notes)
```
esfoobar-usds, line 138: We should add these as types on the YAML config rather than deducing from the column name?

esfoobar-usds (changes requested):
Some feedback to address.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+    """helper function to create a dataframe from yaml"""
+    to_frame_dict, fields = _set_up_infrastructure(
+        yaml_path=yaml_path,
+        fields_to_store=fields_to_store,
+    )
+    for details in fields:
+        to_frame_dict[CEJST_SCORE_COLUMN_NAME].append(
+            details[native_score_column]
+        )
+        for column_name, dict_field in fields_to_store:
```
lucasmbrown-usds, line 105: Some general comments about this logic would be really helpful -- to be honest I'm having a little bit of a hard time following the logic.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+                to_frame_dict[column_name].append(np.nan)
+    return pd.DataFrame(to_frame_dict).set_index(CEJST_SCORE_COLUMN_NAME)
+
+
+def _get_datatype(
+    input_name: str,
+    input_type: str,
+    percentile_string: str = PERCENTILE_STRING,
+    loss_rate_string: str = LOSS_RATE_STRING,
+):
```
lucasmbrown-usds, line 119: nit: always valuable to type the return as well, e.g., `) -> str`.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+        native_score_column=CEJST_SCORE_COLUMN_NAME,
+    )
+
+    score_detail_notes_df = _create_df_from_yaml(
+        yaml_path=SCORE_FIELD_CONFIG,
+        fields_to_store=[
+            (NOTES_FIELD, "notes"),
+            (THRESHOLD_CATEGORY_FIELD, "category"),
+        ],
+        native_score_column="column_name",
```
lucasmbrown-usds, line 161: I gotta be honest, I'm scratching my head a bit here about what all of this is doing. Could you add some comments making it a bit more obvious why this is `"column_name"`?

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
lucasmbrown-usds: Jorge's suggestion makes a lot of sense to me! Could put booleans on the YAML config, like, 

```
is_low_field: True
is_island_areas_field: True
```

etc

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+    if PERCENTILE_STRING in column_name:
+        calculation_notes += [PERCENTILE_EXPLANATION]
+    if LOW_STRING in column_name:
+        calculation_notes += [LOW_PERCENTILE_EXPLANATION]
+    if ISLAND_STRING in column_name:
+        calculation_notes += [ISLAND_AREAS_EXPLANATION]
+    return " ".join(calculation_notes)
+
+
+if __name__ == "__main__":
```
lucasmbrown-usds, line 141: I wonder if we could wrap all of this overall logic into a single function, and then calling that function part of `generate_score_post` in `application.py` so that this file gets automatically generated along with the rest of the downloads whenever we run those steps as part of existing processes? Would be easier for others to use than remembering to one-off run this script.

data/data-pipeline/data_pipeline/content/config/score_fields.yml
```diff
@@ -0,0 +1,56 @@
+# This is a temporary file. We should make sure this *type* of information is maintained when we refactor. 
+fields:
+    - column_name: Total threshold criteria exceeded
+      notes: Lists out the total number of criteria (where each category has one or more criteria) exceeded. For example, a tract that exceeds the 90th percentile for linguistic isolation (1) and unemployment (2), and meets the training and workforce development socioeconomic criteria (high school attainment rate and low percentage of higher ed students) would have a 2 in this field. 
+    - column_name: Definition M (communities)
```
lucasmbrown-usds, line 5: Should this be `Identified as disadvantaged (v0.1)` or whatever?

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
lucasmbrown-usds: It can also be helpful to put in some nice logger statements so that when someone is running this from the command line, they can "follow along". 

E.g., ``logger.info(f"Loading CSV configuration from file path `{CSV_CONFIG}`.")``. Same for Excel and Score. And then ``logger.info(f"Writing data dictionary to file path `{OUTPUT_PATH}`.")``. These loggers are helpful in debugging later and understanding what was done when.

lucasmbrown-usds (changes requested):
Looking great that there's an automated way to do this and that the inputs are easily trackable in the YAML. 

I'm wondering if perhaps there's a simpler way to approach this task in the code. I don't have an immediately obvious suggestion, but I am finding the code rather hard to follow for what is mostly a straightforward task. Maybe now that you have some experience with the inputs and outputs of this function, you can brainstorm a little bit the simplest, cleanest way to do the transformation? I'm mostly worried about how easy it would be for any developer / data scientist to understand and update this code.

I could also just be brain foggy and not following the logic which is actually fairly clear! Happy to discuss further.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+    settings.APP_ROOT / "data" / "score" / "downloadable" / "codebook.csv"
+)
+
+
+def _set_up_infrastructure(
+    yaml_path: pathlib.Path,
+    fields_to_store: list,
+):
+    """helper function to parse yaml"""
+    yaml_contents = yaml.safe_load(
```
lucasmbrown-usds, line 72: Oh, and it doesn't have to be in this PR, but at some point we should add some safety here by loading from YAMLs using a schema like Marshmallow to enforce that the YAML has the right format, acceptable values, etc. Probably part of the Big Refactor?

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
emma-nechamkin: I was hoping that our new field names after the refactor would have programmatic column names that would make this clear, but I'm happy to put this in the yaml!

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
emma-nechamkin: this makes a lot of sense to me! The code is pretty quick on my machine, so I didn't think to add logging but am happy to do so. Similarly, I like the idea of adding it to `generate_score_post`!

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+    if ISLAND_STRING in column_name:
+        calculation_notes += [ISLAND_AREAS_EXPLANATION]
+    return " ".join(calculation_notes)
+
+
+if __name__ == "__main__":
+    # read component yamls
+    csv_codes_df = _create_df_from_yaml(
+        yaml_path=CSV_CONFIG,
+        fields_to_store=[(CSV_LABEL_FIELD, "label"), (CSV_FORMAT, "format")],
```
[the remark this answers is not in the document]
emma-nechamkin: I should have known you'd want a namedtuple 😂 Happy to implement.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
emma-nechamkin: Yes! I can also just change it in the yaml.

data/data-pipeline/data_pipeline/content/config/score_fields.yml
emma-nechamkin: It's the native field, so the one in usa.csv, not the downloaded field.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+    + "and the Island Territories, this uses different underlying data for the islands."
+)
+
+
+OUTPUT_PATH = (
+    settings.APP_ROOT / "data" / "score" / "downloadable" / "codebook.csv"
+)
+
+
+def _set_up_infrastructure(
```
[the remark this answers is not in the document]
emma-nechamkin: Sounds good to me!

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
+"""
+Quick script to generate a data dictionary for the download packet.
+
+A few notes:
+(1) This will ONLY crosswalk the downloadable CSV and the downloadable EXCEL file.
+    The SHP file has a LOT more columns, and we should almost certainly
+    make a data dictionary by hand once the column names are final.
```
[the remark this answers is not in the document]
emma-nechamkin: Well, I'm not sure it makes sense to give a much more robust codebook for the shapefile than the one we already provide (which has plain english and the column name). But, if we do, and we wanted to automate, it would probably make sense to wait until after the refactor.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+    CEJST_SCORE_COLUMN_NAME,
+    CSV_FIELD_TYPE_FIELD,
+    CSV_LABEL_FIELD,
+    EXCEL_LABEL_FIELD,
+    CALCULATION_NOTES_FIELD,
+    THRESHOLD_CATEGORY_FIELD,
+    NOTES_FIELD,
+]
+
+# for re-typing & notes
```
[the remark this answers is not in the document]
emma-nechamkin: updates

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
```diff
@@ -0,0 +1,185 @@
...
+# note this doesn't perfectly square with field names
+PERCENTILE_STRING = field_names.PERCENTILE_FIELD_SUFFIX
```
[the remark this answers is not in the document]
emma-nechamkin: nope! updated

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
emma-nechamkin: Done!

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
emma-nechamkin: oh yes! ty!

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
esfoobar-usds: Yup, this is this ticket #1327

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
esfoobar-usds: I think we can punt to the refactor, in the interest of merging this PR.

emma-nechamkin:
one general note here (@esfoobar-usds and @lucasmbrown-usds) -- this relies on having the current names as long form strings. After the refactor, we should take the current names and join on them as the "description" field, like a typical codebook might have.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
emma-nechamkin: Not taking action here.

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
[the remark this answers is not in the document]
emma-nechamkin: To address this -- 
> I'm wondering if perhaps there's a simpler way to approach this task in the code. I don't have an immediately obvious suggestion, but I am finding the code rather hard to follow for what is mostly a straightforward task. Maybe now that you have some experience with the inputs and outputs of this function, you can brainstorm a little bit the simplest, cleanest way to do the transformation? I'm mostly worried about how easy it would be for any developer / data scientist to understand and update this code.

I tried to simplify this a little bit, largely by putting it in `generate_score_post` and leveraging some of our existing utilities / code. So it's a bit dryer. I also tried to update the variable names / comments to make the operations clearer. 

The code itself is quite simple. Please let me know if these changes make it more understandable.

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -54,10 +54,44 @@
 SCORE_DOWNLOADABLE_EXCEL_FILE_PATH = (
     SCORE_DOWNLOADABLE_DIR / f"communities-{timestamp_str}.xlsx"
 )
+SCORE_DOWNLOADABLE_CODEBOOK_FILE_PATH = (
+    SCORE_DOWNLOADABLE_DIR / f"codebook-{timestamp_str}.csv"
+)
```
emma-nechamkin, line 59: This now gets generated with the rest of the zip packet

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -54,10 +54,44 @@
...
 SCORE_DOWNLOADABLE_ZIP_FILE_PATH = (
     SCORE_DOWNLOADABLE_DIR / "Screening_Tool_Data.zip"
 )
 
+# For the codebook
```
emma-nechamkin, line 64: These also get referenced from constants.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -448,7 +448,7 @@ def _load_score_csv_full(
 
     def _load_excel_from_df(
         self, excel_df: pd.DataFrame, excel_path: Path
-    ) -> None:
+    ) -> dict:
```
emma-nechamkin, line 451: I slightly changed this so that I would not have to redo the i/o BUT am happy to revert it

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -534,10 +536,29 @@ def _load_downloadable_zip(self, downloadable_info_path: Path) -> None:
         )
         downloadable_df.to_csv(csv_path, index=False)
 
+        logger.info("Creating codebook for download zip")
+        extra_score_info = load_yaml_dict_from_file(
+            self.CONTENT_CONFIG / "score_fields.yml"
```
emma-nechamkin, line 541: set identically to the csv.yml and excel.yml code

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -1,6 +1,9 @@
+from cmath import sin
 import os
 import sys
 from pathlib import Path
+from collections import namedtuple
+from isort import code
```
emma-nechamkin, line 6: I do not know what isort is?

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +113,167 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
     )
 
     return floored_series
+
+
+def _create_df_from_yaml_contents(
```
emma-nechamkin, line 118: I put this in `etl_utils` because that seemed like the best place for it. Happy to move it though, if you disagree!

esfoobar-usds (approved):
🍏 -- great job!

esfoobar-usds:
@emma-nechamkin the only thing missing is the addition of the data dictionary file to the zip payload?

emma-nechamkin:
@esfoobar-usds and @lucasmbrown-usds -- I believe this tox error is because I used `|` to join dictionaries. This is not part of python 3.8, but I think it's so elegant 😢 . Updated, but just wanted to confirm that was the right call. 

```
>       codebook_dictionary = {constants.CEJST_SCORE_COLUMN_NAME: []} | {
            field.new_label_in_codebook: [] for field in fields_to_store_in_codebook
        }
E       TypeError: unsupported operand type(s) for |: 'dict' and 'dict'
```

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
emma-nechamkin: One question I have now that this is in `generate_score_post` is if we need tests for it? When it was a one-off thing, I was less concerned about it breaking. Here, I'd envision the test being a snapshot, but a snapshot of the whole codebook. @lucasmbrown-usds wdyt?

lucasmbrown-usds: I always think we need tests, obviously! Would those be easy to add? Or you can create a backlog ticket for it, but as mentioned on the other PR, tickets for tests alone can be a real drag to add at a later date. 

Snapshot of the whole codebook makes sense to me! And would be v easy to update with our snapshot code.

As for logging, regardless of how long it takes to run they can be helpful on a remote server for pinpointing how far a process got before it crashed, and can be helpful to someone reading that output to know exactly what is generally happening under the hood during a process even when it's successful.

data/data-pipeline/data_pipeline/content/config/score_fields.yml
lucasmbrown-usds: oh i see

data/data-pipeline/data_pipeline/content/config/generate_data_dictionary.py
emma-nechamkin: I added an issue to the product backlog for now; I'll aim to take care of it sooner rather than later.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
lucasmbrown-usds: Seems like a good home!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
+    fields_list_from_yaml: list,
+    fields_to_store_in_codebook: list,
+) -> pd.DataFrame:
+    """Helper function to create a dataframe from yaml fields
+
+    This function does:
+        1. Creates a dictionary that will eventually become a dataframe.
```
lucasmbrown-usds, line 122: when is "eventually"?

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
     return floored_series
+
+
+def _create_df_from_yaml_contents(
+    fields_list_from_yaml: list,
+    fields_to_store_in_codebook: list,
+) -> pd.DataFrame:
+    """Helper function to create a dataframe from yaml fields
+
+    This function does:
```
lucasmbrown-usds, line 121: I'm sorry, but I'm having a hard time understanding your steps 1-3 in this documentation. Do you mind rewriting this to try to be more clear? There's a lot that seems a little handwavey in your explanation.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
+        1. Creates a dictionary that will eventually become a dataframe.
+           This dictionary has columns dictated by whichever fields
+           the yaml file has that we want to keep, specified by fields_to_store.
+        2. Goes through the fields and adds each type of field to the dictionary. If
+           the field is missing, appends a null value.
+        3. Returns a dataframe that includes all of the fields specified and
+           also is indexed by the column name that is native to the CEJST team
```
lucasmbrown-usds, line 128: "the column name that is native to the CEJST team" - this feels like a very imprecise way of saying what you're trying to say?

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
+    """
+    # this becomes the codebook frame for each  yaml source. In particular,
```
lucasmbrown-usds, line 130: helpful!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
+    downloadable_csv_config: dict, excel_config: dict, extra_score_info: dict
+) -> pd.DataFrame:
+    """Runs through all logic of creating the codebook.
+
+    First it reads in each component yaml file for the codebook.
+    Then it merges all of them.
+    Finally, it applies any transforms to the columns (like getting the
+        datatype or adding calculation_notes.
+    """
+    CodebookLabelFields = namedtuple(
```
lucasmbrown-usds, line 210: 🎉

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
     )
 
     return floored_series
+
+
+def _create_df_from_yaml_contents(
+    fields_list_from_yaml: list,
+    fields_to_store_in_codebook: list,
+) -> pd.DataFrame:
+    """Helper function to create a dataframe from yaml fields
```
lucasmbrown-usds, line 119: Maybe also add note that it'll be used by three different configs.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
lucasmbrown-usds: Would definitely be good to test this function.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
+        fields_list_from_yaml=excel_config,
+        fields_to_store_in_codebook=[
+            CodebookLabelFields(
+                new_label_in_codebook=constants.EXCEL_LABEL_FIELD,
+                existing_yaml_label="label",
+            )
+        ],
+    )
+
+    score_detail_notes_df = _create_df_from_yaml_contents(
```
lucasmbrown-usds, line 240: on the hard problem of naming things, consistency wherever possible is essential. you've referred to the information in `extra_score_info` by three different names: `score_fields.yml`, `extra_score_info`, and `score_detail_notes_df`. 

you can instead have consistency like `field_descriptions_for_codebook.yml`, `field_descriptions_for_codebook`, and `field_descriptions_for_codebook_df` that will make it much more obvious what's happening

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
+    codebook_dictionary[constants.CEJST_SCORE_COLUMN_NAME] = []
+
+    # we reshape the data from a list of dictionaries to a dictionary of lists
+    # so that we can cast it as a dataframe
+    for single_field_details in fields_list_from_yaml:
+        assert constants.CEJST_SCORE_COLUMN_NAME in single_field_details, (
+            "Error: the yaml codebook should crosswalk to the native column "
+            + f"from the CEJST pipeline, called {constants.CEJST_SCORE_COLUMN_NAME}"
+        )
+        codebook_dictionary[constants.CEJST_SCORE_COLUMN_NAME].append(
```
lucasmbrown-usds, line 146: Could you add a comment about what's happening here?

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
+            single_field_details[constants.CEJST_SCORE_COLUMN_NAME]
+        )
+        for field_information in fields_to_store_in_codebook:
+            try:
```
lucasmbrown-usds, line 150: Could you add a comment about what's happening here? Why is there a try/except?

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -108,3 +110,174 @@ def floor_series(series: pd.Series, number_of_decimals: int) -> pd.Series:
...
+    fields_to_store_in_codebook: list,
+) -> pd.DataFrame:
+    """Helper function to create a dataframe from yaml fields
+
+    This function does:
+        1. Creates a dictionary that will eventually become a dataframe.
+           This dictionary has columns dictated by whichever fields
+           the yaml file has that we want to keep, specified by fields_to_store.
+        2. Goes through the fields and adds each type of field to the dictionary. If
+           the field is missing, appends a null value.
```
lucasmbrown-usds, line 126: why?

lucasmbrown-usds (dismissed):
Lots of little comments. I have to admit, I'm struggling with this PR a bit still. If I were assigned to update this code, I'm not sure I understand what's happening in all these methods enough to understand how to update the code. It seems like there are a ton of one-off exceptions, if/else logic, etc. 

It feels to me like the inputs and outputs of these functions are a little bit convoluted. I feel like it's only possible to understand how one piece of this works if you understand how all the pieces of it work -- but that means I can't build up an understanding of the whole thing when I'm stuck on each individual module. 

I'm wondering if it's possible to redesign this so that the inputs and outputs of each function are really clean and clear. 

At the same time, this is just the codebook, so we could also create a ticket to consider coming back and refactoring later if we just want to move ahead and merge it.

data/data-pipeline/data_pipeline/content/config/score_fields.yml
```diff
@@ -0,0 +1,56 @@
+# This is a temporary file. We should make sure this *type* of information is maintained when we refactor.
```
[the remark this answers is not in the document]
emma-nechamkin: This makes sense to me, but I wonder what @esfoobar-usds thinks?

emma-nechamkin:
A few high-level thoughts:

> Lots of little comments. I have to admit, I'm struggling with this PR a bit still. If I were assigned to update this code, I'm not sure I understand what's happening in all these methods enough to understand how to update the code. It seems like there are a ton of one-off exceptions, if/else logic, etc.

Happy to address the little comments, but I'm not really sure how to address the larger point. The code in our codebase for parsing yaml files has a lot of if / else statements -- it seems like they are somewhat impossible to avoid? I'm honestly not really sure how to make this simpler... there are four functions: one to create a dataframe from the contents of each yaml, one to return the datatype of the column based on the column name, one to programmatically add notes to each row based on the column name (which will eventually change when we refactor, I assume!), and then one to produce the codebook from three yaml files. I certainly don't disagree with you that it's a little convoluted, but given that we are essentially just pivoting three yaml files, I'm not really sure how it *wouldn't* be convoluted -- and, at a higher level, even if we had a single yaml file with the column long name, the column short name, the csv name, the excel name, the datatype, and any notes, I'm not actually convinced it would be less convoluted unless we explicitly specified everything, and if we explicitly specified everything, I don't understand why we wouldn't just manually create a codebook?

> It feels to me like the inputs and outputs of these functions are a little bit convoluted. I feel like it's only possible to understand how one piece of this works if you understand how all the pieces of it work -- but that means I can't build up an understanding of the whole thing when I'm stuck on each individual module.
> 
> I'm wondering if it's possible to redesign this so that the inputs and outputs of each function are really clean and clear.

So I simply do not understand what you're suggesting. Did you have an example of something specifically that isn't clear? I don't mean to be particularly truculent, but I'm not sure I understand the assignment here. I don't think this code is perfect, but if I saw a way to make it cleaner, I would have made it cleaner when I updated it. 

> At the same time, this is just the codebook, so we could also create a ticket to consider coming back and refactoring later if we just want to move ahead and merge it.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: because we want the row to be blank

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: yes

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: Yup!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: 🎉 sounds good and makes sense!

data/data-pipeline/data_pipeline/content/config/score_fields.yml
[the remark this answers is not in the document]
emma-nechamkin: This makes sense to me :)

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -534,10 +536,27 @@ def _load_downloadable_zip(self, downloadable_info_path: Path) -> None:
         )
         downloadable_df.to_csv(csv_path, index=False)
 
+        logger.info("Creating codebook for download zip")
+        extra_score_info = load_yaml_dict_from_file(
```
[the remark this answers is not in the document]
emma-nechamkin: updated

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -534,10 +536,27 @@ def _load_downloadable_zip(self, downloadable_info_path: Path) -> None:
...
+            self.CONTENT_CONFIG / "score_fields.yml"
+        )
+
+        excel_fields = []
```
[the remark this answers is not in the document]
emma-nechamkin: done

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: done

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: Updated

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: at the end of the functioin. updated comment.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: updated

PR #1486, opened 2022-03-24
Adding category count to BE signals 

emma-nechamkin:
New frontend design means we now show both threshold count and category count. I constructed a new variable, category count, for this. I have also added it to the downloadable csv and excel files.

Variable is correct and makes it to the `tiles.csv` file. 
Screen Shot 2022-03-24 at 1 07 24 PM


vim-usds:
LGTM 🍏 🚀 🌟

KameronKerger:
🚀

esfoobar-usds (approved):
🍏

PR #1492, opened 2022-03-25
Add wording to UI that calls out improvements to display of census tract information

vim-usds:
# Purpose
Bring attention to the new census track side panel

### QA reviewer
- [x] Kameron

### QA Spec
- [x] Is the Alert's vertical alignment good?
- [x] Is the Alert's horizontal alignment good?
- [x] Is the Alert's copy correct?
- [x] Does the Alert render on mobile correctly?


KameronKerger:
QA passed

PR #1495, opened 2022-03-25
Update sidepanel non-selected copy

### LINKED ISSUE #1359: Add new wording to the map panel that appears upon load when no tract is selected
Kameron has new wording & design for this area on the map.

vim-usds:
# Purpose
- Modify non-selected side panel based on CEQs latest updates
- closes #1359 

### QA tester
- [x] Kameron

### QA Spec
- [x] Things to know
- [x] This tool identifies communities that are marginalized, underserved, and overburdened by pollution. These communities are located in census tracts that are at or above the thresholds in one or more of eight categories of criteria.
- [x] The tool uses census tracts that represent about 4,000 people, which is the smallest unit of geography for which consistent data can be displayed on the tool.
- [x] The tool ranks each census tract using percentiles that show how much burden each tract experiences relative to all other tracts, for each criteria.
- [x] Percentages are used for certain variables, i.e. those relating to high school achievement rate and to the share of individuals not currently enrolled in higher education.
- [x] Thresholds for each category determine if a tract should be identified as disadvantaged because it has exceeded a certain value for the relevant indicators.

### DIFF
--- a/.github/workflows/deploy_fe_staging.yml
+++ b/.github/workflows/deploy_fe_staging.yml
@@ -55,7 +55,7 @@ jobs:
           working-directory: ${{env.WORKING_DIRECTORY}}
           browser: chrome
           start: npm start
-          wait-on: "http://localhost:8000"
+          wait-on: "http://localhost:8000/en/"
       - name: Upload Artifact
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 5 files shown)


KameronKerger:
QA passed. Editors note: there is a weird comma but apparently that is what CEQ wants.

PR #1512, opened 2022-03-30
Modify data tile server to static URL to fix CORS

vim-usds:


vim-usds:
> That was fast! 🍏

When I realized it was set as an env, I thought.. "ohh a one-liner! I've heard about these fixes.. Let's just knock it out" 🤣

vim-usds:
@jorge - this staging link is loading from the new static URL:

Screen Shot 2022-03-30 at 12 27 18 PM

Can you validate on your end?

vim-usds:
@jorge - how does the GHA action look?

esfoobar-usds (approved):
🍏 looks good!

PR #1524, opened 2022-04-01
Add video to PEC

vim-usds:
# Purpose
Add video link to PEC

### QA spec
- [x] How is the light blue background?
- [x] How is the yellow button?
- [x] How is the box width?
- [x] Is the 8:4 column satisfactory?
- [x] how is the placement ? (lower than mock due to a11y failure in original location). We can get it on the original position, it's just that we'll have to create a custom component (rather than using the SummaryBox which is provided by USWDS). Custom component shouldn't be too long if that is desired.


esfoobar-usds (approved):
🍏

Issue #1518, opened 2022-03-31
Additional YAML based score datasets


BethMattern:
- [ ] DOE
- [ ] COI

mattbowen-usds:
This was closed with #1796

Issue #1221, opened 2022-02-02
As a data developer, I want to understand how much tile space is saved by short names.


lucasmbrown-usds:
**Description**
Run an experiment with and without shortnames on the tiles in order to determine whether this step is really saving us much size on the tiles. 

Generate the tiles and check their file size.

widal001:
@widal001 tagging myself on this so I remember to follow up on it

widal001:
@lucasmbrown-usds and @esfoobar-usds Thanks for your patience as I carved out some time to dig into this. Before running some tests I wanted to make sure I was understanding the goal and the portion of the related to this issue. I've summarized my current best understanding of the objective and how to actually validate this question, but would love to confirm with you all that my interpretation of how to conduct this test and evaluate the results is correct:

### Summary of Current Process

- In an attempt to reduce the file size of the tiles generated by [`etl_score_geo.py`](https://github.com/usds/justice40-tool/blob/main/data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py) we are mapping the default "long" name for each column to a corresponding "short" name that will be stored in the tile data. 
- The mapping of long name to short name is captured by the [`TILE_SCORE_COLUMNS`](https://github.com/usds/justice40-tool/blob/main/data/data-pipeline/data_pipeline/etl/score/constants.py#L149) constant in `etl/score/constants.py`
- And the actual renaming of the tiles from long to short happens in [this step of `GeoScoreETL.write_esri_shapefile()`](https://github.com/usds/justice40-tool/blob/main/data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py#L344)

### Question to Test

Would preserving the "long" name of each column when we generate the tiles meaningfully increase the size of the output files, or prevent the tiles from building at all?

### Testing Methodology

- Add a feature flag `shorten_column_names` that when set to `false` will skip the `.rename(columns=renaming_map)` when writing `self.geojson_score_usa_high` to `self.SCORE_SHP_FILE`
- Run two tests:
  - **Control:** One with the `shorten_column_names` flag set to `true` to get a baseline file size with short names
  - **Experimental:** A second with the `shorten_column_names` flag set to `false` to establish deviation from the baseline

### Definition of Done

- The two tests have been run and we've analyzed the differences in:
  - Output file size
  - Build time
  - _Anything else?_
- An ADR has been drafted which outlines the findings from these tests and a recommendation for either continuing to use the short names or returning to the use of the full column names.

[...]

Issue #1217, opened 2022-02-01
As a data developer, I want to copy ETL testing framework from NRI throughout all other 24 ETL sources.


lucasmbrown-usds:
**Description**
In #1075 , we worked to develop a sensible approach to testing ETL classes that we applied to one class only (NationalRiskIndex). 

This task is track the work of rolling out that approach, once finalized, to the rest of the ETL classes. 

These are checklists for making updates that will need to be implemented to succeed in this:

# ETL class changes

- [ ] Make sure these properties are set on the ETL class
        * NAME
        * LAST_UPDATED_YEAR
        * SOURCE_URL
        * GEO_LEVEL
        * COLUMNS_TO_KEEP (might already be set)

- [ ] Make sure that `def transform` sets `self.output_df` with its results 
- [ ] Delete any child-class specific `def load` if it's no longer necessary. 

# `etl_score.py` changes
- [ ] Replace the section in `extract` where it loads the dataframe from this ETL source, or create one if it doesn't exist yet
        * Replace `read_csv` etc with `<ClassNameETL>.get_data_frame()` 
        * (Only do this for tract-level data. So at this time, ignore this for the ETL for tree equity which is at the level of census block groups.) 

# `scoring_comparison.ipynb` changes 
- [ ] If the source is used directly in the comparison tool (i.e., not being passed through the big usa.csv file), then instead move this source to being directly passed through the `etl_score.py` pipeline 
        * Delete cell in comparison tool that loads this data source directly
        * Add section on `<ClassNameETL>.get_data_frame()` to `etl_score.py` as mentioned above 

# Creation of test file

PR #1534, opened 2022-04-05
NRI dataset and initial score YAML configuration

### LINKED ISSUE #1417: Refactor the NRI dataset and initial score YAML configuration

esfoobar-usds:
Closes #1417


.github/workflows/deploy_be_staging.yml
```diff
@@ -109,7 +109,7 @@ jobs:
           # Deploy to S3 for the staging URL
           message: |
             ** Map Deployed! **   
-            Map with Staging Backend: https://screeningtool.geoplatform.gov/en/cejst/?flags=stage_hash=${{env.PR_NUMBER}}/${{env.SHA_NUMBER}}
+            Map with Staging Backend: https://screeningtool.geoplatform.gov/en?flags=stage_hash=${{env.PR_NUMBER}}/${{env.SHA_NUMBER}}
```
vim-usds, line 112: nice catch!

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,47 @@
...
+    load_fields:
+      - short_name: "ex_loss"
+        df_field_name: "RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME"
+        long_name: "FEMA Risk Index Expected Annual Loss Score"
+        field_type: float
+        tile_include: true
+        csv_download: true
+        excel_download: true
+      - short_name: "ex_pop_loss"
+        df_field_name: "EXPECTED_POPULATION_LOSS_RATE_FIELD_NAME"
```
vim-usds, line 21: The tile name for `EXPECTED_POPULATION_LOSS_RATE_FIELD_NAME` that the FE expects for this is `EPLR_PFS` along with a boolean of `EPL_ET`. Should these values also be included in this datasets.yml file?

vim-usds (approved):
Some quick questions

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Great question, and the short answer is we'll figure the side effects in another PR. We might change the names -- or not that the tile file uses, but it will be with your input of course.

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -77,8 +91,51 @@ class ExtractTransformLoad:
...
+            sys.exit()
+
+        # set some of the basic fields
+        cls.LAST_UPDATED_YEAR = dataset_config["last_updated_year"]
+        cls.SOURCE_URL = dataset_config["source_url"]
+        cls.INPUT_EXTRACTED_FILE_NAME = dataset_config["extracted_file_name"]
+        cls.INPUT_GEOID_TRACT_FIELD_NAME = dataset_config[
+            "input_geoid_tract_field_name"
+        ]
+        cls.NULL_REPRESENTATION = dataset_config["null_representation"]
```
emma-nechamkin, line 127: Is this field optional? (if not, can it be?)

lucasmbrown-usds (changes requested):
Overall looking great! Lots of comments, but I'm excited about this. It would be great to update https://github.com/usds/justice40-tool/issues/1217 to document all the changes that will need to happen to other ETL classes to make them match this framework.

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -55,6 +59,9 @@ class ExtractTransformLoad:
     # SOURCE_URL is used to extract source data in extract().
     SOURCE_URL: str = None
 
+    # INPUT_EXTRACTED_FILE_NAME is the name of the file after extract().
+    INPUT_EXTRACTED_FILE_NAME: str = None
```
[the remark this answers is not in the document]
emma-nechamkin: this is what gets unpacked from the download, i believe... so it's not `usa.csv` because it's the data input

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -64,6 +71,13 @@ class ExtractTransformLoad:
     # COLUMNS_TO_KEEP is used to identify which columns to keep in the output df.
     COLUMNS_TO_KEEP: typing.List[str] = None
 
+    # INPUT_GEOID_TRACT_FIELD_NAME is the field name that identifies the Census Tract ID
+    # on the input file
+    INPUT_GEOID_TRACT_FIELD_NAME: str = None
+
+    # NULL_REPRESENTATION is how nulls are represented on the input field
+    NULL_REPRESENTATION: str = None
```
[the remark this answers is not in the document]
emma-nechamkin: We definitely should use a null representation (think about how much easier it would be for census data, which uses `-666666666` or something) and that representation is specified in some, not all, ETLs. This is to say -- I think it makes sense to have but should be an optional argument.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,47 @@
+---
+datasets:
+  - long_name: "FEMA National Risk Index"
+    short_name: "nri"
+    module_name: national_risk_index
+    last_updated_year: 2020
+    source_url: https://hazards.fema.gov/nri/Content/StaticDocuments/DataDownload//NRI_Table_CensusTracts/NRI_Table_CensusTracts.zip
+    extracted_file_name: "NRI_Table_CensusTracts.csv"
+    description: "Dataset from FEMA that identifies communities most at risk to 18 natural hazards."
```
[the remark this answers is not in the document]
emma-nechamkin: Hmm, why? Are you thinking then we could port this to the front end from the YAML file?

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -77,8 +91,51 @@ class ExtractTransformLoad:
...
+            )
+        except StopIteration:
+            # Note: it'd be nice to log the name of the dataframe, but that's not accessible in this scope.
+            logger.error(
+                f"Exception encountered while extracting dataset config for dataset {cls.NAME}"
+            )
+            sys.exit()
+
+        # set some of the basic fields
+        cls.LAST_UPDATED_YEAR = dataset_config["last_updated_year"]
```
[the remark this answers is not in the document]
emma-nechamkin: This is interesting. I think we could go back to the `dataset_year` folder structure, but we still need to think through column names. I think this is actually going to be a bigger headache than we can easily address?

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -77,8 +91,51 @@ class ExtractTransformLoad:
...
+        cls.SOURCE_URL = dataset_config["source_url"]
```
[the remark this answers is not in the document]
emma-nechamkin: couldn't we just pass a list?

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -77,8 +91,51 @@ class ExtractTransformLoad:
...
+        cls.INPUT_EXTRACTED_FILE_NAME = dataset_config["extracted_file_name"]
+        cls.INPUT_GEOID_TRACT_FIELD_NAME = dataset_config[
```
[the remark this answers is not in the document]
emma-nechamkin: This is my bad. I thought we had streamlined down to just tracts. We certainly do for stuff that makes it into the final score. 

As far as I can tell, excepting `tree_equity_score` everything uses tract (`persistent_poverty` has two fields)

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
emma-nechamkin: I think the idea here is the short name is part of the renamed column that we use internally, vs the string that describes the column.

esfoobar-usds: I think more than `short_name`, these would become column IDs that we can reference in other ETL or CI/CD processes, so even if we don't use them for the tiles, there's a value to adding them.

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
esfoobar-usds: Correct, this is the name of the extracted zip file from the source.

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -64,6 +71,13 @@ class ExtractTransformLoad:
     # COLUMNS_TO_KEEP is used to identify which columns to keep in the output df.
     COLUMNS_TO_KEEP: typing.List[str] = None
 
+    # INPUT_GEOID_TRACT_FIELD_NAME is the field name that identifies the Census Tract ID
+    # on the input file
+    INPUT_GEOID_TRACT_FIELD_NAME: str = None
```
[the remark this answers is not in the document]
esfoobar-usds: TODO

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
esfoobar-usds: Makes sense to make it a list.

data/data-pipeline/data_pipeline/etl/base.py
esfoobar-usds: Per comment below, we will make it optional and make it a list.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -27,8 +25,16 @@ class NationalRiskIndexETL(ExtractTransformLoad):
     AGRIVALUE_LOWER_BOUND = 408000
 
     def __init__(self):
-        self.INPUT_CSV = self.get_tmp_path() / "NRI_Table_CensusTracts.csv"
+        # load YAML config
+        self.DATASET_CONFIG = super().yaml_config_load()
```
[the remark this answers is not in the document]
esfoobar-usds: I tried but you can't define an `__init__` method on the base class, but if you know a way to do it, I would totally go for it.

data/data-pipeline/data_pipeline/utils.py
```diff
@@ -8,6 +8,7 @@
 import uuid
 import zipfile
 from pathlib import Path
+from marshmallow import ValidationError
```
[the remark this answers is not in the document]
esfoobar-usds: No, this is a `marshmallow` validation when we load the YAML for the dataset, per line 355:

```
        try:
            yaml_config_schema().load(yaml_dict)
        except ValidationError as e:
            logger.error(f"Invalid YAML config file {yaml_file_path}")
```

data/data-pipeline/data_pipeline/utils.py
```diff
@@ -350,7 +351,12 @@ def load_yaml_dict_from_file(
 
         # validate YAML
         yaml_config_schema = class_schema(schema_class)
-        yaml_config_schema().load(yaml_dict)
+
+        try:
+            yaml_config_schema().load(yaml_dict)
+        except ValidationError as e:
+            logger.error(f"Invalid YAML config file {yaml_file_path}")
+            logger.error(e.normalized_messages())
```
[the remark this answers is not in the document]
esfoobar-usds: You're right, we should stop execution of this thread at this point.

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
esfoobar-usds: I think we should kick this can to the versioning discussion and continue on as is right now.

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -0,0 +1,40 @@
+from dataclasses import dataclass, field
```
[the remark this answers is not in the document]
esfoobar-usds: Yeah, they're a bit different. Marshmallow deals with serialization, deserialization, whereas dataclasses allow you to define classes easier and more readable. And they both work together really nicely. By all means ask me any question.

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
esfoobar-usds: Nor sure what the resolution of this would be @lucasmbrown-usds @emma-nechamkin?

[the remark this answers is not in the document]
emma-nechamkin: (I personally think forcing the tool to use tracts and crosswalking from blocks to tracts when we clean the data makes the most sense, but I'm totally happy to be overruled)

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
esfoobar-usds: This is a bit more complicated, because if we pass a list of source urls, we also a corresponding list of extracted file names. Since this is not saving tremendous amounts of code, I can remove both properties and let the ETL handle manually.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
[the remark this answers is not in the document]
esfoobar-usds: We could place the method on the runner, let me try that.

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
esfoobar-usds: I have removed `source_url` and the input file name,

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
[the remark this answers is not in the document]
esfoobar-usds: So I spoke too soon. It seems it's hard to do this one, since the runner instance doesn't seem to have the scope to load the class variables from the base `yaml_config_load` and a multitude of tests failed. Unless you can think of something else, I don't know how to tackle this one.

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
esfoobar-usds: I have added the documentation on the ticket as requested.

emma-nechamkin:
lgtm!

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
lucasmbrown-usds: Got it, thanks

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
lucasmbrown-usds: Got it, so we'll make this a list?

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
lucasmbrown-usds: Hmm. We can kick the can but we also know this will definitely be a problem for v1.0. Definition M (CEJST Beta) uses older EJScreen data, and we're planning to update the EJScreen data year for Narwhal. Without being able to maintain two years at once, we will either not be able to accurately recreate Definition M or we will not be able to accurately create Narwhal.

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
lucasmbrown-usds: Hmm. I think forcing things to tracts could make sense. But if we do that, we need to create tickets to 1) update the code for `tree_equity_score` to aggregate to tracts and 2) remove all of the various code throughout the codebase that supports both block groups and tracts.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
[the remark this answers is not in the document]
lucasmbrown-usds: Ok. Have you created a ticket for that work? Removing front-end configuration would be a big part of reducing Beth's famous "19 steps to update a dataset."

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,44 @@
...
+    module_name: national_risk_index
+    last_updated_year: 2020
+    description: "Dataset from FEMA that identifies communities most at risk to 18 natural hazards."
+    input_geoid_tract_field_name: "TRACTFIPS"
+    load_fields:
+      - short_name: "ex_loss"
+        df_field_name: "RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME"
+        long_name: "FEMA Risk Index Expected Annual Loss Score"
+        field_type: float
+        tile_include: true
```
lucasmbrown-usds, line 14: As discussed in the design discussion in `YAML-Based Score Datasets`, I don't this this is the right place for this field. Whether or not a field should be included on the tile, csv, and excel files is dependent on the score that is currently active. For instance, if Score N is active, it will have certain fields in the tiles that are not there when Score M is active.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,44 @@
...
+        excel_download: true
+      - short_name: "ex_pop_loss"
+        df_field_name: "EXPECTED_POPULATION_LOSS_RATE_FIELD_NAME"
+        long_name: "Expected population loss rate (Natural Hazards Risk Index)"
+        field_type: float
+        tile_include: true
+        csv_download: true
+        excel_download: true
+      - short_name: "ex_ag_loss"
+        df_field_name: "EXPECTED_AGRICULTURE_LOSS_RATE_FIELD_NAME"
```
lucasmbrown-usds, line 25: See comment in design doc but: 

> You could just force the transform step of the ETL class to convert the field to long_name, and then run a check as part of ETL.validate to make sure all those fields are present?

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
lucasmbrown-usds: > For the "include in CSV/Tiles/Excel" prompt, whether or not they should be included will actually depend on what score is "active" in the tool. 
> 
> So maybe the data model is something like: 
> 
> load_fields is a list, and within load_fields, there's a dictionary of {"score_name": {"tile_include": bool, "csv_download": bool, "excel_download": bool}}. 
> 
> E.g., 
> 
> {"Definition M": {"tile_include": True, "csv_download": True, "excel_download": True},
> "Definition N": {"tile_include": True, "csv_download": False, "excel_download": False}}

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,44 @@
...
+    description: "Dataset from FEMA that identifies communities most at risk to 18 natural hazards."
+    input_geoid_tract_field_name: "TRACTFIPS"
+    load_fields:
+      - short_name: "ex_loss"
+        df_field_name: "RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME"
+        long_name: "FEMA Risk Index Expected Annual Loss Score"
+        field_type: float
+        tile_include: true
+        csv_download: true
+        excel_download: true
```
lucasmbrown-usds, line 16: From design doc: 

> 
> This is a nitty nit, but I wonder if we can make these names a bit more obvious and consistent, something along the lines of: 
> 
> * include_in_tiles
> * include_in_csv
> * include_in_excel

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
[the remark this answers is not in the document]
lucasmbrown-usds: Cool cool. My main interest is in making sure we provide validation on the deserialization of loading from the YAML. E.g., will something (either dataclasses or Marshmallow) make sure that every DatasetsConfig has all the required fields, valid values in every field, etc?

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -0,0 +1,41 @@
...
+    @dataclass
+    class Dataset:
+        @dataclass
+        class LoadField:
+            short_name: str
+            df_field_name: str
+            long_name: str
+            field_type: FieldType = field(
+                metadata={"by_value": True}
+            )  # this will load the field type's Enum value
```
lucasmbrown-usds, line 26: nice!

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,44 @@
...
+      - short_name: "ex_ag_loss"
+        df_field_name: "EXPECTED_AGRICULTURE_LOSS_RATE_FIELD_NAME"
+        long_name: "Expected agricultural loss rate (Natural Hazards Risk Index)"
+        field_type: float
+        tile_include: true
+        csv_download: true
+        excel_download: true
+      - short_name: "ex_bldg_loss"
```
lucasmbrown-usds, line 31: I feel like we're missing a few fields to configure per data source / fields that would be really helpful, such as: do we want to create percentiles and/or reverse percentiles for this field?

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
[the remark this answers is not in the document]
lucasmbrown-usds: Hmm, yeah, this is interesting. I don't have an immediate solution. The scopes are a little funny. Maybe you and I can pair on it, or you can pair with another engineer? To be honest with scope issues like this, when I'm doing it alone I tend to throw spaghetti at the wall, but talking it out with someone can be super helpful.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -59,22 +66,12 @@ def __init__(self):
         )
         self.CONTAINS_AGRIVALUE = "Contains agricultural value"
 
-        self.COLUMNS_TO_KEEP = [
```
lucasmbrown-usds, line 62: This logic is all replaced by the YAML config now, right?

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
lucasmbrown-usds: Also do we want to include the field in the comparison tool either as 1) an index to compare, or 2) a characteristic to compare (such as tract income).

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,44 @@
+---
```
lucasmbrown-usds, line 1: I'm wondering if we want to set a "single switch" for the entire application of which CEJST definition of DACs (e.g., `Definition M`) is currently live in the app. For instance, the front-end code could be updated to branch as well so that "if Def M is live, show these fields in the side bar, and if Def N is live, show this in the sidebar". That would make it easier to update the user interface and backend in a single deploy by switching the name of the definition of DACs that is active.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,44 @@
...
+        csv_download: true
+        excel_download: true
+      - short_name: "ex_bldg_loss"
+        df_field_name: "EXPECTED_BUILDING_LOSS_RATE_FIELD_NAME"
+        long_name: "Expected building loss rate (Natural Hazards Risk Index)"
+        field_type: float
+        tile_include: true
+        csv_download: true
+        excel_download: true
+      - short_name: "has_ag_val"
```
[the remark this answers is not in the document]
lucasmbrown-usds: Oh, and similarly, it'd be great to move the data dictionary logic in here as well.

lucasmbrown-usds (changes requested):
Here are some high-level comments: 

It seems like we currently have a *lot* of config for each field that’s covered in other places that’s not currently covered here: 

* Column sort order (depends on what the active DACs definition is, e.g. Definition M vs Definition N)
* score_name and label when writing CSVs and Excel (TBH I’m not exactly remembering what these configs do) 
* Field format (string, bool, int64, etc) 
* Data dictionary logic and conversion 
* Whether field should be included in the comparison tool either as 1) an index to compare (such as Definition M, Federal Poverty Line, CalEnviroScreen), or 2) a characteristic to compare (such as tract income).
* Field description in the front end 
* Short names used for compression for tiles (if necessary: see https://github.com/usds/justice40-tool/issues/1221. Also note this is different than the field listed as “short_name” here.) 
* Whether or not to create percentiles and/or reverse percentiles from this field

The design document `Proposal(s) for streamlining adding new data / methodology to tool` has a bunch of these additional fields listed in Proposal A. 

I know some of the PR comments say we'll tackle this in the future. TBH I'm uncomfortable deferring too much to the future, because I think we want to get our general framework near-perfect for this example dataset (NRI), and then roll through one-by-one and update every other dataset. So we don't want to have to roll through and update all 24 datasets multiple times.

Similarly, if we can figure out the `init` methods so we don't have to copy redundant code across 24 refactors, that'd really streamline things as well. 

As mentioned in the design doc, 

> I'm wondering if we can copy-paste Beth's list of 19 steps we take when we update the methodology into here, and show which ones would be eliminated/reduced by this refactor. This would be my biggest request, I think!

I'd really love to go through those 19 and show which steps are eliminated or at least reduced by this effort and which aren't. I believe they're listed in `Proposal(s) for streamlining adding new data / methodology to tool`.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Per our 1x1 conversation, I don't agree with having score definitions on the codebase. I will be working on a score versioning doc, but what I told you on the call, summarized, is that score versions will be tied to Git tags, side effects and static assets will be tied to S3 folders and we will have an upgrade/downgrade score version GHA that will automatically generate these artifacts and also inform the frontend on a version list that could power a pull down on the frontend to select and visualize the different versions.,

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: I don't understand what we're gaining here, maybe you can explain in our next pair?

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: TODO

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
lucasmbrown-usds: As discussed, fine with versioning approach, look forward to seeing the design doc and tickets for versioning.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Spoke to @lucasmbrown-usds about this. 

We want to try to dynamically set (using a Python internal setter) the variables that are related to the score output and their long strings to be loaded from the YAML. @esfoobar-usds will investigate how to do this. The variables that are just internal to the dataset and are used for internal calculations should *not* be set here.

One downside could be the auto-complete on editors will not work, but @lucasmbrown-usds will comment here on how that could be solved.

lucasmbrown-usds: it would be something like: 

instead of: 
```
self.RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME = (
            "FEMA Risk Index Expected Annual Loss Score"
```

do something like:  
```
        self.RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME: str
        [set all these]

        for field in fields_from_config:
            setattr(self, field.df_field_name, field.long_name)

```

This way, you'd still have IDE-autocompletes on fields like self.RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME, but at least you wouldn't have to redundantly specific the long_name.

lucasmbrown-usds: Looks like you have to define it statically to get dynamic completion in IDEs like Pycharm: https://stackoverflow.com/a/63594033/1569221

lucasmbrown-usds: But maybe we don't care that much about IDE autocompletes so we don't worry about it.

esfoobar-usds: @lucasmbrown-usds and I agree that the auto-completion is not as important, but definitely would be a nice feature. We'll move forward without auto-completion for this PR.

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -0,0 +1,41 @@
...
+    LOSS_RATE_PERCENTAGE = "loss_rate_percentage"
+
+
+@dataclass
+class DatasetsConfig:
+    @dataclass
+    class Dataset:
+        @dataclass
+        class LoadField:
+            short_name: str
```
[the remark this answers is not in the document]
esfoobar-usds: TODO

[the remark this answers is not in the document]
lucasmbrown-usds: Relevant: https://stackoverflow.com/a/51131941/1569221. There's a few good options on that thread.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
esfoobar-usds: Yes!

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
lucasmbrown-usds: Reverse percentiles code now: data/data-pipeline/data_pipeline/etl/score/etl_score.py:477

esfoobar-usds: We discussed with @lucasmbrown-usds and it totally makes sense to add any field metadata to be described on the `YAML` dataset. However anything related to `etl/score` and `etl/score_post` will come in separate PRs.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Per convo with @lucasmbrown-usds, the versioning functionality for both frontend and backend will come in a separate release that will include both.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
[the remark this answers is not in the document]
esfoobar-usds: Once this release is completed, we won't have the `csv.yaml` or `excel.yaml` anymore. Some of the properties that are specific to the side effects (.e.g excel column length) would be ported here.

For example, the excel yaml "column width" could be part of a global config excel section. 

The overall strategy is to have one and just one config file that "rules them all".

esfoobar-usds:
Per conversation with @lucasmbrown-usds and @vim-usds:

- [A Spike ticket](https://github.com/usds/justice40-tool/issues/1699) was created by @vim-usds to tackle the frontend component of this
- @lucasmbrown-usds will Slack @esfoobar-usds how to implement the comparison tool component (Whether field should be included in the comparison tool either as 1) an index to compare (such as Definition M, Federal Poverty Line, CalEnviroScreen), or 2) a characteristic to compare (such as tract income).)

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: TODO: Versioning Doc

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Done!

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Done

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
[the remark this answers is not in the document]
esfoobar-usds: Done.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Added `reverse_percentile` optional.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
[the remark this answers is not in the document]
esfoobar-usds: @lucasmbrown-usds Do we ever expect to have different order between Excel and CSV?

esfoobar-usds:
The Reduction on “Beth’s List”
This is what the original 19 point list would look like with this implementation:

- Write and approve new copy (long and short descriptions)
- Add new dataset fields on `datasets.yml`, including front end copy
- Create etl module to write `usa.csv`
- Update to score code to implement new methodology
- Update comparison tool (if necessary) for a new methodology and re-run to look at outcomes
- Add new copy to living copy doc
- Update copy on .pdf in download package
- Update technical support document (if necessary - TSD will include the final .pdf from download package)

Here are the items that would not be necessary and would be taken care by the YAML implementation and an explanation of why:

- Update ETL code to add fields: typically this means adding fields to a constants file, which is not needed anymore, since they’re loaded from the datasets.yml
- Update score_etl to calculate percentiles or reverse percentiles: the reverse percentiles would be defined in the datasets.yml
- Re-run score to update data: not required
- Add new data indicators (if any) to usa.csv: the data indicators would automatically populate from datasets.yml
- Add new indicators about whether or not thresholds are exceeded (if any): the thresholds could be a general line or a custom per field prop on the yaml file
- In the backend, add new indicators to tiles that will be used for map and side panel: driven by datasets.yml
- In the frontend, display new indicators in side panel: A JSON file would be generated by the backend process with the indicators map
- Add new logic about when to display the indicator as blue in the side panel (if any): this could be set on the YAML
- Add new indicators to export file: driven by datasets.yml
- Add new indicators to spreadsheet: driven by datasets.yml
- Adjust sort order for spreadsheet: driven by datasets.yml
- (Eventually) add to spreadsheet data dictionary” driven by datasets.yml
- Update copy on methodology page: A JSON file would be generated by the backend process with the methodology indicators

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
lucasmbrown-usds: @esfoobar-usds -- do you want to create the two tickets above?

lucasmbrown-usds:
> * @lucasmbrown-usds will Slack @esfoobar-usds how to implement the comparison tool component (Whether field should be included in the comparison tool either as 1) an index to compare (such as Definition M, Federal Poverty Line, CalEnviroScreen), or 2) a characteristic to compare (such as tract income).)

@esfoobar-usds -- here's my suggestion, to match the style of `include_in_X`: 

```
include_in_comparison_tool_as_index (bool): Whether or not to include this field in the comparison tool as an index used as comparison (e.g., this field might be a state or national index that identifies priority communities). The field itself must be a boolean for the comparison tool to work appropriately. 
include_in_comparison_tool_as_statistical_descriptor (bool): Whether or not to include this field in the comparison tool as a statistical descriptor of census tracts (e.g., this field might income levels, life expectancy, etc). This will be used to generate reports that produce information such as, tracts identified by Index A but not Index B have higher income levels but lower life expectancy. 
``` 

As examples, here are some examples of what would have `include_in_comparison_tool_as_index=True` (it'll be False for 95% of fields, so you can default it to False everywhere): 

``` 
field_names.EJSCREEN_AREAS_OF_CONCERN_NATIONAL_80TH_PERCENTILE_COMMUNITIES_FIELD
field_names.SCORE_M_COMMUNITIES
field_names.MAPPING_FOR_EJ_PRIORITY_COMMUNITY_FIELD
field_names.EPA_RSEI_SCORE_THRESHOLD_FIELD
``` 

Here is the complete list of what will be `include_in_comparison_tool_as_statistical_descriptor=True` (it'll be False for 95% of fields, so you can default it to False everywhere except for these fields): 

```
[
  field_names.POVERTY_LESS_THAN_100_FPL_FIELD,
  field_names.POVERTY_LESS_THAN_200_FPL_FIELD,
  field_names.COLLEGE_ATTENDANCE_FIELD,
  field_names.MEDIAN_INCOME_AS_PERCENT_OF_AMI_FIELD,
  field_names.LINGUISTIC_ISO_FIELD,
  field_names.UNEMPLOYMENT_FIELD,
  field_names.HIGH_SCHOOL_ED_FIELD,
  field_names.MEDIAN_INCOME_FIELD,
  field_names.URBAN_HEURISTIC_FIELD,
  field_names.LIFE_EXPECTANCY_FIELD,
  field_names.HEALTH_INSURANCE_FIELD,
  field_names.PHYS_HEALTH_NOT_GOOD_FIELD,
  field_names.DIABETES_FIELD,
  field_names.LOW_READING_FIELD + field_names.PERCENTILE_FIELD_SUFFIX,
]
```

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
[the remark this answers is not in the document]
lucasmbrown-usds: I don't think we would have a different order between excel and CSV!

It'd be nice to keep things as simple as possible, so it'd be great to keep them in the same order if that's fine

If we split territories into a separate sheet, then you can still use the same column order -- like in the excel columns 1-25 are in the first tab/sheet, columns 26-31 are in the second tab/sheet, and in the CSV, they are all in order 1-31 in the CSV. Then i think we wouldn't need to define column order twice. 

@esfoobar-usds what's our plan for getting column order into the YAML here? 

One option is to have a float value for column order -- so you don't have to update every single field's order precisely as integers in order. 

E.g., it'd be easy to update: 

```
field1 column_order: 1
field2 column_order: 2
field3 column_order: 3
...
field37 column_order: 37
```

To 

```
field1 column_order: 1
field2 column_order: 2
field38 column_order: 2.5
field3 column_order: 3
...
field37 column_order: 37
```

Without having to re-specify the order of the other 36 fields. Does that make sense?

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -16,24 +16,50 @@ class FieldType(Enum):
 class DatasetsConfig:
     @dataclass
     class Dataset:
+        """A class that defines a dataset and its load variables.
+
+        Attributes:
```
lucasmbrown-usds, line 21: love it! 🎉

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -16,24 +16,50 @@ class FieldType(Enum):
...
+                df_field_name (str): Name for the field in the etl class.
+                long_name (str): Column name for the dataset's output csv.
+                field_type (FieldType): An enum that dictates what type of field this is. This will be used on the `etl_score_post`
+                for the data manipulation. The `by_value` metadata prop will load the field type's Enum value instead of the index, i.e. "string"
+                and not STRING reverse_percentile (Optional bool): An optional bool to denote this field to be a reverse_percentile.
+                TODO: data/data-pipeline/data_pipeline/etl/score/etl_score.py:477
+                include_in_tiles (bool): Include this field on the tile export.
+                include_in_csv (bool): Include this field on the CSV export.
+                include_in_excel (bool): Include this field on the Excel export.
+                column_in_position (int): column position in CSV and Excel files
```
lucasmbrown-usds, line 47: We might actually want to make this a float? 

That way you don't have to update every single field's order precisely as integers in order. 

E.g., it'd be easy to update: 

```
field1 column_order: 1
field2 column_order: 2
field3 column_order: 3
...
field37 column_order: 37
```

To 

```
field1 column_order: 1
field2 column_order: 2
field38 column_order: 2.5 [only value that has changed]
field3 column_order: 3
...
field37 column_order: 37
```

Which is less complicated of a change than: 

```
field1 column_order: 1
field2 column_order: 2
field38 column_order: 3 [updated]
field3 column_order: 4 [updated]
...
field37 column_order: 38 [updated]
```


Without having to re-specify the order of the other 36 fields. Does that make sense?

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -0,0 +1,67 @@
...
+        class LoadField:
+            """A class to define the fields to be saved on the dataset's output.
+
+            These fields will be then imported by the score generation ETL.
+
+            Attributes:
+                short_name (str): Used in conjunction with the dataset's `short_name` for files where short names are needed.
+                df_field_name (str): Name for the field in the etl class.
+                long_name (str): Column name for the dataset's output csv.
+                field_type (FieldType): An enum that dictates what type of field this is. This will be used on the `etl_score_post`
```
lucasmbrown-usds, line 40: I'm having a hard time following this particular comment? Is there supposed to be newline before `reverse_percentile`?

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -0,0 +1,67 @@
...
+                for the data manipulation. The `by_value` metadata prop will load the field type's Enum value instead of the index, i.e. "string"
+                and not STRING reverse_percentile (Optional bool): An optional bool to denote this field to be a reverse_percentile.
```
lucasmbrown-usds, line 42: I think we'll need both of the following: 

```
create_percentile (bool): Whether or not the backend processing should create a percentile field (ranked in ascending order) from the values in this field.
create_reverse_percentile (bool): Whether or not the backend processing should create a "reverse percentile" field (ranked in descending order) from the values in this field.
```

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -0,0 +1,67 @@
...
+from typing import List, Optional
+
+
+class FieldType(Enum):
+    STRING = "string"
+    INT64 = "int64"
+    BOOL = "bool"
+    FLOAT = "float"
+    PERCENTAGE = "percentage"
+    LOSS_RATE_PERCENTAGE = "loss_rate_percentage"
```
lucasmbrown-usds, line 12: Hmm. The only thing different about `LOSS_RATE_PERCENTAGE` from `PERCENTAGE` is that loss rate percentage is printed with more significant digits when we write the output to Excel. 

Maybe instead of doing this, we just add an optional new characteristic to the LoadFields called `number_of_decimals_in_output` or something and have it default to 2 for all floats?

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -0,0 +1,67 @@
...
+    @dataclass
+    class Dataset:
+        """A class that defines a dataset and its load variables.
+
+        Attributes:
+            long_name (str): A human readable title for the dataset.
+            short_name (str): used to compose the short variable names for tiles/arcgis. All short variable names will be prepended
+            with the short name of the data set it comes from, i.e. `nri__ex_loss`.
+            module_name (str): A string that matches both the Python module name for the dataset and the `NAME` property on the ETL class.
+            description (str): A human readable description of the dataset.
```
lucasmbrown-usds, line 26: I think we'll actually want two descriptions, sorry for not catching that sooner. Maybe `description_short` and `description_long`? 

One is the description that shows up in the sidebar. E.g., for field "Expected agriculture loss rate", the `description_short = Economic loss rate to agricultural value resulting from natural hazards each year.` 

Screen Shot 2022-07-06 at 6 59 35 PM

The other description is on the data and methodology page. E.g., for field "Expected agriculture loss rate", the `description_long = Percent of agricultural value at risk from losses due to fourteen types of natural hazards that have some link to climate change: avalanche, coastal flooding, cold wave, drought, hail, heat wave, hurricane, ice storm, landslide, riverine flooding, strong wind, tornado, wildfire, and winter weather. Rate calculated by dividing the agricultural value at risk in a census tract by the total agricultural value in that census tract.` 

Screen Shot 2022-07-06 at 7 01 02 PM

lucasmbrown-usds:
Overall looking great! 

I nitted a bit about the field definitions, but just make sure we want to get those right before rolling out everywhere. I think we're very very close; I'll probably be ready to approve on the next pass incorporating responses to the comments so far!

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
esfoobar-usds: Totally! Updated the code.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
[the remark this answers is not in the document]
esfoobar-usds: Per @KameronKerger, the answer is no.

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
esfoobar-usds: Great! Done!

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
esfoobar-usds: Agreed! Done!

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
esfoobar-usds: Yup, this was a boo-boo. Fixed.

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
esfoobar-usds: Done!

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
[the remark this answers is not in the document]
mattbowen-usds: There are three option I see ---

* Do it in  `ExtractTransformLoad.__init__` and then children need to call `super().__init__()` --- not beautiful but also not uncommon. Given that `ExtractTransformLoad` doesn't have an init right now and there's not much else to put in there, this doesn't feel like a net win, unless there's other common setup code that could be extracted.
* Use a metaclass --- pretty sure you could write `yaml_config_load` in terms of a metaclass's `__init__`. Then, you set the metaclass of `ExtractTransformLoad` to your YAML-loading metaclass and children get that functionality. Downside is future maintainability, since like most programmers are not using metaclasses regularly, which can make it hard to reason about them quickly. Upside is what `yaml_config_load` does --- inspecting an external configuration to reflect data there back into the object --- is pretty classic metaclass work and the one place people won't be like "why did you do this?!"
* Write a class decorator that does it --- I think this would be my chosen approach. So extract `yaml_config_load` from `ExtractTransformLoad` and make it a class decorator (which might require a tiny bit of rewriting but not much), then decorate the base `ExtractTransformLoad` with the new yaml_config_load decorator. Children will inherit this decoration and you save the boiler plate but don't fall into madness writing a metaclass.

Finally, tangentially related, not sure you need to call `super()` to call `yaml_config_load` there, since `yaml_config_load` should be inherited like any other member and thereby available on `self`?

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -27,54 +26,41 @@ class NationalRiskIndexETL(ExtractTransformLoad):
...
+        self.DATASET_CONFIG = super().yaml_config_load()
+
+        # define the full path for the input CSV file
         self.INPUT_CSV = self.get_tmp_path() / "NRI_Table_CensusTracts.csv"
 
+        # this is the main dataframe
+        self.df: pd.DataFrame
+
+        # Output score variables (values set on datasets.yml)
+        self.RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME: str
```
mattbowen-usds, line 39: Given how `yaml_config_load` works, I think these should be class variables and not instance variables --- it doesn't matter a ton, but no need to accidentally confuse ourselves later thinking they're instance vars when they're set at the class level.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,49 @@
...
+    input_geoid_tract_field_name: "TRACTFIPS"
+    load_fields:
+      - short_name: "ex_loss"
+        df_field_name: "RISK_INDEX_EXPECTED_ANNUAL_LOSS_SCORE_FIELD_NAME"
+        long_name: "FEMA Risk Index Expected Annual Loss Score"
+        field_type: float
+        include_in_tiles: true
+        include_in_csv: true
+        include_in_excel: true
+        column_position: 1
```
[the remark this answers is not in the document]
emma-nechamkin: Just two questions here: 
1. When would we have something in the csv but not in the excel? Do we need both controls?
2. And, is the code in the ETL going to be reworked entirely to get rid of `field_names.py` and `constants.py`? I'm just a bit nervous because it seems like we are going to have to carry this refactor all the way through the etl code and it's already August 5.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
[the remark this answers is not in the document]
mattbowen-usds: okay ignore all of what I wrote --- I went to do an example of the decorator approach and learned about `__init_subclass__` and man I guess things have really changed since python 2.4. Anyway if you want to see what it would look like, see this commit: https://github.com/usds/justice40-tool/commit/39229c52fbb7662e94a3e88b05c483d2be020aab

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -3,7 +3,8 @@ datasets:
   - long_name: "FEMA National Risk Index"
     short_name: "nri"
     module_name: national_risk_index
-    description: "Dataset from FEMA that identifies communities most at risk to 18 natural hazards."
+    description_short: "Dataset from FEMA that identifies communities most at risk to 18 natural hazards."
+    description_long: "Dataset from FEMA that identifies communities most at risk to 18 natural hazards."
```
lucasmbrown-usds, line 7: This is not actually what I meant, sorry. 

There is a dataset description which is helpful. 

But for the short/long description fields, I was referring to the short/long description fields *per field*. So these would go into each of the `load_fields`. For instance, for the `load_field` that has `long_name: "Expected agricultural loss rate (Natural Hazards Risk Index)"`, this is the long description: https://screeningtool.geoplatform.gov/en/methodology#exp-agr-loss-rate:~:text=Percent%20of%20agricultural,that%20census%20tract. 

> Percent of agricultural value at risk from losses due to fourteen types of natural hazards that have some link to climate change: avalanche, coastal flooding, cold wave, drought, hail, heat wave, hurricane, ice storm, landslide, riverine flooding, strong wind, tornado, wildfire, and winter weather. Rate calculated by dividing the agricultural value at risk in a census tract by the total agricultural value in that census tract.

And this is the short description (from the sidebar): 

Screen Shot 2022-08-02 at 5 17 34 PM

> Economic loss rate to agricultural value resulting from natural hazards each year.

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -23,7 +22,8 @@ class Dataset:
             short_name (str): used to compose the short variable names for tiles/arcgis. All short variable names will be prepended
             with the short name of the data set it comes from, i.e. `nri__ex_loss`.
             module_name (str): A string that matches both the Python module name for the dataset and the `NAME` property on the ETL class.
-            description (str): A human readable description of the dataset.
+            description_short (str): Description used in the side panel.
```
lucasmbrown-usds, line 25: Same comment as above.

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -37,30 +37,46 @@ class LoadField:
...
                 df_field_name (str): Name for the field in the etl class.
                 long_name (str): Column name for the dataset's output csv.
-                field_type (FieldType): An enum that dictates what type of field this is. This will be used on the `etl_score_post`
-                for the data manipulation. The `by_value` metadata prop will load the field type's Enum value instead of the index, i.e. "string"
-                and not STRING reverse_percentile (Optional bool): An optional bool to denote this field to be a reverse_percentile.
-                TODO: data/data-pipeline/data_pipeline/etl/score/etl_score.py:477
+                field_type (FieldType): An enum that dictates what type of field this is. This will be used on the `etl_score_post` for the
+                    data manipulation. The `by_value` metadata prop will load the field type's Enum value instead of the index, i.e. "string"
+                    and not STRING
+                number_of_decimals_in_output (int): Used to represent number of decimals in side effects, like Excel.
```
lucasmbrown-usds, line 43: :tada:

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -37,30 +37,46 @@ class LoadField:
...
                 include_in_tiles (bool): Include this field on the tile export.
                 include_in_csv (bool): Include this field on the CSV export.
                 include_in_excel (bool): Include this field on the Excel export.
-                column_in_position (int): column position in CSV and Excel files
+                column_position (float): column position in CSV and Excel files. Value can be 2.5 to include an item between 2 and 3.
```
lucasmbrown-usds, line 47: I like it!

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -37,30 +37,46 @@ class LoadField:
...
+                create_percentile (bool): Whether or not the backend processing should create a percentile field (ranked in ascending order) from the
```
lucasmbrown-usds, line 48: love all these!

lucasmbrown-usds: Could you note that the default behavior (when the param is not specified) is equivalent to `False`?

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
lucasmbrown-usds: Can you specify what the default value is if number_of_decimals_in_output is not specified? 

Also - many of your doc strings are for optional fields but don't include the `Optional` note (e.g. `(Optional bool)`). Could you add that?

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -0,0 +1,49 @@
+---
+datasets:
+  - long_name: "FEMA National Risk Index"
+    short_name: "nri"
+    module_name: national_risk_index
+    description_short: "Dataset from FEMA that identifies communities most at risk to 18 natural hazards."
+    description_long: "Dataset from FEMA that identifies communities most at risk to 18 natural hazards."
+    input_geoid_tract_field_name: "TRACTFIPS"
+    load_fields:
+      - short_name: "ex_loss"
```
lucasmbrown-usds, line 10: Since this NRI class is the "canonical" refactor example, can we include all of the optional fields and explicitly set them to their default values (for the bools i think the defaults are all `False`)?

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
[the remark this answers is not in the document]
lucasmbrown-usds: @mattbowen-usds -- good question. I think the answer is that the column order applies *across* data sources that come from different ETL classes. So in this NRI class, we might set columns 7, 11, 18, 18.5, and 19, and in another class we set columns 8, 9.5, 12, etc. 

@emma-nechamkin good point in that `include_in_csv` and `include_in_excel` could probably be collapsed into a single boolean: `include_in_csv_and_excel`. 

I'm not sure the answer to @emma-nechamkin 's Question #2 -- that's a good one. @esfoobar-usds ?

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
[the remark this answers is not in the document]
lucasmbrown-usds: Bumping this -- where do you check that every DatasetsConfig has all the required fields, valid values in every field, etc?

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
```diff
@@ -0,0 +1,83 @@
...
+            """A class to define the fields to be saved on the dataset's output.
+
+            These fields will be then imported by the score generation ETL.
+
+            Attributes:
+                short_name (str): Used in conjunction with the dataset's `short_name` for files where short names are needed.
+                df_field_name (str): Name for the field in the etl class.
+                long_name (str): Column name for the dataset's output csv.
+                field_type (FieldType): An enum that dictates what type of field this is. This will be used on the `etl_score_post` for the
+                    data manipulation. The `by_value` metadata prop will load the field type's Enum value instead of the index, i.e. "string"
```
lucasmbrown-usds, line 41: This is very nitty but you could move the comment `The `by_value` metadata prop will load the field type's Enum value instead of the index, i.e. "string" and not STRING` down to the code (line 64). I'm not sure it's necessary for the docstring, it's more to explain it to future engineers working with this code.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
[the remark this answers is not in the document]
lucasmbrown-usds: What's the resolutions of this - are we comfortable with the code as is?

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
lucasmbrown-usds: That's a good point.

lucasmbrown-usds: I'll admit I'm often guilty of setting instance variables because `self` is my default rather than `cls`.

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
```diff
@@ -109,6 +104,6 @@ def test_get_output_file_path(self, mock_etl, mock_paths):
 
         output_file_path = etl._get_output_file_path()
         expected_output_file_path = (
-            data_path / "dataset" / "national_risk_index_2020" / "usa.csv"
+            data_path / "dataset" / "national_risk_index" / "usa.csv"
```
lucasmbrown-usds, line 107: Is this a breaking change for running the full ETL and score process?

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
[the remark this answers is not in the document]
mattbowen-usds: I went ahead and just refactored it in my work on #1518 anyway for fun, so I'm fine with this staying here and you all can decide if you like the `__init_subclass__` approach when you get to that one.

data/data-pipeline/data_pipeline/utils.py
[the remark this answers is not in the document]
lucasmbrown-usds: Nice!

lucasmbrown-usds (changes requested):
Getting close!

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
mattbowen-usds: I would not hold up merging to change them --- just like, I was reading what `yaml_config_load` actually does and was like "one day some future person might, _might_ find this confusing"

data/data-pipeline/data_pipeline/tests/sources/national_risk_index/test_etl.py
esfoobar-usds: No, we run the full ETL on each PR commit and it didn't fail, nor the tests.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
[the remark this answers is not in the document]
esfoobar-usds: I think this is brilliant @mattbowen-usds, had never heard about `__init__subclass__` dunder  -- I'll leave as is for this PR and can get updated when we merge yours.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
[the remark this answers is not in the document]
esfoobar-usds: About point 2., we decided to wait until the new indicators are done, and then do the score and score post generation refactor using YAML.

About the column order, in our existing CSV and Excel YAML files that drive the column order for the side effects, it is assumed the column order is the list order. However, datasets YAML could have fields that are not going to be included in the downloadable assets, so we'd need to signify that. However, that flag can come from the `include_in_csv_and_excel` column, which I would rename as `include_in_downloadable_files`, so once again, I think we might be able to just make the list order be the column order. Let's discuss in the meeting today.

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
esfoobar-usds: Great point, Changed.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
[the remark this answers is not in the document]
esfoobar-usds: After a conversation with the content stakeholder, @KameronKerger, we have settled on implicit column order within datasets and dataset order based on their position in the `datasets.yml` file. The index columns, like GEOID, county, DAC and demographic data will come first.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Got it! Will update.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: Will pair with @emma-nechamkin on setting these up.

emma-nechamkin: I don't really understand what this means? We have fields that go to percentiles, reverse percentiles, and booleans. Low median income is a reverse percentiles -- anything with "low" is. Booleans are for FUDS, eAMLIS, and HRS data. The other special case here is imputation. It will be hard to do the imputation based on geographic neighbors here.

esfoobar-usds: > I don't really understand what this means? We have fields that go to percentiles, reverse percentiles, and booleans. Low median income is a reverse percentiles -- anything with "low" is. Booleans are for FUDS, eAMLIS, and HRS data. The other special case here is imputation. It will be hard to do the imputation based on geographic neighbors here.

I have no idea what this means and that's why I'd be grateful if @emma-nechamkin and @lucasmbrown-usds hash this part out.

data/data-pipeline/data_pipeline/etl/base.py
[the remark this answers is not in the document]
lucasmbrown-usds: See https://github.com/usds/justice40-tool/issues/1806 and https://github.com/usds/justice40-tool/issues/1805

lucasmbrown-usds:
I made a few edits, because they all kind of went together because I was trying to fill out all the info for each of the fields in the refactor. 

Here are the edits I made!

1. Short/long description

I moved:
```
            description_short: Optional[str]
            description_long: Optional[str]
```

to be on the LoadFields rather than the overall data class. I also made them optional. Not all of these fields make it to the front-end, so it doesn't seem like we need a short/long description for all of them. 

2. CSV/Excel 

I collapsed include_in_csv and include_in_excel to one variable named include_in_downloadable_files as @esfoobar-usds suggested. 

3. Column order 

I got rid of column order. As discussed and agreed in a prior meeting, going forward after the refactor there will be like 4-5 columns that are manually specified as the first columns to appear in the CSV/XLS, and then everything else is sorted alphabetically. So we no longer need this field.

4. Default value of `number_of_decimals_in_output` 

I created a default value of `number_of_decimals_in_output` as 2. 

5. More optional parameters and default values. 

Since we're tilting towards mostly optional parameters, I made `include_in_tiles` and `include_in_downloadable_files` optional too. 

I set default values for all of our optional parameters, mostly to `False`, so if you don't actively specify them as true, they won't be included in either. For instance, if you don't specify `create_percentile=True`, then it will not create a percentile for that field (because `create_percentile=False`).

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
esfoobar-usds: @lucasmbrown-usds fixed this.

data/data-pipeline/data_pipeline/etl/score/schemas/datasets.py
esfoobar-usds: Fixed.

esfoobar-usds:
All looks good to me @lucasmbrown-usds and I fixed the last comment, so if you're okay with it, I can merge as soon as tests pass.

lucasmbrown-usds (approved):
LGTM! Thanks for fixing that last thing, @esfoobar-usds

esfoobar-usds:
And 126 later, merging this PR!

PR #1551, opened 2022-04-12
Add hash to  send feedback email

vim-usds:
# Purpose
The send feedback button creates an email with census tract info. It's sometimes hard to find the exact census tract. This PR will add the URL hash to the subject line allowing the support team to find the census tract with more ease.  This will place the URL hash in the subject line of the email

<details>
This PR will 
- detected bug with zoom/lat/lng values and created ticket 1550
- get realtime hash from mapRef
- pass hash to AreaDetail as prop
- update tests and snapshots
</details>

## QA testing
What should happen is that when the Send Feedback button is clicked in the side panel, it grabs the URL hash (at that exact moment in time) and places it in the subject line. If the map is changed (e.g. zooming in/out or panning) after the button is clicked, the new URL hash will not be captured in the email. 

### QA spec
- [ ] does the email provide the correct hash?
- [ ] if you move around the map and select and deselect various tracts, does the email still show the correct (ie latest) hash?
- [ ] given the hash, can the support team re-create the URL under question?

### DIFF
--- a/client/src/components/AreaDetail/AreaDetail.tsx
+++ b/client/src/components/AreaDetail/AreaDetail.tsx
@@ -21,6 +21,7 @@ import mailIcon from '/node_modules/uswds/dist/img/usa-icons/mail_outline.svg';
 
 interface IAreaDetailProps {
   properties: constants.J40Properties,
+  hash: string[],
 }
 
 /**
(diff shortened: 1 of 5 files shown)


esfoobar-usds (approved):
🍏

katherinedm-usds:
Works as described and as expected! sends the z/lat/long in the subject that I can then paste into a map URL. Adjusts the z/lat/long based on what's in the browser while still sending the same census tract ID, even if I pan so long as I don't select a different tract.

image

PR #1559, opened 2022-04-14
Imputing income using geographic neighbors

emma-nechamkin:
This includes code to impute income as follows:
1. Remove college population from below 200% FPL share
2. Re-calculate income percent
3. Impute missing income based on neighbors

This PR also (annoyingly) has CRE comparisons. I accidentally pushed that branch here.


emma-nechamkin:
This error, I think, is because the US.JSON file doesn't exist necessarily. I've added code to pull it down from s3 if it doesn't exist; this code isn't awesome and I'm interested in suggestions to improve it.

data/data-pipeline/data_pipeline/content/config/csv.yml
```diff
@@ -26,6 +26,9 @@ fields:
   - score_name: Total population
     label: Total population
     format: float
+  - score_name: Percent of individuals below 200% Federal Poverty Line, imputed and adjusted
```
emma-nechamkin, line 29: Added to download data.

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -1,14 +1,26 @@
...
+
+from data_pipeline.utils import get_module_logger, unzip_file_from_url
 from data_pipeline.score import field_names
 
 logger = get_module_logger(__name__)
 
+# because now there is a requirement for the us.json, this will port from
+# AWS when a local copy does not exist.
+# @Jorge -- is this too much of a hack?
+CENSUS_DATA_S3_URL = settings.AWS_JUSTICE40_DATASOURCES_URL + "/census.zip"
```
emma-nechamkin, line 22: Jorge -- this is not the most elegant. I'm wondering if you had a better solution?

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -361,22 +440,56 @@ def transform(self) -> None:
             1 - df[self.COLLEGE_ATTENDANCE_FIELD]
         )
 
-        # strip columns
-        df = df[self.COLUMNS_TO_KEEP]
+        # We preserve the null character of missing income even when college
```
emma-nechamkin, line 443: this is not the cleanest code. any tips?

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl_utils.py
```diff
@@ -18,6 +19,113 @@ def _fips_from_censusdata_censusgeo(censusgeo: censusdata.censusgeo) -> str:
     return fips
 
 
+def impute_by_geographic_neighbors(
```
emma-nechamkin, line 22: This is also not so clear. Open to any tips!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -405,6 +405,7 @@ def _prepare_initial_df(self) -> pd.DataFrame:
             df[field_names.MEDIAN_INCOME_FIELD] / df[field_names.AMI_FIELD]
         )
 
+        # QQ: why don't we just filter to the numeric columns by type?
```
emma-nechamkin, line 408: this is a question about design decisions

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl_utils.py
emma-nechamkin: This is so confusing. Basically, we want to pull the value for each tract from the value in neighbor tracts.  When there's no values in the neighbor tracts, it pulls the county values. If there are no county values, it pulls the state values. 

It's expensive to compute neighbors -- here, because we are not imputing that many tracts, I don't make a geospatial index or anything. As an aside: When we go to do this for the donut holes, since we will have to find neighbors for ~30k tracts, it will be super expensive and so I had planned to use states first as a simpler version of a geospatial index.

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -361,22 +440,56 @@ def transform(self) -> None:
...
+        # attendance is not null (otherwise this would be hard set to 0)
+        # and left-cut-off values at 0%
+        logger.info("Adjusting income information")
+        df[
```
emma-nechamkin, line 447: Note to self: I am wondering if it makes more sense to impute both raw income and raw college students, and then compute from what's imputed. Since the vast majority of tracts missing income are also missing college students, this won't be a particularly big shift, but may help smooth high college / low income places. 

Here, as a reminder, we make the assumption that 100% of college students are earning below 200FPL (or $34000). This is slightly below the median income of full time workers who have some college but no degree (https://nces.ed.gov/programs/coe/pdf/coe_cba.pdf)

emma-nechamkin:
I believe I have fixed this -- see a lower higher ed threshold below than what we might take. Woohoo!
Screen Shot 2022-04-20 at 6 50 22 PM

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -188,18 +220,50 @@ def __init__(self):
...
+        self,
+        df: pd.DataFrame,
+        usa_geo_df: gpd.GeoDataFrame,
+        geoid_field: str = "GEOID10",
+        geometry_field: str = "geometry",
+        state_code_field: str = "STATEFP10",
+        county_code_field: str = "COUNTYFP10",
+    ) -> gpd.GeoDataFrame:
+        usa_geo_df[geoid_field] = (
+            usa_geo_df[geoid_field].astype(str).str.zfill(11)
```
emma-nechamkin, line 250: Is there a reason we don't just read this in as a string?

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
```diff
@@ -227,6 +291,27 @@ def transform(self) -> None:
...
+                self.DATA_PATH,
+            )
+
+        geo_df = gpd.read_file(
+            self.DATA_PATH / "census" / "geojson" / "us.json"
+        )
+        df = self._merge_geojson(
+            df=df,
+            usa_geo_df=geo_df,
+        )
```
emma-nechamkin, line 314: line 323: is there a reason we don't just read nulls as -666666666?

esfoobar-usds (approved):
LGTM! 🍏

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl.py
emma-nechamkin: Yep! I guess we can address it though when the YML stuff comes in

data/data-pipeline/data_pipeline/etl/sources/census_acs/etl_utils.py
emma-nechamkin: Yeah, potentially -- like clusters of neighbors or a geospatial index as a first pass.

PR #1563, opened 2022-04-15
Updating higher education threshold

emma-nechamkin:
Updated higher education threshold to reflect 80% or more college students as qualifier.


emma-nechamkin:
Closing for now as no longer relevant (pending final checks)

Issue #1529, opened 2022-04-05
Make shapefile columns.csv use column descriptions consistent with communities xlsx/csv YAML


katherinedm-usds:
To help people use the shapefile https://justice40-data.s3.amazonaws.com/data-pipeline/data/score/shapefile/usa.zip, and help us give shapefile users better instructions for using, the Columns.csv in that usa.zip should have descriptions that match the column name in the communities xlsx/csv files.

In case it's helpful for reference, attached is a recent Columns.csv from the shapefile (I needed to change the .csv attachment file type to .txt for the purposes of uploading to this issue, can be renamed to .csv after downloading)
[https://app.zenhub.com/files/358242767/e2198b85-a712-4fca-99e5-b5ec989b1125/download](https://app.zenhub.com/files/358242767/e2198b85-a712-4fca-99e5-b5ec989b1125/download)

Also, here's Excel file I have with the communities xlsx/csv column names transposed and listed in column A so that these column names can be listed alphabetically in case it's helpful. [https://app.zenhub.com/files/358242767/3d1a5107-e5ce-4288-b4f5-662589c14034/download](https://app.zenhub.com/files/358242767/3d1a5107-e5ce-4288-b4f5-662589c14034/download) 

Ideally this could be fixed soon enough to give people the chance to interact with the CEJST data via the shapefile before end of public beta period.

katherinedm-usds:
I downloadedd

PR #1573, opened 2022-04-18
Adding updated codebook to shapefile zip

emma-nechamkin:
This PR adds the csv column descriptions to the shapefile, while keeping original column names as well (in case we refactor and change the names). 

This is a little bit clumsy in that it takes place in etl score geo, but there isn't really a great place to put it right now.

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
+++ b/data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
@@ -12,7 +12,13 @@ from data_pipeline.etl.sources.census.etl_utils import (
 )
 from data_pipeline.etl.score.etl_utils import check_score_data_source
 from data_pipeline.score import field_names
-from data_pipeline.utils import get_module_logger, zip_files
+from data_pipeline.content.schemas.download_schemas import CSVConfig
+from data_pipeline.utils import (
+    get_module_logger,
+    zip_files,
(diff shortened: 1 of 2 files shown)


esfoobar-usds (approved):
🍏

katherinedm-usds:
Hi, I might not have been looking at the right links just now (I tried https://screeningtool.geoplatform.gov/en/methodology/?flags=stage_hash=1573/ab973fcabcf2cc909119ea4c7f7ecc67b88e0d6d and https://screeningtool.geoplatform.gov/en/methodology to download the shapefile) but columns.csv rows referencing internal names i.e. "Water Factor (Definition M)" are still there, so I'm not yet able to point shapefile users to a columns.csv that can use for their analysis. Can #1529 be updated with the link for QA'ing this? Thanks!

emma-nechamkin:
You can see that when we have a column name specified in the yaml config (see the circled bit), it propagates for me.

emma-nechamkin:
I restored the branch, but am wondering if this is an update thing? The staging link suggests the data has not been updated for download since 4/6.

esfoobar-usds:
I guess the moral of this one -- let's be more explicit when we want product QA on a branch so that we (devs) don't merge before QA is completed.

emma-nechamkin:
thank you both!!!

PR #1579, opened 2022-04-19
Adding HOLC indicator

emma-nechamkin:
Right now, this PR creates the redline share indicator WITHOUT pulling it through the score


esfoobar-usds (approved):
🍏

emma-nechamkin:
This should be added back in as of 8/2

PR #1586, opened 2022-04-26
Updating EJScreen data

emma-nechamkin:


vim-usds:
@emma-nechamkin - this seem to have a number of other commits in this PR. Is that purposeful?

vim-usds:
Also, it seem that the back end staging script is failing on the last run, should we make sure that passes?

emma-nechamkin:
> @emma-nechamkin - this seem to have a number of other commits in this PR. Is that purposeful?

no definitely not! also these files did not match what I have locally o.O. Updates pending

emma-nechamkin:
@vim-usds i created a new base instead of rebasing this :/

PR #1590, opened 2022-04-27
Put codebook into <callout></callout> section for Download description

TomNUSDS:
I missed this the first time around.

Also, updated intl and tests snapshots.


TomNUSDS:
@esfoobar-usds bump.

esfoobar-usds (approved):
🍏

PR #1618, opened 2022-05-04
Columnn headers update

### LINKED ISSUE #1607: Column headers need updating
**Describe the bug**
A few outstanding issues with the column headers. Assumption is that they are the same in the .csv and the .xlsx. These may all be fixable via the YAML file?
- [x] higher ed, included with every boolean needs to be changed from "and has a low percent of higher ed students?" to "a high percent of residents that are not higher ed students?"
- [x] Proximity to NPL sites (percentile) should be Proximity to NPL (Superfund) sites (percentile)
- [x] Proximity to NPL sites should be Proximity to NPL (Superfund) sites
- [x] Percent of individuals < 100% Federal Poverty Line needs to move one column to the left
- [x] ~All of the island data is missing, it is just the booleans~ -- created #1617 

**To Reproduce**
1. Go to '.xlxs or .csv'
2. See errors

**Expected behavior**
See above

**Screenshots**

**Additional context**

esfoobar-usds:
Fixes #1607


vim-usds (commented):
@esfoobar-usds - was expecting the en.json to be updated that exists in the `intl` folder. I'm assuming that the `compile` command was run. We may want to run `extract` as well or instead of to ensure the en.json file that the translation team expects also gets updated. Thanks!

esfoobar-usds:
@vim-usds ready for re-review!

vim-usds:
@esfoobar-usds - still don't see the en.json? There should be a git hook that is run automatically to create it, without having to run anything extra. Maybe we should jump on a zoom?

esfoobar-usds:
@vim-usds I ran `intl:extract`, but it seems it's not picking up any changes:

```
npm run intl:extract

> justice40-tool@1.0.0 intl:extract
> formatjs extract 'src/**/*.tsx' --out-file src/intl/en.json
```

vim-usds (approved):

PR #1674, opened 2022-06-08
Update downloadable assets URL

esfoobar-usds:


KameronKerger:
:green_apple: thanks for the fix @esfoobar-usds

PR #1681, opened 2022-06-15
Do not drop Guam and USVI from ETL

switzersc-usds:
Closes 1662.

This removes code that drops these two territories from the ETL and subsequent score data generation process.

Notably, tests are not currently passing on my machine (and they didn't pass before making the change). I need to figure that out before this can be fully done, assuming that this PR's build will fail when it runs the python tests.

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/score/constants.py
+++ b/data/data-pipeline/data_pipeline/etl/score/constants.py
@@ -99,6 +99,9 @@ ISLAND_AREAS_EXPLANATION = (
 # Column subsets
 CENSUS_COUNTIES_COLUMNS = ["USPS", "GEOID", "NAME"]
 
+# Drop FIPS codes from map
+DROP_FIPS_CODES = []
+
 # Drop FIPS codes from incrementing
 DROP_FIPS_FROM_NON_WTD_THRESHOLDS = "72"
 
--- a/data/data-pipeline/data_pipeline/etl/score/constants.py
+++ b/data/data-pipeline/data_pipeline/etl/score/constants.py
@@ -99,9 +99,6 @@ ISLAND_AREAS_EXPLANATION = (
 # Column subsets
 CENSUS_COUNTIES_COLUMNS = ["USPS", "GEOID", "NAME"]
 
-# Drop FIPS codes from map
-DROP_FIPS_CODES = ["66", "78"]
-
 # Drop FIPS codes from incrementing
 DROP_FIPS_FROM_NON_WTD_THRESHOLDS = "72"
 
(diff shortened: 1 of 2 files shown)


data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -99,9 +99,6 @@
 # Column subsets
 CENSUS_COUNTIES_COLUMNS = ["USPS", "GEOID", "NAME"]
 
-# Drop FIPS codes from map
-DROP_FIPS_CODES = ["66", "78"]
```
esfoobar-usds, line 103: Leave this as an empty list `[]`

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -254,32 +254,6 @@ def _create_tile_data(
...
-
-        score_tiles[constants.TILES_SCORE_FLOAT_COLUMNS] = score_tiles[
-            constants.TILES_SCORE_FLOAT_COLUMNS
-        ].apply(
-            func=lambda series: floor_series(
-                series=series,
-                number_of_decimals=constants.TILES_ROUND_NUM_DECIMALS,
-            ),
-            axis=0,
-        )
```
esfoobar-usds, line 281: I would leave this code block and just update the comments on 257 - 258 saying something like "these are tracts we drop from the score" or something generic like that.

esfoobar-usds (changes requested):
Even though we won't use the drop tracts functionality, I'd still leave the code in just in case we need it down the road.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
switzersc-usds: Oh I see, so leave everything but make the constant be an empty array. Sounds good!

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -100,7 +100,7 @@
 CENSUS_COUNTIES_COLUMNS = ["USPS", "GEOID", "NAME"]
 
 # Drop FIPS codes from map
-DROP_FIPS_CODES = ["66", "78"]
+DROP_FIPS_CODES = []
```
emma-nechamkin, line 103: i think this is the only change and it looks right to me! did you confirm that in the output, we see fips 66 and 78 with t&wd?

emma-nechamkin (approved):
I am not sure if I have reviewing power, but this looks right to me. Did you confirm in the output? (I assume so but might be worth a 2 min check?)

esfoobar-usds (approved):
🍏

data/data-pipeline/data_pipeline/etl/score/constants.py
switzersc-usds: i haven't! what's the simplest way to do that? when i run etl score post locally, i don't see where the output is.

emma-nechamkin: I think that runs all the way through to the downloadable files -- and if this is in the downloadable files, that is great news! I'd check here `data_pipeline/data/score/downloadable/`

PR #1708, opened 2022-07-01
Add CSP plugin

vim-usds:
- OWASP flagged a minor issue with CSP. Checking if this will fix the issue


vim-usds:
Adding CSP via the gatsby plugin break the site

Issue #1720, opened 2022-07-07
Build function that converts latitude/longitude pairs into census tract proximity metrics


lucasmbrown-usds:
**Is your feature request related to a problem? Please describe.**
There are a number of data sets that have latitude and longitude of specific sites in them -- such as eAMLIS data (https://amlis.osmre.gov/Map.aspx) which maps the lat/long of [Abandoned Mine Lands](https://amlis.osmre.gov/). 

For any data of that type to be included in the tool, it will need to be converted from lat/long into a measure of proximity per census tract. Census tracts obviously have a polygon boundary. 

The method can initially be built with just one metric of measuring distance between each census tract and all nearby lat/long points. Eventually it would be great to support multiple metrics for calculating this distance, which may be used differently for different sites. 

These measurements can include: 

1. What's the count of sites per census tract (integer > 0)? Can be easily calculated using something like https://stackoverflow.com/a/65120026. 
2. A more complex measure such as the mean inverse squared distanced from the site to any of the boundaries of the census tract. (Such as https://gis.stackexchange.com/a/342489 or https://gis.stackexchange.com/a/375512)
3. The number of such sites within K miles of any of the tract's boundaries. 
4. The number of such sites within K miles of the centroid of the tract (https://stackoverflow.com/a/72384543). 

There is a research project about the appropriate way to calculate this distance. 

For instance, imagine the following two lat/long points (red circles) amongst a bunch of census tracts: 

<img width="428" alt="Screen Shot 2022-07-07 at 11 16 33 AM" src="https://user-images.githubusercontent.com/25443529/177809785-8663a15e-7704-4b4a-bc38-61ae534bc66d.png">

We need a method to calculate the distance of each tract to the specific lat/long points (red circles). For instance, tract 031637 (in the middle right) is a short distance from the northmost red circle and a medium distance from the bottommost red circle. We want to create a single metric/score for each tract's distance from all of the specific lat/long points. 

We should make this method generic, so that it can be used on a variety of data sources that have lat/long coordinates. We can assume the data will be in a dataframe with one row per location, with one column for lat and one column for long.

**Describe the solution you'd like**
Build a reusable method that takes 2010 Census Tract boundaries (TIGER shapefiles) and an arbitrary list of lat/long points and a chosen type of distance calculation and calculates a single measure of distance between each census tract and all lat/long points.

The tract boundaries will rarely change, so do not necessarily need to be passed as parameters to the function/method.

[...]

PR #1737, opened 2022-07-11
Issue 1720: Converting lat/lon dataframe to tracts

lucasmbrown-usds:
Work in progress on #1720.


emma-nechamkin (commented):
I have some extremely annoying thoughts here -- apologies! It's a bit hard to comment in line in notebooks, but a few thoughts: 

1. For when you drop duplicates by lat/long and sometimes arbitrarily choose a row, I'm wondering why not `groupby([lat, long])[value_of_interest].size()` or something similar to get the count of mines? (with the caveat that like... should there really be many mines within a single lat/long?). The other thought is that by grouping by rather than dropping duplicates, you can aggregate easily mine severity in some way. 
2. I *think* when you go to match points to polygons, it's faster to do a spatial join (`predicate="within"`) and left join from mines to score[["tract", "geometry"]]. This is because spatial joins use an rtree, I think? (I've been playing around here with `predicate=touches` for the donut holes).

lucasmbrown-usds:
@emma-nechamkin this has been updated to include FUDS as well. 

Both your suggestions are great! On #1, we should investigate further into whether the early results warrant deciding about some grouping criteria. 

On #2, that's totally the right way to do it. I'm going to be lazy right now and continue with the inefficient way we have here, because it would take a (minor) refactor of a couple nested functions to iterate through polygons rather than iterating through points. And right now I can just brute force it in the background while doing something else for 20 minutes. But let's swap it out if we end up using this frequently. 

Feel free to pull locally and make sure it works for you!

emma-nechamkin:
@mattbowen-usds has implemented this in our codebase; closing this PR

PR #1740, opened 2022-07-12
Fix on large AK tracts that are off screen

### LINKED ISSUE #1722: Large census tracts (like in Alaska) are behaving inconsistent w/ other parts of the US when you zoom out the tracts get deselected
**Describe the bug**
Go to Alaska, zoom in and select a census tract (they are very large in Alaska). When you zoom out, the tract gets deselected. This is inconsistent with other parts of the US. I think this is because the boundaries are disappearing. You get a similar behavior in Montana (see screenshots) but the tract is not actually deselected, the boundaries just disappear.

**To Reproduce**
1. Go to 'Alaska'
2. Click on 'a census tract'
3. Zoom out

**Expected behavior**
The tract should not appear to deselect.

**Screenshots**
Screen Shot 2022-07-07 at 8 34 03 AM
Screen Shot 2022-07-07 at 8 33 56 AM
Screen Shot 2022-07-07 at 8 32 41 AM
Screen Shot 2022-07-07 at 8 32 33 AM

**Additional context**

vim-usds:
# Purpose
1. closes #1722

## Implementation notes
- change the low to high transition from 7 to 5. This can not go any lower as high tiles on AWS only go to zoom level 5. The python notebook used for this is also committed.
- reduce the zoom level globally on all census tracts
- refactor map onClick
- The super large (less than 5 tracts) will still be off screen because they require zoom level 4 which is unavailable without a backend change

### QA link [here](http://usds-geoplatform-justice40-website.s3-website-us-east-1.amazonaws.com/justice40-tool/1740-c75b6e/en/#3/33.47/-97.5)

QA spec
- [x] is a larger % of selected tracts within the viewport area of the map now? (Not at 100% yet)
- [x] territory buttons zoom level has not been affected by this
- [x] Search should not have been affected by this


esfoobar-usds (approved):
🍏

vim-usds:
> I am not sure that this addresses the issue but it does work better in terms of zooming on / in Alaska. the issue was more that the census tract gets deselected when you zoom out (I think this is because the tract boundaries disappear at that level). this makes it hard to orient yourself in Alaska because you are constantly having to zoom out to find markers within the census tract and then it zooms you back in when you select. this might not be something we want to address beyond the fix that you have here @vim-usds.

@KameronKerger - yes exactly!

PR #1742, opened 2022-07-13
Emma nechamkin/holc patch

emma-nechamkin:


vim-usds:
@emma-nechamkin - it looks like we're just taking something out of the narwhal score so is it safe to assume no frontend work needed for this at this time?

emma-nechamkin:
> @emma-nechamkin - it looks like we're just taking something out of the narwhal score so is it safe to assume no frontend work needed for this at this time?

100% yes! just backend rn

vim-usds (approved):
LGTM

PR #1743, opened 2022-07-13
Updating housing burden

emma-nechamkin:
Instructions are: 
```
Adjustment to existing housing cost burden indicator to capture percentage of renter households that are both earning less than 50% of HUD Area Median Family Income by county and are spending more than 50% of their income on housing costs. HUD made this recommendation.
```

I interpreted as:
- do not touch owners 
- do not touch places where HAMFI isn't computed
- remove only for renters places where hamfi exceeds 50% but we were including them

I have taken that the existing calculations are correct and only removed these columns from the numerator for renters.


data/data-pipeline/data_pipeline/etl/sources/hud_housing/etl.py
```diff
@@ -13,7 +13,8 @@ def __init__(self):
         self.HOUSING_ZIP_FILE_DIR = self.get_tmp_path() / "hud_housing"
 
         # We measure households earning less than 80% of HUD Area Median Family Income by county
-        # and paying greater than 30% of their income to housing costs.
+        # and paying greater than 30% of their income to housing costs, with the exception of renters.
```
lucasmbrown-usds, line 16: Shouldn't this be `and paying greater than 50% of their income on housing costs`?

emma-nechamkin: yes!

emma-nechamkin:
(Waiting on HUD to clarify guidance)

emma-nechamkin:
From HUD: 

```
I can’t speak for whoever from HUD was giving comment previously, but very low income (50% HAMFI) renter households with severe cost burden is a population of specific interest for the department. Our Worst Case Needs reports, for instance, only consider VLI renter households. VLI renter households with severe cost burden (or overcrowding) is also used as a proxy for households at risk of homelessness in some of our formula allocations and studies. Owner households are excluded in these applications, so it’s not unprecedented. I would say you should either just include 50/50 renters or apply the 50/50 standard to both renters and owners (you’ll find that it will identify far fewer owners, which is probably one reason they’re often simply excluded.) I don’t think applying a different and less stringent standard for owners would be appropriate.
```

For consistency, I excluded owners.

emma-nechamkin:
@lucasmbrown-usds Are we going with this or the original?

emma-nechamkin:
As of 8/3, this is NOT going to be merged in.

Issue #1748, opened 2022-07-20
Adjust linguistic isolation percentiles


emma-nechamkin:

emma-nechamkin:
Because we are now dropping PR from Linguistic Isolation, we should calculate percentiles with PR tracts excluded

PR #1750, opened 2022-07-21
Rescaling linguistic isolation 

emma-nechamkin:
removing PR from linguistic isolation and reconstructing percentiles. #1748 

This also edits the existing code to handle both edge cases (linguistic isolation and ag value) in a single way.


data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -310,61 +316,21 @@ def _add_percentiles_to_df(
 
         else:
             # For agricultural loss, we are using whether there is value at all to determine percentile and then
```
lucasmbrown-usds, line 318: Is this comment now misplaced because it refers to agricultural loss?

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -310,61 +316,21 @@ def _add_percentiles_to_df(
...
-            )
-
-            # Calculate the percentile for only this category
-            this_category_only_percentile_field = (
-                f"{output_column_name_root} "
-                f"(percentile {urban_or_rural_string} only)"
-            )
-            df[this_category_only_percentile_field] = df[
-                this_category_only_value_field
+            ] = df.groupby(df[field_names.GEOID_TRACT_FIELD].isin(drop_tracts))[
```
lucasmbrown-usds, line 324: Can you add a comment explaining this code? Maybe I'm just not fully awake this morning but I have honestly no idea what's going on in this code. 

Here's my cro-magnon attempt to understand what's happening: 

1. You're limiting the data frame to only include tracts that you're intentionally trying to drop? Why, wouldn't it be the other way? 

2. Then you're running a groupby -- I guess grouping by each of the tracts that are in `drop_tracts`? Why? 

3. Then you're only selecting one column (the input), makes sense
4. Running a percentile on this 
5. Setting percentile to the output column name / percentile field. 

I'm honestly just very confused. It feels like maybe you're using the groupby as a "shortcut" to do something that's not traditionally an aggregation. But this code feels kind of error prone because you're packing so much into a one liner.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -310,61 +316,21 @@ def _add_percentiles_to_df(
...
-            )
-
-        # Combine both urban and rural into one field:
-        df[
-            f"{output_column_name_root}{field_names.PERCENTILE_URBAN_RURAL_FIELD_SUFFIX}"
-        ] = df[urban_rural_percentile_fields_to_combine].mean(
-            axis=1, skipna=True
-        )
+            df.loc[
+                df[field_names.GEOID_TRACT_FIELD].isin(drop_tracts),
```
lucasmbrown-usds, line 330: Can you add some comments about this as well? I'm confused about what's going on. 

Why are you running the percentile *before* setting the values to `None`, and then afterwards setting it to `None` for those drop_tracts?

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -523,13 +489,47 @@ def _prepare_initial_df(self) -> pd.DataFrame:
         df_copy[numeric_columns] = df_copy[numeric_columns].apply(pd.to_numeric)
 
         # Convert all columns to numeric and do math
+        # Note that we have a few special conditions here, that we handle explicitly.
+        #     For *Linguistic Isolation*, we do NOT want to include Puerto Rico in the percentile
+        #     calculation. This is because linguistic isolation as a category doesn't make much sense
+        #     in Puerto Rico, where Spanish is a recognized language. Thus, we construct a list
+        #     of tracts to drop from the percentile calculation.
+        #
+        #     For *Expected Agricultural Loss*, we only want to include in the percentile tracts
```
lucasmbrown-usds, line 498: Great comments!

lucasmbrown-usds (changes requested):
I might not have had enough coffee this morning!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
emma-nechamkin: yes! so it's a groupby percentile, so the percentiles get calculated separately (e.g., the drop tracts get their own percentiles), and then we get rid of the percentiles for the droptracts. The idea here is that we want to preserve the raw data for the drop tracts, even if we don't keep the percentiles. Does that make sense to you?

data/data-pipeline/data_pipeline/etl/score/etl_score.py
emma-nechamkin: No, because agricultural loss is one of the two cases covered here. Does that make sense to you? I can adjust the comment to make it clearer

lucasmbrown-usds: Yeah it seems like some adjustment could be useful. It also may not be necessary to mention the specific use cases *inside* this method. Because you've abstracted this all to `drop_tracts` I don't think you need to mention the use cases anymore.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
emma-nechamkin: Going with option 1

data/data-pipeline/data_pipeline/etl/score/etl_score.py
emma-nechamkin: Updated!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
emma-nechamkin: Updated!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -298,73 +299,39 @@ def _add_percentiles_to_df(
...
-            )
-
-        # Combine both urban and rural into one field:
-        df[
-            f"{output_column_name_root}{field_names.PERCENTILE_URBAN_RURAL_FIELD_SUFFIX}"
-        ] = df[urban_rural_percentile_fields_to_combine].mean(
-            axis=1, skipna=True
-        )
+            # Check that "drop tracts" were dropped (quicker than creating a fixture?)
+            assert df[df[field_names.GEOID_TRACT_FIELD].isin(drop_tracts)][
```
emma-nechamkin, line 331: Added this instead of a formal test -- this sort of test doesn't seem fit nicely within our fixture testing framework, and these sort of assert statements are compatible / match the rest of the codebase.

lucasmbrown-usds: Makes sense! Appreciate the spirit of testing even without a formal testing framework.

lucasmbrown-usds (approved):
This is much much easier to read for this dumb dumb. Nice!

PR #1768, opened 2022-07-25
Update Cypress GHA to v4

vim-usds:
- [x] bump to v4
- [x] check if GHA test passes

### DIFF
--- a/client/src/data/blankCEJST/index.html
+++ b/client/src/data/blankCEJST/index.html
@@ -14,6 +14,6 @@
 </head>
 
 <body>
-  <!-- your content here... adding comment to trigger build-->
+  <!-- your content here... changing comment to trigger build-->
 </body>
 </html>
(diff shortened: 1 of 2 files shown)


client/src/data/blankCEJST/index.html
```diff
@@ -14,6 +14,6 @@
 </head>
 
 <body>
-  <!-- your content here... adding comment to trigger build-->
+  <!-- your content here... changing comment to trigger build-->
```
TomNUSDS, line 17: Probably not needed

TomNUSDS (approved):
LGTM.

Maybe remove the touch-edit to index.html? Not required at all though.

client/src/data/blankCEJST/index.html
vim-usds: The GHA requires some change in the client folder to trigger the build. I first ran it without it. Then placed it in as it wasn't triggered the GHA.

PR #1817, opened 2022-08-10
Add FUDS ETL

mattbowen-usds:
Closes #1871


data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
```diff
@@ -0,0 +1,103 @@
...
+        # Note that the length of raw_df will not be exactly the same
+        # because same bases lack coordinated or have coordinates in
+        # Mexico or in the ocean. See the following dataframe:
+        # raw_df[~raw_df.OBJECTID.isin(df_with_tracts.OBJECTID)][
+        # ['OBJECTID', 'CLOSESTCITY', 'COUNTY', 'ELIGIBILITY',
+        # 'STATE', 'LATITUDE', "LONGITUDE"]]
+        df_with_tracts = add_tracts_for_geometries(raw_df)
+
+        # Now, do some counting
+        # XXX: @emma-nechamkin do the conditions for has_eligible_projects and
```
mattbowen-usds, line 69: @emma-nechamkin wanted to flag this section for you to confirm this is the logic you expect re `ELIGIBILITY` and `HASPROJECTS` --- lemme know if i misinterpreted and i can change

data/data-pipeline/data_pipeline/tests/sources/us_army_fuds/data/output.csv
```diff
@@ -0,0 +1,16 @@
+GEOID10_TRACT,Count of eligible Formerly Used Defense Site (FUDS) properties centroids,Count of ineligible Formerly Used Defense Site (FUDS) properties centroids,Boolean expressing whether any eligible Formerly Used Defense Sites (FUDS) have centroids in the tract
+06027000800,3.0000000000,14.0000000000,True
```
mattbowen-usds, line 2: I just realized these should be ints, not floats, so I need to add a line somewhere.

data/data-pipeline/data_pipeline/tests/sources/us_army_fuds/test_etl.py
```diff
@@ -0,0 +1,171 @@
...
+        etl = self._ETL_CLASS()
+        data_path, tmp_path = mock_paths
+
+        output_file_path = etl._get_output_file_path()
+        expected_output_file_path = (
+            data_path / "dataset" / self._ETL_CLASS.NAME / "usa.csv"
+        )
+        assert output_file_path == expected_output_file_path
+
+    def test_fixtures_contain_shared_tract_ids_base(self, mock_etl, mock_paths):
```
mattbowen-usds, line 125: There are all overridden to be able to mock out `add_tracts_for_geometries`, since loading the tracts over and over is slow, and even if I do some caching it's still slower than we really want in tests.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -117,6 +117,34 @@ datasets:
...
+        df_field_name: "INELIGIBLE_FUDS_COUNT_FIELD_NAME"
+        long_name: "Count of ineligible Formerly Used Defense Site (FUDS) properties centroids"
+        description_short: 
+          "The number of FUDS marked as Ineligible or Project in the tract."
+        field_type: int64
+        include_in_tiles: true
+        include_in_downloadable_files: true
+      - short_name: "is_eligibile"
+        df_field_name: "ELIGIBLE_FUDS_BINARY_FIELD_NAME"
+        long_name: "Boolean expressing whether any eligible Formerly Used Defense Sites (FUDS) have centroids in the tract"
```
emma-nechamkin, line 142: Nit: "Is there at least one Formerly Used Defense Site (FUDS) in the tract?"

mattbowen-usds: That IS better!

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
```diff
@@ -0,0 +1,106 @@
...
```
emma-nechamkin, line 69: This is correct! Thanks for checking.

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
```diff
@@ -0,0 +1,106 @@
...
+        # has_ineligble_projects look like what you were thinking?
+        df_with_tracts.loc[
```
emma-nechamkin, line 71: This might be easier and faster to do with a `where` call. As in `df[new_col] = np.where(CONDITIONS, 1, 0)`. Then you wouldn't need to `fillna(inplace=True)`, which can get ~dicey~.  Just a thought!

mattbowen-usds: Thank you! TIL!

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
```diff
@@ -0,0 +1,106 @@
...
+                & (df_with_tracts.HASPROJECTS == "Yes")
+            ),
+            "has_eligible_projects",
+        ] = 1
+        df_with_tracts.has_eligible_projects.fillna(0, inplace=True)
+        df_with_tracts.loc[
+            (
+                (df_with_tracts.ELIGIBILITY == "Inligible")
+                | (df_with_tracts.HASPROJECTS == "No")
+                | (df_with_tracts.ELIGIBILITY.isna())
```
emma-nechamkin, line 83: I think the cleanest way to do this would be to use a single variable ("has eligible projects", created using `where`) and then groupby that variable with a `size` aggregation. It would cut out a bunch of lines and maybe make this code easier to read? But 🚨 I am not an engineer 🚨 and this is just a thought

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
```diff
@@ -0,0 +1,106 @@
...
+        ] = 1
+        df_with_tracts.has_ineligible_projects.fillna(0, inplace=True)
+        self.output_df = (
+            df_with_tracts.groupby("GEOID10_TRACT")[
+                ["has_eligible_projects", "has_ineligible_projects"]
+            ]
+            .sum()
+            .rename(
+                columns={
+                    "has_eligible_projects": self.ELIGIBLE_FUDS_COUNT_FIELD_NAME,
```
emma-nechamkin, line 96: Why do you rename rather than originally name the columns the final name?

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
mattbowen-usds: You are way better at pandas than me so I believe you!

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
mattbowen-usds: Because it was a leftover from when I was exploring! Good catch!

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
```diff
@@ -0,0 +1,106 @@
...
+                    "has_ineligible_projects": self.INELIGIBLE_FUDS_COUNT_FIELD_NAME,
+                }
+            )
+        ).reset_index()
+        for col in [self.ELIGIBLE_FUDS_COUNT_FIELD_NAME, self.INELIGIBLE_FUDS_COUNT_FIELD_NAME]:
+            self.output_df[col] = self.output_df[col].astype('int64')
```
emma-nechamkin, line 102: I think you can do this with `apply` which is *slightly* cleaner

data/data-pipeline/data_pipeline/tests/sources/us_army_fuds/test_etl.py
```diff
@@ -0,0 +1,185 @@
...
+    USArmyFUDS,
+)
+from data_pipeline.tests.sources.example.test_etl import TestETL
+from data_pipeline.utils import get_module_logger
+
+logger = get_module_logger(__name__)
+
+
+def _fake_add_tracts_for_geometries(df):
+    """The actual geojoin is too slow for tests. Use procomputed results."""
```
emma-nechamkin, line 16: typo

data/data-pipeline/data_pipeline/tests/sources/us_army_fuds/test_etl.py
emma-nechamkin: this sounds good to me!

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
```diff
@@ -0,0 +1,106 @@
...
+
+    NAME: str = "us_army_fuds"
+
+    ELIGIBLE_FUDS_COUNT_FIELD_NAME: str
+    INELIGIBLE_FUDS_COUNT_FIELD_NAME: str
+    ELIGIBLE_FUDS_BINARY_FIELD_NAME: str
+    GEO_LEVEL: ValidGeoLevel = ValidGeoLevel.CENSUS_TRACT
+
+    def __init__(self):
+        self.FILE_URL: str = (
```
emma-nechamkin, line 23: I'm late to the party here, but why isn't the URL in the YAML?

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -117,6 +117,34 @@ datasets:
...
+        include_in_downloadable_files: true
+      - short_name: "ineligibile_count"
+        df_field_name: "INELIGIBLE_FUDS_COUNT_FIELD_NAME"
+        long_name: "Count of ineligible Formerly Used Defense Site (FUDS) properties centroids"
+        description_short: 
+          "The number of FUDS marked as Ineligible or Project in the tract."
+        field_type: int64
+        include_in_tiles: true
+        include_in_downloadable_files: true
+      - short_name: "is_eligibile"
```
emma-nechamkin, line 140: Can we change this name to `has_fuds`? When we have this in combination with every other dataset, it's not going to be clear what `is_eligible` means.

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
mattbowen-usds: That I can't answer. @esfoobar-usds ?

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
mattbowen-usds: Gladly! I just guessed at those values anyway.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -117,6 +117,34 @@ datasets:
...
+    load_fields:
+      - short_name: "eligibile_count"
+        df_field_name: "ELIGIBLE_FUDS_COUNT_FIELD_NAME"
+        long_name: "Count of eligible Formerly Used Defense Site (FUDS) properties centroids"
+        description_short: 
+          "The number of FUDS marked as Eligible and Has Project in the tract."
+        field_type: int64
+        include_in_tiles: true
+        include_in_downloadable_files: true
+      - short_name: "ineligibile_count"
```
emma-nechamkin, line 132: maybe `not_fuds_ct`, same comment as below? As an aside -- since shapefiles can only have column names with 10 or fewer characters, I think we want the "meaning" of the column to be clear as early on in the name as possible?

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -117,6 +117,34 @@ datasets:
         field_type: float
         include_in_downloadable_files: true
         include_in_tiles: true
+  - long_name: "Formerly Used Defense Sites"
+    short_name: "FUDS"
+    module_name: "us_army_fuds"
+    load_fields:
+      - short_name: "eligibile_count"
```
emma-nechamkin, line 124: Perhaps just `fuds_count`? I think that for the shapefile and many dataset reasons below, might be good to include this! (unless my original hope of `[module_name]__[short_name]` is the plan :()

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -117,6 +117,34 @@ datasets:
...
+        field_type: int64
+        include_in_tiles: true
+        include_in_downloadable_files: true
+      - short_name: "ineligibile_count"
+        df_field_name: "INELIGIBLE_FUDS_COUNT_FIELD_NAME"
+        long_name: "Count of ineligible Formerly Used Defense Site (FUDS) properties centroids"
+        description_short: 
+          "The number of FUDS marked as Ineligible or Project in the tract."
+        field_type: int64
+        include_in_tiles: true
```
emma-nechamkin, line 138: False -- I don't think we need this rendered on the map nor do I think we need to export it to the end user?

emma-nechamkin (approved):
All of my changes are stylistic -- take or leave them as you see fit!

mattbowen-usds:
@emma-nechamkin they're great and thank you for them!

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
esfoobar-usds: @lucasmbrown-usds and I went back and forth about it and since it's not really a parameter you'll use in other places, we thought it belonged right and center on the ETL class itself.

data/data-pipeline/data_pipeline/etl/sources/us_army_fuds/etl.py
```diff
@@ -0,0 +1,98 @@
...
+        HAS_FUDS = "has_fuds"
+        df_with_tracts[HAS_FUDS] = np.where(
+            (
+                (df_with_tracts.ELIGIBILITY == "Eligible")
+                & (df_with_tracts.HASPROJECTS == "Yes")
+            ),
+            self.ELIGIBLE_FUDS_COUNT_FIELD_NAME,
+            self.INELIGIBLE_FUDS_COUNT_FIELD_NAME,
+        )
+        self.output_df = (
```
emma-nechamkin, line 83: this reads a little bit confusing to me personally but no pressure to conform here! The reason I think it is confusing is because the pivot from column value to column name is atypical. 

I might do something like this:

```
# this will create a boolean series which you can do actually sans np.where
df['tmp_fuds'] = (df_with_tracts.ELIGIBILITY == "Eligible") & (df_with_tracts.HASPROJECTS == "Yes")

# this gives you the count eligible
output_df[column_name_eligible] = df.groupby(tract)[tmp_fuds].sum()

# this gives you the count ineligible
output_df[column_name_ineligible] = df[~df.tmp_fuds].groupby(tract).size()

# this gives you a boolean and converts it to a 1/0 (if that's what you'd like). You can also .astype(int) it if you'd prefer
output_df[binary] = (output_df[column_name_eligible] > 0) * 1.0
```

emma-nechamkin (approved):
My comments are still stylistic so take or leave as you see fit!

mattbowen-usds:
We are having upstream errors with the census API. i have run the pipeline locally and passed it, and none of the stuff I changed broke since my successful build last night.

PR #1823, opened 2022-08-16
Adding first street foundation data

emma-nechamkin:


emma-nechamkin:
Right now, this is only adding the data sources.  They are NOT integrated into scores in any way because I want to minimize merge conflicts.

data/data-pipeline/data_pipeline/etl/score/config/datasets.yml
```diff
@@ -129,3 +129,84 @@ datasets:
         include_in_tiles: true
         include_in_downloadable_files: true
 
+  - long_name: "First Street Foundation Flood Risk"
```
esfoobar-usds, line 132: LOVE you're doing YAML like a baws.

data/data-pipeline/data_pipeline/etl/sources/fsf_flood_risk/etl.py
```diff
@@ -0,0 +1,93 @@
+# pylint: disable=unsubscriptable-object
```
emma-nechamkin, line 1: Note (namely to @lucasmbrown-usds) -- these two datasets get munged basically identically. We could move the work into a separate util function for transform, but the yaml file means that we wouldn't really save any lines of code. LMK if you think this explicit work is the wrong call.

esfoobar-usds (approved):
YAML 🚀 for Nechamkin

emma-nechamkin:
The test failures we have here are ones Matt is working through.

PR #1828, opened 2022-08-17
Add donut hole calculation to score

### LINKED ISSUE #1507: Census tracts surrounded by DACs should also be considered DACs
This issue seeks to address "donut holes" in our DAC map. This will involve the following:
1. For all census tracts that have at least one DAC neighbor, calculate the share of neighbors that are DACs
2. Over a certain threshold, those tracts will also become DACs

In analysis, I did this analysis first within state, and then, nationally for all tracts with 50% or more DAC neighbors within state. 

Some caveats:
1. First, we will have to move up some of the geocoding / reading geojson work from `etl_score_geo` to `etl_score`, since there are now parts of the score that depends on neighboring tracts
2. The exact neighbor share used depends on final analysis; we will start with a threshold of 85%

mattbowen-usds:
Closes #1507


data/data-pipeline/data_pipeline/score/utils.py
```diff
@@ -0,0 +1,50 @@
...
+    adjacent_tracts: gpd.GeoDataFrame = df.sjoin(
+        tract_data, predicate="touches"
+    )
+    return (
+        adjacent_tracts.groupby(ORIGINAL_TRACT)[[score_column]]
+        .mean()
+        .reset_index()
+        .rename(
+            columns={
+                ORIGINAL_TRACT: field_names.GEOID_TRACT_FIELD,
```
emma-nechamkin, line 46: It might make more sense to rename the OTHER frame's tract field so that you don't have to rename here (e.g., "NEIGHBORING_TRACT", keeping the right frame constant)

data/data-pipeline/data_pipeline/tests/score/test_utils/data/scores.csv
```diff
@@ -0,0 +1,10 @@
+GEOID10_TRACT,included
```
emma-nechamkin, line 1: thank you for making these!

data/data-pipeline/data_pipeline/score/utils.py
mattbowen-usds: Good call. For some reason I am addicted to unnecessary renames.

emma-nechamkin:
I did a goofy thing and just checked your branch out, Matt, rather than pull against it. But, my goal is to wrap this up tonight.

data/data-pipeline/data_pipeline/content/config/csv.yml
```diff
@@ -20,9 +20,21 @@ fields:
   - score_name: Total categories exceeded
     label: Total categories exceeded
     format: int64
-  - score_name: Definition M (communities)
+  - score_name: Definition N (communities)
     label: Identified as disadvantaged
     format: bool
+  - score_name: Definition N (communities) (including adjacency index)
```
emma-nechamkin, line 26: The idea here is we now have TWO distinct mechanisms through which a tract can become a DAC. Either it meets the criteria explicitly in our methodology, OR it is surrounded by DACs.

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -207,7 +207,8 @@
     field_names.M_POLLUTION: "M_PLN",
     field_names.M_HEALTH: "M_HLTH",
     # temporarily update this so that it's the Narwhal score that gets visualized on the map
-    field_names.SCORE_N_COMMUNITIES: "SM_C",
+    # The NEW final score value INCLUDES the adjacency index.
+    field_names.SCORE_N_COMMUNITIES + field_names.ADJACENT_MEAN_SUFFIX: "SM_C",
```
emma-nechamkin, line 211: We want to visualize the "overall list" of DACs, including adjacency.

data/data-pipeline/data_pipeline/score/field_names.py
```diff
@@ -1,6 +1,8 @@
 # Suffixes
 PERCENTILE_FIELD_SUFFIX = " (percentile)"
 ISLAND_AREAS_PERCENTILE_ADJUSTMENT_FIELD = " for island areas"
+ADJACENT_MEAN_SUFFIX = " (including adjacency index)"
```
emma-nechamkin, line 4: I kinda hate these column names and stuff, but changed this to be a bit more copasetic with the repo's style

data/data-pipeline/data_pipeline/score/utils.py
```diff
@@ -0,0 +1,56 @@
+"""Utilities to help generate the score."""
+import pandas as pd
+import geopandas as gpd
+import data_pipeline.score.field_names as field_names
+
+# XXX: @jorge I am torn about the coupling that importing from
```
emma-nechamkin, line 6: Cannot speak to best practice here, but I think what you have here is good.

esfoobar-usds: @mattbowen-usds: I just realized the existence of `data_pipeline.etl.sources.geo_utils`. It looks like this was done in a prior PR. Was there a reason to put the functions there instead of adding these to `data_pipeline.utils`? Was there circular imports issues?

esfoobar-usds (approved):
I think this mostly looks good, but we should discuss about geo_utils when @mattbowen-usds is able.

🍏

data/data-pipeline/data_pipeline/score/utils.py
emma-nechamkin: I have referenced this here: https://github.com/usds/justice40-tool/issues/1839

PR #1829, opened 2022-08-17
Frontend release branch to main

vim-usds:
# Purpose
This PR will contain all FE changes for the 1.0 launch

# Features

## Indicators status
- [x] Future flood risk
- [x] Future wildfire risk
- [x] lack of green space
- [x] lack of plumbing
- [x] transportation barriers
- [x] leaky underground tanks
- [x] demographics
- [x] historic underinvestment (redlining)?
- [x] abandoned land mines
- [x] FUDS

## Demographics
- [x] Add demographics UI to sidepanel
- [x] Connect real BE data

## Geolocation
- [x] works on desktop
- [x] works on mobile
- [x] Fix UI on desktop and mobile

## Tribal layer selection
- [x] overlay in base map
- [x] remove all toggle logic and UI
- [x] connect BE signals for TA_COUNT and TA_PERC

## Side panel UI
- [x] remove all `up` and `down` 
- [x] missing data should have a _Value_ of missing data should be '--' with no _Icon_
- [x] when data is missing, the _SubText_ should say "missing data". Reduce size to fit in one line.
- [x] instead of highlighting the entire indicator(with a baby blue background), only highlight the _Value_ (dark blue background around the value) when the indicator is above the threshold. change the text color of the value to white.
- [x] instead of a blue dot on the category, change the entire background (to dark blue) of the category when it is above the associated threshold(s), otherwise leave as is. change the text color and the icon color (the plus sign) to white.
- [x] test on mobile (iOS and Android)
- [x] test on desktop (Chrome, Safari and Firefox)
- [x] review all ind titles and desc from content 1.0 doc
- [x] category title, indicator name, desc, value, subtext should all be grey-warm-90
- [x] _Title_ and _Value_ should have the same font treatment (weight and size as spacer)
- [x] remove initial spacer (at least one)
- [x] AND space remove bg color and ensure that vertical spacing b/w indicator is the same including spacers (check each category to make sure it applies equally)
- [x] all subtext in one line
- [x] on the value that has a background color, tighten up, equal spacing on all sides (2px left and right, 1px top and bottom)
- [x] Methodology version (pt, pb = 1rem)
- [x] update Methodology version to 1.0
- [x] update non-selected copy
- [x] updated 1.0 designs

## Copy
- [x] update 404 page
- [x] all copy updates from Content 1.0 doc on Explore the map page
- [x] copy bugs from bug party

## About page
- [x] Most of About page copy (some missing links that are TBD)
- [x] Navigation links for MethPage and AboutPage (desktop)
- [x] Navigation links for MethPage and AboutPage (mobile) 

## Bugs
- [x] removes duplicate higher ed
- [x] all links should be the same color
- [x] Logo title shouldn't wrap


emma-nechamkin:
@vim-usds this is *awesome*! Historic Underinvestment should already be in the backend - just want to confirm it's making its way to you (`HRS_ET`)? I also noticed there's nothing here for the adjacency index (donut holes). We should make sure this update includes the adjacency index. Is there anything you need from me for this?

vim-usds:
@emma-nechamkin and @esfoobar-usds 

Thanks for flagging! So after digging into this, we downloaded the `usa-high.json` file; used geopandas to get all 124 columns (aka BE signals) and compared that to the BE signals coming into a randomly selected tile on the FE.

This randomly selected tile was missing the following columns (or BE signals):

- HRS_ET
- IAHSEF
- IALMILHSE_PFS
- IAPLHSE_PFS
- IAULHSE_PFS

Given that you were saying they only occur in small number of tracts, I wonder if for the tracts they are not apart of, they are being removed completely rather than having a null and/or 0 value? I think this gets at your correctness effort.

One test would be - How many BE signals should be in every tile? I imagine this number should be the same for every time right? If there is no data, it should have a value of null? We can then check this on the FE and flag errors to the console on tiles that are not meeting the total number of BE signals.

Just my .02 cents!

mattbowen-usds (approved):
Sadly, there's no way I can review all of this, but we need to go live and main has protection, so :+1:

PR #1832, opened 2022-08-17
Adding eamlis and fuds data to legacy pollution in score

emma-nechamkin:
This will require further QA and correctness tests -- to come!


esfoobar-usds (approved):
🍏

PR #1842, opened 2022-08-19
Updating flags for the front end

emma-nechamkin:
Update tile flags as follows:

- [x] add a flag for whether income has been imputed and adjusted (T/F)
- [ ] add a flag for whether the tract is a donut 
- [ ] update Score N to be donut union Score N


emma-nechamkin:
weird that this commit was so weird?

Issue #1846, opened 2022-08-19
Back end score tests


emma-nechamkin:
Create tests for score generation that encompass:

- [x] All tracts for every indicator with the ET flag that meet socio criterion are DACs (sans adjacency)
- [x] ET flag is properly computed, with a weakly greater than sign
- [x] Socioeconomic flag is properly computed, with a weakly greater than sign 
- [x] Non-percentile-based indicators are properly computed (e.g., HRS)

These tests should be similar to what is below:
```
assert tmp[tmp["UST_ET"] & tmp["FPL200S"]]["SM_NO_DON"].min(), "Failed!"
assert (tmp[tmp["UST_ET"]]["UST_PFS"].min() >= 0.9) & (
    tmp[~tmp["UST_ET"]]["UST_PFS"].max() < 0.9
), "Failed!"
```

Issue #1840, opened 2022-08-18
Produce FIPS list for each dataset


emma-nechamkin:
# Overview

The zip file contains all of the data for which tracts we push from end-to-end on our ETL (so only what makes it "out" alive). We should verify that this is all of the data we could conceivably have manually and by data source.

What does it mean to verify a data source? Here, our goal is to make sure that the data we see in the backend is comprehensive relative to what data we "should" have (see the Wisconsin error). We are not looking in code nor are we trying to "fix" anything -- this is *purely* a diagnostic exercise. 

## Files to look at
This issue has two comments below, each of which contain spreadsheets with information. In the ZIP file, there are Excel sheets per data source. 

### Each data source excel sheet has five columns:
- `fips_code`: two digit census code
- `<DATASETNAME>__count_tracts`: how many census tracts the dataset has within the fips code
- `state_name`: name of the state represented by the fips code
- `state_abbreviation`: abbreviation of the state represented by the fips code
- `region`: the region of the state (census defines this)
- `division`: the division of the state (census defines this)

Example of how to read it: in the sheet `hud_housing__fips_check.xlsx`, the first row tells me that the hud housing datasource includes 1181 tracts in Alabama. 

### There is also the `whole_list.xlsx` file, which allows us to compare across data sources. This file has a slightly more confusing structure. 
- each column has the count of tracts within a FIPS code
- each row represents a data source's results

Example of how to read it: the first row tells me that `census_acs_median_income_2019__count_tracts` has 1178 tracts in fips code 01. This means that the data source `census acs median income` includes 1178 tracts in fips code 01. 

## What to do per data source

- [ ] *Check that the data that we have matches the source data in terms of coverage**: Start in the `DATASOURCE__fips_check.csv` file. For every FIPS code with <10 tracts represented, verify that those states are not represented in the source data. The source data is either on S3 or publicly-available. (This checks for errors in the original data we have, a la the WI error)
- [ ] *Check that number of tracts included by FIPS code makes sense*: Open the `whole_list.xlsx` file. The ground truth for all but Island Areas is the `census_acs_2010__count_tracts`. The ground truth for the Island Areas is `census_dicennial_2010__count_tracts`. Compare the **number of tracts** per FIPS code and note down any anomalies. 

### DATASOURCE NAME
__High-level checks__
- [ ] Our data matches the source data in terms of state-level coverage
- [ ] We have somewhat complete data for all 50 states 
- [ ] We have somewhat complete data for Puerto Rico
[...]

PR #1847, opened 2022-08-19
Score tests

emma-nechamkin:
#1846 
The goal of these tests are to ensure that the specific score calculations are correct. We use the *true* score output here and validate to make sure no calculation was obviously wrong. I think this is what our dear @mattbowen-usds would call "smoke tests".  

__Some outstanding tests__

- [x] Test that every percentile >= 0.9 is included, EXCEPT for the special case of lead paint
   - I've included a template for these tests. They are verbose! Would love feedback before blowing them out across categories. 
- [x] Test the specific construction of the lead paint indicator
-~~[ ] Test that HRS uses the right >= 3.25 threshold~~
   - This should actually be a *dataset-level* test since it's a research-derived threshold (@mattbowen-usds)
- [x] Test that low income thresholds (donut and regular) use >= thresholds
   - I've done this for "regular" but would like some feedback before blowing it out
- [x] Test that donut indicator uses the right >= 1.0 threshold
- [x] Test that low HS uses the right threshold
~~- [ ] Tests for the Island Areas~~
    - This is actually a pretty big load, and so I'm going to stick it into a second ticket and count this ticket roughly complete
- [x] Test that (for the percentile field), at least 1000 tracts are returned
- [x] Test that under 40% (strictly speaking) of the population is included in DACs
    - note: this is not a correctness test but it kinda is?
- [x] Test that donuts are added to the score
- [x] Test that DACs are identified 

__Lower priority__
- [ ] Test that the eligibility counter is incremented properly 
- [ ] Test that the categories are summed right


data/data-pipeline/data_pipeline/tests/test_score_output.py
```diff
@@ -0,0 +1,252 @@
...
+import pytest
+import data_pipeline.score.field_names as field_names
+from data_pipeline.utils import get_module_logger
+from data_pipeline.config import settings
+from data_pipeline.score.score_narwhal import ScoreNarwhal
+
+logger = get_module_logger(__name__)
+
+
+# this is so slow?
```
emma-nechamkin, line 12: This specific fixture takes a long time to load... any improvements?

data/data-pipeline/data_pipeline/tests/test_score_output.py
```diff
@@ -0,0 +1,252 @@
...
+@pytest.fixture
+def final_score_df():
+    return pd.read_csv(
+        settings.APP_ROOT / "data" / "score" / "csv" / "full" / "usa.csv",
+        dtype={field_names.GEOID_TRACT_FIELD: str},
+    )
+
+
+# when should things be fixtures??
```
emma-nechamkin, line 21: General comment -- when should things be fixtures vs. constants? I've erred on the side of fixtures because that is how it looks in our codebase... but interested in comments and thoughts

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
```diff
@@ -0,0 +1,186 @@
...
+@pytest.fixture
+def expected_wildfire_percentile_columns():
+    return [
+        field_names.HIGH_FUTURE_WILDFIRE_RISK_FIELD,
+        field_names.FUTURE_WILDFIRE_RISK_FIELD
+        + field_names.PERCENTILE_FIELD_SUFFIX,
+    ]
+
+
+def test_expected_pop_loss_percentile_columns(
```
emma-nechamkin, line 95: this is really verbose, but the other option makes it much less modular. Thoughts?

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
```diff
@@ -0,0 +1,186 @@
...
+import data_pipeline.score.field_names as field_names
+from ..score.score import Score
+from data_pipeline.utils import get_module_logger, score_folder_cleanup
+from data_pipeline.config import settings
+from data_pipeline.score.score_narwhal import ScoreNarwhal
+
+logger = get_module_logger(__name__)
+
+### TODO: we need to blow this out for all eight categories
+def _helper_test_percentile_against_threshold(
```
emma-nechamkin, line 14: this verbosely checks that all percentile comparisons are weakly greater.

lucasmbrown-usds:
This looks like a good set of tests! There are also some things we can think of more as "smell tests" (like `Test that under 40% (strictly speaking) of the population is included in DACs`). These could be run not as tests, but as a step in the creation of a new score, for instance at the end of the `score_n.py` process before we return `Score N` to the scorerunner. If they fail, you can raise a ValueError with a message. 

The smell tests could be like -- there are no null values in a category for rows where we have data in all of the rows, etc. Or that if you `group_by` state/territory, we see data as expected for all the states and territories.

emma-nechamkin:
> This looks like a good set of tests! There are also some things we can think of more as "smell tests" (like `Test that under 40% (strictly speaking) of the population is included in DACs`). These could be run not as tests, but as a step in the creation of a new score, for instance at the end of the `score_n.py` process before we return `Score N` to the scorerunner. If they fail, you can raise a ValueError with a message.
> 
> The smell tests could be like -- there are no null values in a category for rows where we have data in all of the rows, etc. Or that if you `group_by` state/territory, we see data as expected for all the states and territories.

TY @lucasmbrown-usds  but also stop working. 

Yup -- a lot of those smell tests are included in these tickets: 
- https://github.com/usds/justice40-tool/issues/1848
- https://github.com/usds/justice40-tool/issues/1849
- https://github.com/usds/justice40-tool/issues/1840
- TODO: https://github.com/usds/justice40-tool/issues/1850

data/data-pipeline/data_pipeline/tests/test_score_output.py
mattbowen-usds: Once it's loaded once, will it change? If not, we could `lru_cache` it to at least only ever have to load it once per thread. The other option would be to write a faster to load file format, like feather --- but that has problems of its own, since it could mask problems with the CSV or create new ones.

data/data-pipeline/data_pipeline/tests/test_score_output.py
mattbowen-usds: @esfoobar-usds I personally wouldn't make most of these fixtures --- I use fixtures for things I'm calculating (data or complicated mocks) that I want to reset with each test invocation, and I would put the rest just as constants I import to reduce the mental overhead with figuring out where a given name comes from (since fixtures have a little bit of magic to them in how they use DI). What's your-all's overall style though?

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
```diff
@@ -0,0 +1,286 @@
...
+from data_pipeline.utils import get_module_logger
+from data_pipeline.config import settings
+from data_pipeline.score.score_narwhal import ScoreNarwhal
+
+logger = get_module_logger(__name__)
+
+
+@pytest.fixture
+def donut_hole_adjacency_parameters():
+    """Test that donut indicator uses the right threshold"""
```
mattbowen-usds, line 16: I don't really understand this docstring --- this is not a test?

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
```diff
@@ -0,0 +1,286 @@
...
+    """Test that donut income uses the right threshold"""
+    return (
+        # threshold
+        ScoreNarwhal(final_score_df).LOW_INCOME_THRESHOLD_DONUT,
+        field_names.FPL_200_SERIES_IMPUTED_AND_ADJUSTED_DONUTS,
+    )
+
+
+@pytest.fixture
+def final_score_df():
```
mattbowen-usds, line 38: I'm guessing this is slow too, and I think is duplicate below. This is a nit but we can probably factor some of these into a common fixtures.py

emma-nechamkin:
In another round of testing, we need to create these tests for the Islands

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
mattbowen-usds: I agree, this is no fun. Given the structural similarity and the broad scope of these, I'd be pretty okay with a single "test_percentile_columns" that just iterates over these columns, given the clarity of the assertions. I can draft up a version after lunch for folks to think over.

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
```diff
@@ -0,0 +1,286 @@
...
+    )
+
+
+def test_donut_hole_thresholds(
+    final_score_df,
+    donut_hole_adjacency_parameters,
+    donut_hole_income_threshold_and_column,
+    low_income_percentile,
+):
+    """We do it all! Check donut income and adjacency thresholds, and then that
```
mattbowen-usds, line 266: This is my favorite doc string in the suite

mattbowen-usds (commented):
Initial feedback, more to come

data/data-pipeline/data_pipeline/tests/test_score_output.py
emma-nechamkin: It does only get loaded a single time, so I think `lru_cache` will work. Do you happen to have an example of this handy? (if not, I am happy to google)

data/data-pipeline/data_pipeline/tests/test_score_output.py
emma-nechamkin: This is SUPER helpful, @mattbowen-usds. TBH I have no idea on our structure, which is why (again, I am not an engineer!) I don't think I am the best choice for designing all of these tests. @esfoobar-usds could you please provide guidance?

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
emma-nechamkin: That sounds really great to me!

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
emma-nechamkin: Oh yikes, sorry -- the string should say "parameters to test"...

data/data-pipeline/data_pipeline/tests/test_score_calculation.py
emma-nechamkin: yeah that sounds good to me!

data/data-pipeline/data_pipeline/tests/test_score_output.py
mattbowen-usds: I have a great example! See https://github.com/usds/justice40-tool/blob/emma-nechamkin/release/score-narwhal/data/data-pipeline/data_pipeline/etl/sources/geo_utils.py#L13 --- for a use like this, you can just import it from functools and then decorate the fixture and it should work. First load will be slow, but subsequent loads should come out of memory, unless pytest fixtures do dark magic.

data/data-pipeline/data_pipeline/tests/score/test_calculation.py
```diff
@@ -102,6 +101,112 @@ def test_percentile_columns(final_score_df):
...
+    )
+    # Leadpaint is handled below in a separate method
+    housing = PercentileTestConfig(
+        field_names.HOUSING_BURDEN_FIELD,
+        field_names.HOUSING_BURDEN_PCTILE_THRESHOLD,
+        ScoreNarwhal.ENVIRONMENTAL_BURDEN_THRESHOLD
+    )
+    non_natural_space = PercentileTestConfig(
+        field_names.TRACT_PERCENT_NON_NATURAL_FIELD_NAME,
+        field_names.NON_NATURAL_PCTILE_THRESHOLD,
```
emma-nechamkin, line 142: these field names in particular make me so mad at myself lol

data/data-pipeline/data_pipeline/tests/score/test_calculation.py
```diff
@@ -113,6 +218,27 @@ def test_percentile_columns(final_score_df):
...
+        wastewater,
+        ust,
+        diabetes,
+        asthma,
+        heart_disease,
+        low_life_expectancy,
+        unemployment,
+        low_median_income,
+        linguist_isolation,
+        poverty,
```
emma-nechamkin, line 241: and then, just for my own accounting, we have the three boolean factors (aml, hrs, and fuds) and the composite fancy lead indicator. this is great!

emma-nechamkin (commented):
this is awesome imo!

mattbowen-usds:
I just need to fixup the build and whatever lint errors I missed. Should be able to do that first thing tomorrow.

data/data-pipeline/data_pipeline/tests/score/test_calculation.py
```diff
@@ -0,0 +1,288 @@
...
+        )
+    if not is_maximum_not_flagged_ok:
+        errors.append(
+            f"For column {config.threshold_column_name}, there is someone not flagged above {config.threshold} percentile!"
+        )
+    return errors
+
+
+def test_percentile_columns(final_score_df):
+    low_income = PercentileTestConfig(
```
mattbowen-usds, line 57: So having done this, I started wanting to make it fancier --- like make the list happen automatically via some metaprogramming, and be able to impute the names of fields that follow a convention. But also I wanted to keep the test as obvious as possible, so I stuck with what I had. Honestly writing the test reinforced that the actual score code itself could use some refactoring to reduce duplication there, and then the test could rely on stronger conventions within tho score code to be further simplified.

data/data-pipeline/data_pipeline/config.py
```diff
@@ -12,7 +12,7 @@
 
 # set root dir
 settings.APP_ROOT = pathlib.Path(data_pipeline.__file__).resolve().parent
-
+settings.REQUESTS_DEFAULT_TIMOUT = 3600
```
mattbowen-usds, line 15: Since before we had no timeouts, which means wait forever, I picked a very generous hour. This was arbitrary tho --- of course open to changing it.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -496,10 +496,11 @@ def _prepare_initial_df(self) -> pd.DataFrame:
         # >= some threshold.
         # TODO: Add more fields here.
         #  https://github.com/usds/justice40-tool/issues/970
-        ReversePercentile = namedtuple(
-            typename="ReversePercentile",
-            field_names=["field_name", "low_field_name"],
-        )
+        @dataclass
```
mattbowen-usds, line 499: I did this so the newer version of pylint can parse this file on python 3.8. I know @esfoobar-usds reverted pylint, but I'm leaving it so when we do upgrade, everything is copacetic.

data/data-pipeline/data_pipeline/etl/score/etl_score_geo.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, data_source: str = None):
 
         ## TODO: We really should not have this any longer changing
         self.TARGET_SCORE_SHORT_FIELD = constants.TILES_SCORE_COLUMNS[
-            field_names.SCORE_N
+            field_names.FINAL_SCORE_N_BOOLEAN
```
mattbowen-usds, line 54: @emma-nechamkin is this the right change lol?

mattbowen-usds:
We are explicitly punting the last two todos:

- Test that the eligibility counter is incremented properly
- Test that the categories are summed right

mattbowen-usds (approved):
Given that I wrote some of this, I clearly approve of it, but to be explicit.

.github/workflows/deploy_be_staging.yml
```diff
@@ -62,6 +62,9 @@ jobs:
       - name: Generate Score Post
         run: |
           poetry run python3 data_pipeline/application.py generate-score-post -s aws
+      - name: Run Smoketests
+        run: |
+          poetry run pytest data_pipeline/ -m smoketest
```
esfoobar-usds, line 67: Nice! 🎉

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -1,5 +1,5 @@
 import functools
-from collections import namedtuple
+from dataclasses import dataclass
```
esfoobar-usds, line 2: YAY DATACLASS, DOWN WITH NAMEDTUPLE /cc @lucasmbrown-usds

data/data-pipeline/data_pipeline/etl/score/etl_score.py
esfoobar-usds: Thank you! I will try another time with a branch off of Narwahl to see if the pylint errors resolved after you merge this

data/data-pipeline/data_pipeline/tests/score/fixtures.py
```diff
@@ -0,0 +1,12 @@
+import pandas as pd
+import pytest
+from data_pipeline.config import settings
+from data_pipeline.score import field_names
+
+
+@pytest.fixture(scope="session")
```
esfoobar-usds, line 7: SCOPE!!

esfoobar-usds (approved):
🍏

emma-nechamkin:
this is great! ty for the catch!

PR #1852, opened 2022-08-22
Update airport color on mapbox base map

vim-usds:


vim-usds:
@KameronKerger - looks like you fixed it! Awesome! If we are satisfied with the map, can I push this to production ?

Issue #1836, opened 2022-08-17
Correctness tests for score calculation. 


emma-nechamkin:
- [ ] There should be NO tracts for which 90th percentile is exceeded for a given indicator AND the tract meets the socioeconomic threshold, and the tract is NOT a DAC. We should check this for every single indicator -- this will ensure we've included them all and included them correctly
- [ ] For each indicator, we should print out a list of FIPS codes (crosswalked to States) to have a go-to list of which places should have each variable. Then, we should confirm that every state that does not have the data is nulled out. (This will also have manual review)
- [ ] Must add a unit test for "fancy" percentile calculations
- [ ] Manually check "outliers" -- tracts that have low building value but a high percentile, for example (others include low acreage but high non-greenspace, low property count but high fire / flood percentiles, etc). This makes sure we are catching true trends and not noise (we can always clip data). 
- [ ] Check manually income imputation, and then see how many "imputed income" tracts are low income. Check that these make sense. 
- [ ] HOLC data scalar comparison – here we want to ensure that we have two fields, a field from the raw data that has T/F and only the tracts for which redlining exists, and a field for the usa.csv that has F filled in for NaN, I think 

We might have to break this into two tickets, but this is (right now) a list of data-intensive tests we can do.

emma-nechamkin:
i've cleaned this up

PR #1864, opened 2022-08-24
Pipeline tile tests

emma-nechamkin:
__Tiles and data pipeline:__ The goal here is to smoke test correctness
- [ ] We should check for wrong variable ported over to the tiles by making sure the column name for each thing we want in `usa.csv` has the same value as the column name for the tiles csv. This is NOT the same as checking that the column has been renamed. Rather, this requires looking at the methodology of the tool, seeing what should be represented, and ensuring that variable is indeed represented. 
**This is something that we have to do by hand, I suspect -- it's a LOT of hardcoding for limited payoff, and needs to be verified in QA**
- [x] Check for USVI and Guam in tiles
~~- [ ] We should check that even census tracts that have NULLs for every indicator get included in the tiles and in the csv and are never dropped~~
**it is my understanding that this is a tippecanoe problem**


emma-nechamkin:
Check for USVI and Guam in tiles --> here i actually just check that there are all the states we might expect by number, in the tiles AND not as nulls.

data/data-pipeline/data_pipeline/tests/test_tiles_smoketests.py
```diff
@@ -0,0 +1,96 @@
...
+    ), "Some states are missing from tiles"
+    pfs_columns = tiles_df.filter(like="PFS").columns.to_list()
+    assert (
+        tiles_df.dropna(how="all", subset=pfs_columns)["GTF"].str[:2].nunique()
+        == 56
+    ), "Some states do not have any percentile data"
+
+
+# For each data point that we visualize, we want to confirm that
+# (1) the column is represented in tiles_columns
```
emma-nechamkin, line 95: See the percentile fields above for an example of this. 

Note the the islands are commented out here. We use more complicated logic for the Island areas (we use the full 2010 data and then null out non-Islands). We need some way to write tests that this combination is correct. 

I think for now, though, since we've tested in #1836 all of the calculations, we can be satisfied with the rename. We do have to check for type, though. Booleans for tiles should always be T/F, for example -- we've definitely sent Vim 1,  0 in the past.

emma-nechamkin (commented):
I agree that this is the simplest possible change and I think it's nice enough. If we ever do refactor (seems unlikely tbh) we could constrain the tract list by source -- for now I think this is great.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -384,6 +394,21 @@ def _prepare_initial_df(self) -> pd.DataFrame:
...
+        # Drop tracts that don't exist in the 2010 tracts
+        tracts_to_drop_count = len(
+            set(census_tract_df.GEOID10_TRACT)
+            - set(self.national_tract_df.GEOID10_TRACT)
+
+        logger.info(
+            "Dropping %s tracts not in the 2010 tract data",
+            tracts_to_drop_count,
+        )
+        census_tract_df = census_tract_df.loc[
```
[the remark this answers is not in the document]
mattbowen-usds: Based on where we join that dataset in elsewhere, I think there may be tracts in there that we do not have data for, so I didn't want to accidentally introduce them. But also maybe I'm thinking something goofy.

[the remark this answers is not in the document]
emma-nechamkin: inner join?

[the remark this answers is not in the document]
mattbowen-usds: yah, inner with the merged df on the left. probably faster than my subquery version. will switch.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -384,7 +394,19 @@ def _prepare_initial_df(self) -> pd.DataFrame:
...
+        # Drop tracts that don't exist in the 2010 tracts
+        pre_join_len = census_tract_df.shape[0] 
+
+        census_tract_df = census_tract_df.merge(
+            self.national_tract_df,
+            on="GEOID10_TRACT",
+            how="inner",
+        )
+
+        logger.info(
```
emma-nechamkin, line 406: This is me being CRAZY but I would assert the size is <= to make sure the join performed as expected -- it obviously should but 🤷

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -45,6 +45,7 @@ def __init__(self):
         self.persistent_poverty_df: pd.DataFrame
         self.census_decennial_df: pd.DataFrame
         self.census_2010_df: pd.DataFrame
+        self.national_tract_df: pd.DataFrame
```
emma-nechamkin, line 48: We might want to drop the fields that aren't used in the score from this, just so we don't have to continue the support. persistent poverty, child opportunity index i think can be removed. wdyt?

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -384,7 +394,19 @@ def _prepare_initial_df(self) -> pd.DataFrame:
...
+            how="inner",
+        )
+
+        logger.info(
+            "Dropped %s tracts not in the 2010 tract data",
+            pre_join_len - census_tract_df.shape[0],
+        )        # If GEOID10s are read as numbers instead of strings, the initial 0 is dropped,
         # and then we get too many CBG rows (one for 012345 and one for 12345).
 
         # Now sanity-check the merged df.
```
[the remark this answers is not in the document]
emma-nechamkin: I also want us to get rid of some of the arithmetic below for like, variables we calculate but do not use in the score. WDYT?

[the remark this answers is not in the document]
emma-nechamkin: I'm gonna add it to clean up

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -217,28 +219,8 @@ def _create_score_data(
             right_on=self.STATE_CODE_COLUMN,
             how="left",
         )
-
```
mattbowen-usds, line 220: Now that we truncate the tiles in the main score to the 2010 tract list, we don't need to do any of this here, which saves us several steps.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -217,28 +219,8 @@ def _create_score_data(
...
-        # merge census tracts with score
-        merged_df = national_tract_df.merge(
-            score_county_state_merged,
-            on=self.GEOID_TRACT_FIELD_NAME,
-            how="left",
-        )
-
-        # recast population to integer
-        score_county_state_merged["Total population"] = (
```
mattbowen-usds, line 232: I'm pretty sure this line was a no-op, given `score_county_state_merged` never got used in the return value --- and since we join to national_tract_df, we don't convert any other ints to float64 because of nans now.

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -0,0 +1,210 @@
...
+        self._is_dtype_ok = (
+            self.final_score_column.dtype == self.tiles_column.dtype
+        )
+        self._is_value_ok = False
+        if self._is_dtype_ok:
+            if self.final_score_column.dtype == np.dtype("float64"):
+                self._is_value_ok = np.allclose(
+                    self.final_score_column,
+                    self.tiles_column,
+                    atol=float(f"1e-{constants.TILES_ROUND_NUM_DECIMALS}"),
```
mattbowen-usds, line 121: if this weren't a test, i would feel bad about this line, but instead i do not feel anything.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
mattbowen-usds: that is a good call and i added it!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
mattbowen-usds: Want me to do that here, or in the cleanup ticket? I kinda just want to keep focusing on correctness tests and then move to cleanup when I have a nice test suite under me, but I don't feel strongly.

mattbowen-usds (commented):
Added comments on my own code for future reviewers

data/data-pipeline/data_pipeline/etl/score/etl_score.py
emma-nechamkin: I put this into the cleaning ticket. I definitely think "get it right, get it tight" is the right approach here!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -384,8 +394,21 @@ def _prepare_initial_df(self) -> pd.DataFrame:
 
         census_tract_df = self._join_tract_dfs(census_tract_dfs)
 
-        # If GEOID10s are read as numbers instead of strings, the initial 0 is dropped,
-        # and then we get too many CBG rows (one for 012345 and one for 12345).
+        # Drop tracts that don't exist in the 2010 tracts
+        pre_join_len = census_tract_df.shape[0]
```
emma-nechamkin, line 398: an additional psychotic check -- I'm wondering if we should check for unique census tracts instead of just shape (e.g., `census_tract_df[GEOID10_TRACT].nunique()`)?

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
emma-nechamkin: this is lovely

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -217,28 +219,8 @@ def _create_score_data(
...
-            merged_df["Total population"].fillna(0).astype(int)
-        )
-
-        de_duplicated_df = merged_df.dropna(
-            subset=[DISADVANTAGED_COMMUNITIES_FIELD]
-        )
-
         # set the score to the new df
-        return de_duplicated_df
+        return score_county_state_merged
```
emma-nechamkin, line 223: we should confirm that there's no duplication, either here or on merge

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
emma-nechamkin: we also shouldn't be filling NaN with 0!

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -0,0 +1,195 @@
...
+    pfs_columns = tiles_df.filter(like="PFS").columns.to_list()
+    assert (
+        tiles_df.dropna(how="all", subset=pfs_columns)["GTF"].str[:2].nunique()
+        == 56
+    ), "Some states do not have any percentile data"
+
+
+def test_column_presence(tiles_df):
+    expected_column_names = set(TILES_SCORE_COLUMNS.values()) | {
+        THRESHOLD_COUNT_TO_SHOW_FIELD_NAME,
```
emma-nechamkin, line 87: i think we are going to cut these two fields out (at least UI Experience, for sure)

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -0,0 +1,195 @@
...
+    assert not (
+        extra_columns
+    ), f"tiles/usa.csv has columns not specified in TILE_SCORE_COLUMNS: {extra_columns}"
+    assert not (
+        missing_columns
+    ), f"tiles/usa.csv is missing columns from TILE_SCORE_COLUMNS: {missing_columns}"
+
+
+def test_tract_equality(tiles_df, final_score_df):
+    assert tiles_df.shape[0] == final_score_df.shape[0]
```
emma-nechamkin, line 102: i love this

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
emma-nechamkin: 😂

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -0,0 +1,195 @@
...
+            "CF",  # County field, added at geoscore,
+            constants.THRESHOLD_COUNT_TO_SHOW_FIELD_NAME,
+            constants.USER_INTERFACE_EXPERIENCE_FIELD_NAME,
+        ]
+    )
+    errors = []
+
+    # Are the dataframes the same shape truly
+    assert tiles_df.shape == final_score_df.shape
+    assert (tiles_df.GTF == final_score_df.GTF).all()
```
emma-nechamkin, line 177: this is all very clever. 

One operation that could come in handy here is `equals` which tests equality and has no problem with  `NaNs`

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -0,0 +1,195 @@
...
+    assert sorted(tiles_df.columns) == sorted(final_score_df.columns)
+
+    # Are all the dtypes and values the same?
+    comparisons = []
+    for col_name in final_score_df.columns:
+        value_comparison = ColumnValueComparison(
```
emma-nechamkin, line 183: this is very nice

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -0,0 +1,195 @@
...
+            final_score_df[col_name], tiles_df[col_name], col_name
+        )
+        comparisons.append(value_comparison)
+    errors = [comp for comp in comparisons if not comp]
+    error_message = "\n".join(error.error_message for error in errors)
+    assert not errors, error_message
+
+
+def test_for_state_names(tiles_df):
+    states = tiles_df.SF.value_counts(dropna=False).index
```
emma-nechamkin, line 193: one general note -- I think we want to access columns not by `df.name` but rather as `df[name]` simply because we might refactor and change the shortnames to match the yaml.

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -0,0 +1,195 @@
...
+    final_score_df = final_score_df.rename(columns=TILES_SCORE_COLUMNS).drop(
+        final_score_df.columns.difference(TILES_SCORE_COLUMNS.values()),
+        axis=1,
+        errors="ignore",
+    )
+
+    # Drop the UI-specific fields from the tiles dataframe
+    tiles_df = tiles_df.drop(
+        columns=[
+            "SF",  # State field, added at geoscore
```
emma-nechamkin, line 167: one general note -- I think we want to access columns not by `df.name` but rather as `df[name]` simply because we might refactor and change the shortnames to match the yaml.

emma-nechamkin (commented):
I am checking this out and going to run it / check the output -- but wanted to submit review comments first :)

emma-nechamkin:
I checked this quickly locally and it looks 👍

data/data-pipeline/data_pipeline/etl/score/etl_score.py
mattbowen-usds: Good thinking --- that is more precise and conveys better what we're trying to do. Will implement.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
mattbowen-usds: Great call, added.

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
mattbowen-usds: Done below!

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
mattbowen-usds: Cool I fixed those up. Thank you.

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -108,13 +108,43 @@ class ColumnValueComparison:
...
+            fake_bool = {1.0, 0.0, None}
+            # Replace the nans in the column values with None for
+            # so we can just use issubset below
+            col_values = set(
+                not np.isnan(val) and val or None
+                for val in self.tiles_column.value_counts(dropna=False).index
+            )
+            return (
+                len(col_values) <= 3
+                and col_values.issubset(fake_bool)
```
emma-nechamkin: nice

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -108,13 +108,43 @@ class ColumnValueComparison:
...
+            )
+        return False
+
+    @property
+    def _is_dtype_ok(self) -> bool:
+        if self.final_score_column.dtype == self.tiles_column.dtype:
+            return True
+        if (
```
emma-nechamkin, line 128: got it. so here and only here we handle this case. this looks great

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -108,13 +108,43 @@ class ColumnValueComparison:
...
     def __post_init__(self):
-        self._is_dtype_ok = (
-            self.final_score_column.dtype == self.tiles_column.dtype
-        )
         self._is_value_ok = False
         if self._is_dtype_ok:
-            if self.final_score_column.dtype == np.dtype("float64"):
+            if self._is_tiles_column_fake_bool:
+                # Cast to actual bool for useful comparison
+                self.tiles_column = self.tiles_column.apply(
```
emma-nechamkin, line 141: nice catch on np.nan as true!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
esfoobar-usds: I removed it in my tribal PR, so that should be taken care once we merge to Narwahl Release

data/data-pipeline/data_pipeline/tests/score/test_tiles_smoketests.py
```diff
@@ -0,0 +1,221 @@
...
+    "LPF_PFS",
+    "KP_PFS",
+    "NPL_PFS",
+    "RMP_PFS",
+    "TSDF_PFS",
+    "TF_PFS",
+    "UF_PFS",
+    "WF_PFS",
+    "UST_PFS",
+]
```
esfoobar-usds, line 56: Argh -- these types of lists will/should come from the YAML

esfoobar-usds (approved):
Good stuff! 🍏

PR #1871, opened 2022-08-25
Adding tests to ensure proper calculations

emma-nechamkin:
The following tests can be unit tests to check the accuracy of each calculation:
- [x] Ag value percentile is calculated correctly (floored properly)
- [x] Ling iso is not included for PR in the percentile
~~- [ ] Imputation of income (see code)~~ This is a real yikes because I was lazy when I initially wrote the code. I don't have the energy to like, FIX it right now 

We also have some analysis-type tests we should run as more of "smell tests"
- [ ] Distributions of each raw variable (this is hard because it depends on the variables -- anything with denominators we have to be really careful about)
- [x] Counts of tracts exceeding 90th percentile (this is easy)


data/data-pipeline/data_pipeline/tests/score/test_score_narwhal_methods.py
```diff
@@ -0,0 +1,84 @@
...
+        ],
+        on=[field_names.GEOID_TRACT_FIELD],
+    )
+
+    return check_frame["true_rank"].equals(
+        check_frame["to_rank_auto" + field_names.PERCENTILE_FIELD_SUFFIX]
+    )
+
+
+def test_drop_0_tracts(toy_score_df):
```
emma-nechamkin, line 55: @mattbowen-usds you do these nice configs for this stuff, but I'm wondering if that's necessary since this is so... trivial. Wdyt?

data/data-pipeline/data_pipeline/tests/score/test_score_narwhal_methods.py
```diff
@@ -0,0 +1,84 @@
...
+    assert _helper_test_dropping_tracts(
+        toy_score_df, drop_tracts=[]
+    ), "Percentile in score fails when we do not drop any tracts"
```
emma-nechamkin, line 58: @mattbowen-usds you did nice lists of errors that got printed for the different configs. I can see how to emulate it here, but I'm wondering if you think I should mirror that structure?

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
```diff
@@ -153,6 +153,19 @@ def transform(self) -> None:
             lower=self.AGRIVALUE_LOWER_BOUND
         )
 
+        ## Check that this clip worked -- that the only place the value has changed is when the clip took effect
```
emma-nechamkin, line 156: This isn't the nicest -- but we don't port out of the ETL a lot of these columns, so I check in place. Is it too sloppy?

data/data-pipeline/data_pipeline/tests/score/test_score_narwhal_methods.py
mattbowen-usds: For tests like this I wouldn't --- i made those configs for places I knew I'd be repeating the same structure over and over. I honestly think `_helper_test_dropping_tracts` is like exactly the right abstraction here and these are good the way they are.

data/data-pipeline/data_pipeline/tests/score/test_score_narwhal_methods.py
mattbowen-usds: As above, I think these are good the way they are. Like in general, this is the better pattern to follow than what I did in those output and calculation tests --- in those, because there was so much inherent repetition and because of the place in the pipeline we were, I wanted to try to make extending the tests as painless as I could. These are much closer to true unit tests, and so having them be simple like this with a custom assertion the way you've done is better than what I did in those other examples. :1st_place_medal:

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
mattbowen-usds: Honestly the only thing I'd change is the error message in the assertion at 167 --- like, can you be any more descriptive about how the clip didn't work? Otherwise, this fits the pattern we discussed pretty well of adding assertions within the code to programmatically enforce our assumptions.

mattbowen-usds (approved):
I have the tiniest request that i would not block merging on

data/data-pipeline/data_pipeline/tests/score/test_score_narwhal_methods.py
emma-nechamkin: thank you!

data/data-pipeline/data_pipeline/tests/score/test_score_narwhal_methods.py
emma-nechamkin: great! thank you for the gut check!

data/data-pipeline/data_pipeline/etl/sources/national_risk_index/etl.py
emma-nechamkin: 👍 thank you so much! updated

PR #1887, opened 2022-09-07
Issue 1831: missing life expectancy data from Maine and Wisconsin

### LINKED ISSUE #1831: Missing life expectancy data for WI and potentially another state
**Describe the bug**
We are only ingesting 49 of 51 available states from https://www.cdc.gov/nchs/nvss/usaleep/usaleep.html

**Expected behavior**
We should be including all available states

NOTE: it appears Maine is also missing.

lucasmbrown-usds:
Fixes #1831. 

It turns out the issue is in the source data. I fixed it and added some validations to ensure that if they change their data we'll be notified.


data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -399,5 +399,5 @@
     # that use null to signify missing information in a boolean field.
     field_names.ELIGIBLE_FUDS_BINARY_FIELD_NAME,
     field_names.AML_BOOLEAN,
-    field_names.HISTORIC_REDLINING_SCORE_EXCEEDED
+    field_names.HISTORIC_REDLINING_SCORE_EXCEEDED,
```
lucasmbrown-usds, line 402: I ran black on the whole `.py` codebase.

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -23,32 +35,122 @@ def __init__(self):
...
+            low_memory=False,
+        )
+
+        # Check which states are missing
+        states_in_life_expectancy_usa_file = all_usa_raw_df[
+            self.STATE_INPUT_COLUMN_NAME
+        ].unique()
+
+        # Find which states are missing from the expected set.
+        states_missing = sorted(
```
emma-nechamkin, line 84: this is a confusing way to do this and relies on sorting -- when you could just check the set difference and the set union and ensure it's the same as the set of states.

emma-nechamkin: oh also "nit"

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -23,32 +35,122 @@ def __init__(self):
...
         )
 
+        combined_df = pd.concat(
+            objs=[all_usa_raw_df, maine_raw_df, wisconsin_raw_df],
+            ignore_index=True,
+            verify_integrity=True,
+            axis=0,
+        )
+
+        states_in_combined_df = combined_df[
```
emma-nechamkin, line 134: again this is imo a slightly confusing way to do this! 

I also wonder -- should we move something like this into a util and validate all of the data we expect to be nationally comprehensive with this, rather than catch these errors kind of one off?

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -23,32 +35,122 @@ def __init__(self):
...
+            self.STATE_INPUT_COLUMN_NAME
+        ].unique()
+
+        # Find which states are missing from the combined df.
```
emma-nechamkin, line 138: I also think there's a cleaner way to do this -- like, you could test the length of the number of states represented?

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -2,19 +2,31 @@
...
 
 logger = get_module_logger(__name__)
 
 
 class CDCLifeExpectancy(ExtractTransformLoad):
     def __init__(self):
-        self.FILE_URL: str = "https://ftp.cdc.gov/pub/Health_Statistics/NCHS/Datasets/NVSS/USALEEP/CSV/US_A.CSV"
+        self.USA_FILE_URL: str = "https://ftp.cdc.gov/pub/Health_Statistics/NCHS/Datasets/NVSS/USALEEP/CSV/US_A.CSV"
+
+        # For some reason, LEEP does not include Maine or Wisconsin in its "All of
```
emma-nechamkin, line 19: This is great!

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -2,19 +2,31 @@
...
+        # USA" file. Load these separately.
+        self.WISCONSIN_FILE_URL: str = "https://ftp.cdc.gov/pub/Health_Statistics/NCHS/Datasets/NVSS/USALEEP/CSV/WI_A.CSV"
+        self.MAINE_FILE_URL: str = "https://ftp.cdc.gov/pub/Health_Statistics/NCHS/Datasets/NVSS/USALEEP/CSV/ME_A.CSV"
+
         self.OUTPUT_PATH: Path = (
             self.DATA_PATH / "dataset" / "cdc_life_expectancy"
         )
 
         self.TRACT_INPUT_COLUMN_NAME = "Tract ID"
+        self.STATE_INPUT_COLUMN_NAME = "STATE2KX"
```
emma-nechamkin, line 29: a class design question: why can't this just be a constant for the file? (this is not a nit, but a genuine question)

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -2,19 +2,31 @@
...
         self.LIFE_EXPECTANCY_FIELD_NAME = "Life expectancy (years)"
```
emma-nechamkin, line 30: As we are going back into these, I'm wondering if it's possible for us to pull these names from `field_names`? I think that might make it a bit easier from a "constants keep changing" perspective. I of course am too lazy to do this myself, but just a thought!

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -23,32 +35,122 @@ def __init__(self):
             self.LIFE_EXPECTANCY_FIELD_NAME,
         ]
 
+        # Set some constants that will be helpful for debugging the source data later.
+        self.STATE_FIPS_CODES = get_state_fips_codes(self.DATA_PATH)
+
+        self.EXPECTED_STATES_SET = (
+            set(self.STATE_FIPS_CODES)
```
emma-nechamkin, line 42: Same question about constant vs variable :)

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -23,32 +35,122 @@ def __init__(self):
...
+            # We don't expect LEEP to have data for island areas or Puerto Rico.
+            - set(TILES_ISLAND_AREA_FIPS_CODES)
+            - set(TILES_PUERTO_RICO_FIPS_CODE)
+        )
+
+        # These states are currently missing from LEEP's whole USA file.
```
emma-nechamkin, line 48: Thank you!

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -23,32 +35,122 @@ def __init__(self):
...
+        )
 
-        self.raw_df = pd.read_csv(
-            filepath_or_buffer=download_file_name,
-            dtype={
-                # The following need to remain as strings for all of their digits, not get converted to numbers.
-                self.TRACT_INPUT_COLUMN_NAME: "string",
-            },
+        logger.info("Downloading data for Wisconsin")
+        wisconsin_download_file_name = (
```
emma-nechamkin, line 113: nit: can we make these three steps into a single function call? ("download and prep data" or something?) It might be DRYer?

emma-nechamkin (approved):
This is so rad! Thank you for updating and adding the data integrity checks!

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
lucasmbrown-usds: good point!! fixing now

data/data-pipeline/data_pipeline/config.py
```diff
@@ -12,6 +12,7 @@
 
 # set root dir
 settings.APP_ROOT = pathlib.Path(data_pipeline.__file__).resolve().parent
+settings.DATA_PATH = settings.APP_ROOT / "data"
```
lucasmbrown-usds, line 15: I moved this here b/c i needed to access this value in the utils, and I needed to remove a circular import between the ETL base class and the utils.

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -82,6 +85,19 @@ class ExtractTransformLoad:
     # NULL_REPRESENTATION is how nulls are represented on the input field
     NULL_REPRESENTATION: str = None
 
+    # Whether this ETL contains data for the nation (the US states)
+    NATION_EXPECTED_IN_DATA: bool = True
```
lucasmbrown-usds, line 89: I set these to sensible defaults for most data sources.

lucasmbrown-usds: See also: https://github.com/usds/justice40-tool/issues/1892

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -1,20 +1,36 @@
...
+from data_pipeline.etl.sources.census.etl_utils import get_state_fips_codes
 from data_pipeline.utils import get_module_logger, download_file_from_url
 
 logger = get_module_logger(__name__)
 
 
 class CDCLifeExpectancy(ExtractTransformLoad):
     def __init__(self):
-        self.FILE_URL: str = "https://ftp.cdc.gov/pub/Health_Statistics/NCHS/Datasets/NVSS/USALEEP/CSV/US_A.CSV"
+        self.GEO_LEVEL = ValidGeoLevel.CENSUS_TRACT
```
lucasmbrown-usds, line 16: See https://github.com/usds/justice40-tool/issues/1891

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
lucasmbrown-usds: Oh it totally could just be a file constant and not a property on the instance. For some reason, we started writing these as instance properties and we've kept at it. Now I'm just trying to be consistent. 🤷‍♂️

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
lucasmbrown-usds: good point - fixed!

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -131,6 +131,59 @@
 # 60: American Samoa, 66: Guam, 69: N. Mariana Islands, 78: US Virgin Islands
 TILES_ISLAND_AREA_FIPS_CODES = ["60", "66", "69", "78"]
 TILES_PUERTO_RICO_FIPS_CODE = ["72"]
+TILES_NATION_FIPS_CODE = [
```
mattbowen-usds, line 134: I feel unusually happy these are here now

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
mattbowen-usds: FWIW I've been setting GEO_LEVEL as a classvar and not an instance var, since it can be

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -1,58 +1,137 @@
...
+            / "cdc_life_expectancy"
+            / "usa.csv",
+        )
+
+        # Check which states are missing
+        states_in_life_expectancy_usa_file = list(
+            all_usa_raw_df[self.STATE_INPUT_COLUMN_NAME].unique()
+        )
+
+        # Expect that PR, Island Areas, and Maine/Wisconsin are missing
```
mattbowen-usds, line 81: I can't decide if this is over-engineering or not, but given the missing codes are a bug anyway, it _almost_ feels worth doing the extra work to just look at the states we got back and download any missing ones but inferring them, since the URLs can be constructed pretty easily. Then, if ME and WI come back, we'll just stop doing the extra work, and if they drop MD randomly, we would automatically recover from that too. But, I also get that that hides some weird stuff happening on a data partner from us and maybe we want to explicitly fail if anything changes on their side...

data/data-pipeline/data_pipeline/etl/score/tests/test_etl_utils.py
```diff
@@ -70,3 +73,265 @@ def test_floor_series():
         match="Argument series must be of type pandas series, not of type list.",
     ):
         floor_series(invalid_type, number_of_decimals=3)
+
+
+def test_compare_to_list_of_expected_state_fips_codes():
+    fips_codes_test_1 = [
```
mattbowen-usds, line 79: It feels like these would be easier to read if you just imported all the FIPS codes from constants then dropped a few from each list, so it's obvious what's missing.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -305,3 +312,73 @@ def create_codebook(
     return merged_codebook_df[constants.CODEBOOK_COLUMNS].rename(
         columns={constants.CEJST_SCORE_COLUMN_NAME: "Description"}
     )
+
+
+def compare_to_list_of_expected_state_fips_codes(
+    actual_state_fips_codes: typing.List[str],
+    nation_expected: bool = True,
```
mattbowen-usds, line 319: I want this to be conus_expected but I understand why no normal person would do that.

mattbowen-usds (approved):
Just nits, with one substantive design comment that you are free to take or leave.

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -399,5 +452,5 @@
     # that use null to signify missing information in a boolean field.
     field_names.ELIGIBLE_FUDS_BINARY_FIELD_NAME,
     field_names.AML_BOOLEAN,
-    field_names.HISTORIC_REDLINING_SCORE_EXCEEDED
+    field_names.HISTORIC_REDLINING_SCORE_EXCEEDED,
```
emma-nechamkin, line 455: ty!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: i kinda like conus_expected lol

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
lucasmbrown-usds: Yeah - I get what you're saying. But I think I'd prefer to be explicit and fail fast rather than engineer something that probably but not 100% definitely works to fully recover. 

I mainly put this code in because @BethMattern is gonna send an email to the CDC asking them to fix this source data, so I wanted us to have a heads-up when it changes.

data/data-pipeline/data_pipeline/etl/score/tests/test_etl_utils.py
lucasmbrown-usds: Good point. I'll keep the first one explicit (so the test also checks whether the constants change), but I'll make the others subtractions to improve readability.

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
mattbowen-usds: Works for me.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
lucasmbrown-usds: The problem - the nation is not actually the same as CONUS. Continental US does *not* include Alaska and Hawaii, right?

data/data-pipeline/data_pipeline/etl/score/tests/test_etl_utils.py
```diff
@@ -70,3 +73,157 @@ def test_floor_series():
...
+            island_areas_expected=False,
+        )
+    partial_expected_error_message = (
+        "FIPS state codes in the data that were not expected:\n"
+        "['60', '66', '69', '78']\n"
+    )
+    assert partial_expected_error_message in str(exception_info.value)
+
+    # List missing PR and Guam
+    fips_codes_test_2 = [x for x in fips_codes_test_1 if x not in ["66", "72"]]
```
lucasmbrown-usds, line 168: does this feel satisfying, @mattbowen-usds ?

data/data-pipeline/data_pipeline/etl/sources/ncld_nature_deprived/etl.py
```diff
@@ -1,77 +0,0 @@
-# pylint: disable=unsubscriptable-object
```
lucasmbrown-usds, line 1: I'm not sure how we got to this state but we had one folder called `ncld_*` and one called `nlcd_*`. I deleted the one we're not using that has the typo.

data/data-pipeline/data_pipeline/tests/sources/child_opportunity_index/test_etl.py
```diff
@@ -59,7 +59,7 @@ def test_init(self, mock_etl, mock_paths):
 
     def test_get_output_file_path(self, mock_etl, mock_paths):
         """Tests the right file name is returned."""
-        etl = self._ETL_CLASS()
+        etl = self._get_instance_of_etl_class()
```
lucasmbrown-usds, line 62: @mattbowen-usds and I did this to (1) improve test style and (2) make the monkey-patching of the class to only expect the states in the test fixtures work.

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -289,6 +307,22 @@ def validate(self) -> None:
                         f"`{geo_field}`."
                     )
 
+        # Check whether data contains expected states
+        states_in_output_df = list(
```
emma-nechamkin, line 311: nit -- try: `
self.output_df[self.GEOID_TRACT_FIELD_NAME].str[0:2].unique().to_list()``
This should be a string and not need to be coerced, and it's more in-line with pandas to use `to_list()`vs explicitly casting it.

data/data-pipeline/data_pipeline/etl/score/constants.py
emma-nechamkin: a pretty dumb question -- why not just do a list comprehension dropping 52?

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: i would actually do conus_expected AND not_conus_expected as flags maybe to make it explicit. A few datasets like NCLD are just conus, some datasets are conus, nconus, and PR, some are comprehensive, etc. In the same way you have pr / islands separated out, i think conus / nconus could be useful.

data/data-pipeline/data_pipeline/etl/score/tests/test_etl_utils.py
mattbowen-usds: it does! thank you!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -305,3 +312,89 @@ def create_codebook(
...
+    actual_state_fips_codes_set = set(actual_state_fips_codes)
+
+    # Start with the list of all FIPS codes for all states and territories.
+    expected_states_set = set(get_state_fips_codes(settings.DATA_PATH))
+
+    # If nation (states and DC) are not expected to be included, remove it from the
+    # expected
+    # states set.
+    if not nation_expected:
+        expected_states_set = expected_states_set - set(TILES_NATION_FIPS_CODE)
```
emma-nechamkin, line 361: you could just do `expected_states_set.difference(TILES_NATION_FIPS_CODE)` i think, without the second casting to set -- I happen to prefer the `set.difference` and `set.union` language because I think it's clearer, but this is pretty obviously a personal preference you should feel free to disregard.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -305,3 +312,89 @@ def create_codebook(
...
+        additional_fips_codes_not_expected
+    )
+
+    dataset_name_phrase = (
+        f" for dataset `{dataset_name}`" if dataset_name is not None else ""
+    )
+
+    if expected_states_set != actual_state_fips_codes_set:
+        raise ValueError(
+            f"The states and territories in the data{dataset_name_phrase} are not "
```
emma-nechamkin, line 389: missing a space in line 389

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -305,3 +312,89 @@ def create_codebook(
...
+        )
+
+    # If island areas are not expected to be included, remove them from the expected
+    # states set.
+    if not island_areas_expected:
+        expected_states_set = expected_states_set - set(
+            TILES_ISLAND_AREA_FIPS_CODES
+        )
+
+    # If additional FIPS codes are not expected to be included, remove them from the
```
emma-nechamkin, line 377: this is a really annoying quibble and i am sorry. 

Above you have a control statement to fill additional fips codes if it is none, and then to  do this. I wonder why not make it default as None of type set, and then "if not none" do this (vs fill it with the empty list if None and do this calculation every time)? This is a nitty nit but 🤷 i learn from the best

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
```diff
@@ -305,3 +312,89 @@ def create_codebook(
...
+    # expected states set.
+    expected_states_set = expected_states_set - set(
+        additional_fips_codes_not_expected
+    )
+
+    dataset_name_phrase = (
```
emma-nechamkin, line 383: I low key don't think this is necessary and it just adds a lot of text -- most of the classes don't have a dataset name right now!

emma-nechamkin: and, if you want to keep it, it should have a space so that my comment about spaces becomes irrelevant

data/data-pipeline/data_pipeline/etl/sources/fsf_wildfire_risk/etl.py
```diff
@@ -16,6 +16,10 @@ class WildfireRiskETL(ExtractTransformLoad):
     NAME = "fsf_wildfire_risk"
     SOURCE_URL = settings.AWS_JUSTICE40_DATASOURCES_URL + "/fsf_fire.zip"
     GEO_LEVEL = ValidGeoLevel.CENSUS_TRACT
+    PUERTO_RICO_EXPECTED_IN_DATA = False
+
+    # Alaska and Hawaii are missing
+    EXPECTED_MISSING_STATES = ["02", "15"]
```
emma-nechamkin, line 22: this would also be easier if we did conus nconus above!

emma-nechamkin (commented):
this is great and i have only the most annoying nits

data/data-pipeline/data_pipeline/etl/base.py
lucasmbrown-usds: cool, i didn't know that about the `to_list` vs casting!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
lucasmbrown-usds: The name logging actually was very helpful debugging *which* ETL classes were failing when they were running in parallel. I was just seeing a bunch of "these states are not as expected!" but because the logs were jumbled from parallel runs, I had no idea which ETL was failing.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
lucasmbrown-usds: done!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
lucasmbrown-usds: done!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
lucasmbrown-usds: Oh, well, there will always be a weird-looking lack of a space on this. Because it'll either be `data{dataset_name_phrase} are` or `data {dataset_name_phrase}are` but if `dataset_name_phrase` is blank you don't want the text to read `The states and territories in the data  are not` with two spaces between data and are. 

Does that make sense? Or am I missing a cleaner way to do this.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
lucasmbrown-usds: Hmm, interesting. I see your point. I started without having the nation explicitly defined so it made sense to subtract, but now we have everything. 

The one weird part of starting "positively" would be you'd have to say "ALASKA_AND_HAWAII_INCLUDED=TRUE" for some sources but then ALSO that "15" is an expected missing state. (Some have Alaska but not Hawaii, etc.) 

I'm tempted to just leave it for now to avoid the refactor?

mattbowen-usds (approved):
At this point I don't have anything really substantive. Just one question.

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: wait @mattbowen-usds i know they are lists! i'm saying set.differnce(list) is cleaner than set - set(list) imo

mattbowen-usds: Ohhh I'm sorry, I was thinking of what I do with indexes. Apologies!

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: i see i see, i am sold

data/data-pipeline/data_pipeline/etl/score/etl_utils.py
emma-nechamkin: that's fine by me

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -86,7 +87,23 @@ def setup_method(self, _method, filename=__file__):
...
-        return self._ETL_CLASS()
+        etl_class = self._ETL_CLASS()
+
+        # Find out what unique state codes are present in the test fixture data.
+        states_expected_from_fixtures = {
+            x[0:2] for x in self._FIXTURES_SHARED_TRACT_IDS
+        }
+
+        # Set values to match test fixtures
+        etl_class.EXPECTED_MISSING_STATES = [
```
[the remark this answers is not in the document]
lucasmbrown-usds: Oh, this is actually because `etl_class.EXPECTED_MISSING_STATES` in the test is going to need to be the *opposite* of `states_expected_from_fixtures`. `states_expected_from_fixtures` will be just two states `['06','15']` and etl_class.EXPECTED_MISSING_STATES is going to be every single FIPS code except those two.

Issue #1885, opened 2022-09-07
Restrict transportation barriers to pop > threshold


emma-nechamkin:
Tracts to look at 

['01097990000',
 '01015981901',
 '01015981902',
 '04027980003',
 '04027980004',
 '04012980000',
 '04013980100',
 '05119003205',
 '06071980100',
 '06037980004',
 '06083980100',
 '06053980000',
 '10003980100',
 '12083980000',
 '12057980400',
 '12099990100',
 '12111980000',
 '12043990000',
 '12051990000',
 '12105980000',
 '12093990000',
 '12085990100',
 '12055980100',
 '13183980000',
 '13089980000',
 '13063980000',
 '13121980000',
 '15001990300',
 '19193940200',
 '20209043804',
 '21221980100',
 '21047980100',
 '21143980100',
 '21163980100',
 '21111980100',
 '22073980000',
 '22075990000',
 '22071980100',
 '22017980000',
 '22019980000',
 '22033980000',
 '23029990000',
 '24005980100',
 '26033980100',
 '26163985600',
 '26163985700',
 '26095980000',
 '26131990100',
 '26099982200',
 '27123980000',
 '28143980000',
 '28047980000',
 '32031990100',
 '32023980500',
 '36029990000',
 '36081010701',
 '37095990100',
 '37087980100',
 '37051980100',
 '37055990200',
 '37173980200',
 '37119980100',
 '39035980500',
 '41041990100',
 '42003980400',
 '42079980100',
 '45061920301',
 '45041980100',
 '45029990100',
 '45003980100',
 '45011980100',
 '47029980100',
 '47155980100',
 '47157980100',
 '47157980200',
 '47157980300',
 '48375980000',
 '48027980003',
 '48061980001',
 '48061980100',
 '48453980000',
 '48465980000',
 '48029980002',
 '48303980000',
 '48423980000',
 '48039990000',
 '48201980000',
 '48485980000',
 '48215980000',
 '49035980000',
 '51001980200',
 '51001990100',
 '51131990100',
 '51670980100',
 '51590980100',
 '51053980100',
 '53055990100',
 '53061990100',
 '53057990100',
 '53049990100',
 '53021980100']

PR #1888, opened 2022-09-07
Updating traffic barriers

emma-nechamkin:
#1885 
The goal of this is to eliminate "low population tracts" from the traffic barriers percentile calculation.


emma-nechamkin:
ugh i messed up the pr

PR #1896, opened 2022-09-09
1.0 side panel indicator designs

vim-usds:
## Purpose
- Update the indicator styles in the side panel for 1.0
 
## Note:
Styling the categories entirely blue (with white +/-), didn't seem possible to me before browsers adopted the new CSS _psuedo-selector_ of `:has`. Check [here](https://caniuse.com/?search=has) to see the latest browser adoption.

Here's a pic showing what browser versions support the `:has` psuedo-selector:
Screen Shot 2022-09-09 at 12 00 52 AM

In summary, the 
- **Chrome 105+** (anyone who updated their browser after 8/22/22) will show disadv. categories entirely blue with white +/-, otherwise fall back to beta style (blue dot)
- **Edge 105+** (anyone who updated their browser after 8/31/22) will show disadv. categories entirely blue with white +/-, otherwise fall back to beta style (blue dot)
- **Safari 15.4+** (anyone who updated their browser after 5/16/22) will show disadv. categories entirely blue with white +/-, otherwise fall back to beta style (blue dot)

BTW - I've never been so close to the cutting edge on browser features and feels pretty amazing! 🥳 

## Kudos
This design update actually _removed_ more code than it added. Isn't always possible, however when it is, let's celebrate! 🎉 

### QA link [here](https://screeningtool-staging.geoplatform.gov/1896-9edcef/en/?flags=stage_hash=1822/60164c863791af413cbc3838180f3fb7ed991d1b#3.02/33.61/-97.97)

- [x] remove all `up` and `down` 
- [x] missing data should have a _Value_ of missing data should be '--' with no _Icon_
- [x] when data is missing, the _SubText_ should say "missing data". Reduce size to fit in one line.
- [x] instead of highlighting the entire indicator(with a baby blue background), only highlight the _Value_ (dark blue background around the value) when the indicator is above the threshold. change the text color of the value to white.
- [x] instead of a blue dot on the category, change the entire background (to dark blue) of the category when it is above the associated threshold(s), otherwise leave as is. change the text color and the icon color (the plus sign) to white.
- [x] test on mobile (iOS and Android)
- [x] test on desktop (Chrome, Safari and Firefox)
- [x] review all ind titles and desc from content 1.0 doc
- [x] category title, indicator name, desc, value, subtext should all be grey-warm-90
- [x] _Title_ and _Value_ should have the same font treatment (weight and size as spacer)
- [x] remove initial spacer (at least one)
- [x] AND space remove bg color and ensure that vertical spacing b/w indicator is the same including spacers (check each category to make sure it applies equally)
- [x] all subtext in one line
- [x] on the value that has a background color, tighten up, equal spacing on all sides (2px left and right, 1px top and bottom)
- [x] Methodology version (pt, pb = 1rem)
- [x] update Methodology version to 1.0
- [ ] tooltip on indicator value (on HOLC - tbd)

### DIFF
--- a/client/src/components/AreaDetail/AreaDetail.tsx
+++ b/client/src/components/AreaDetail/AreaDetail.tsx
@@ -8,7 +8,6 @@ import {Accordion, Button} from '@trussworks/react-uswds';
 import Category from '../Category';
 import TractDemographics from '../TractDemographics';
 import DisadvantageDot from '../DisadvantageDot';
-import ExceedBurden from '../ExceedBurden';
 import Indicator from '../Indicator';
 import TractInfo from '../TractInfo';
 
@@ -745,12 +744,6 @@ const AreaDetail = ({properties, hash, isCensusLayerSelected}: IAreaDetailProps)
     title: <Category name={category.titleText} isDisadvantaged={category.isDisadvagtaged} />,
     content: (
       <>
-        {/* Exceeds one or more burdens */}
-        <ExceedBurden
-          text={EXPLORE_COPY.SIDE_PANEL_SPACERS.EXCEED_ONE_OR_MORE}
-          isBurdened={category.isExceed1MoreBurden}
-        />
-
         {/* Indicators - filters then map */}
         {category.indicators
             .filter(indicatorFilter(EXPLORE_COPY.SIDE_PANEL_INDICATORS.HIST_UNDERINVEST))
--- a/client/src/components/AreaDetail/AreaDetail.tsx
+++ b/client/src/components/AreaDetail/AreaDetail.tsx
@@ -315,9 +315,9 @@ const AreaDetail = ({properties, hash, isCensusLayerSelected}: IAreaDetailProps)
 
 
   // Energy category
-  const energyBurden: indicatorInfo = {
-    label: intl.formatMessage(EXPLORE_COPY.SIDE_PANEL_INDICATORS.ENERGY_BURDEN),
-    description: intl.formatMessage(EXPLORE_COPY.SIDE_PANEL_INDICATOR_DESCRIPTION.ENERGY_BURDEN),
+  const energyCost: indicatorInfo = {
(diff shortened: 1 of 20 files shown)


esfoobar-usds (approved):
🍏

KameronKerger:
there is an alignment issue on safari 
Screen Shot 2022-09-14 at 12 09 55 PM
everything else seems ok although I am suspicious of why this tract is missing the FSF data (it is also missing wastewater for what it is worth)
Screen Shot 2022-09-14 at 12 10 33 PM
Screen Shot 2022-09-14 at 12 10 39 PM

vim-usds:
@KameronKerger - What version of Safari are you running? I'm unable to see that alignment issue on my end?

Screen Shot 2022-09-14 at 2 59 59 PM

KameronKerger:
Version 15.0 (17612.1.29.5)

vim-usds:
> Version 15.0 (17612.1.29.5)

Did you try clearing your cache?

KameronKerger:
problem persists after clearing cache
Screen Shot 2022-09-14 at 3 46 47 PM

vim-usds:
> everything else seems ok although I am suspicious of why this tract is missing the FSF data (it is also missing wastewater for what it is worth)

> tractmissingfireflood


In regards to the missing data (above); when downloading the high tiles geojson: we can see ([here](https://jsonhero.io/j/MKU4bLcOfCKn)) that the wildfire (WFR_PFS) and flood (FLD_PFS) BE signals are both `null` in the tiles for _this_ census tract. This is the reason why they are showing up as missing in the side panel.

mattbowen-usds:
Looking at this, it looks like none of CATRON county, NM (FIPS code 35003) is in either the wildfire or flood datasets that the backend uses. I'll look at how those were constructed upstream to see if i can see a reason those counties would be missing, since every other NM county is in the fire tracts at least.

PR #1899, opened 2022-09-15
Add tests for all non-census sources

### LINKED ISSUE #1848: Back end "pipeline-y" tests
The following is a list of tests that we should add for the backend to ensure that we are passing the correct values forward into the front end. 

__Per data source:__ The goal here is tracts in should equal tracts out
- [x] All census tracts with a entry per numeric column (percentile, etc) should be represented in the `usa.csv` for that data source
- [x] All census tracts that do not have a entry per numeric column should NOT be included in the `usa.csv` for that source, column combination 

__Tiles and data pipeline:__ The goal here is to smoke test correctness
- [x] We should check for wrong variable ported over to the tiles by making sure the column name for each thing we want in `usa.csv` has the same value as the column name for the tiles csv. This is NOT the same as checking that the column has been renamed. Rather, this requires looking at the methodology of the tool, seeing what should be represented, and ensuring that variable is indeed represented. 
- [ ] Check for USVI and Guam in tiles
- [x] We should check that even census tracts that have NULLs for every indicator get included in the tiles and in the csv and are never dropped

__Tests we don't need right now__
- [ ] Threshold counts in island areas don’t match number of thresholds possible --> we've dropped this for now
- [ ] Codebook test (I think for now, this is fine -- will be no longer needed with refactor?)

mattbowen-usds:
Closes #1848

NOTE: This **does not** have tests for census-based sources. But everything else!

### DIFF
--- a/data/data-pipeline/data_pipeline/etl/sources/fsf_flood_risk/etl.py
+++ b/data/data-pipeline/data_pipeline/etl/sources/fsf_flood_risk/etl.py
@@ -14,6 +14,8 @@ class FloodRiskETL(ExtractTransformLoad):
     """ETL class for the First Street Foundation flood risk dataset"""
 
     NAME = "fsf_flood_risk"
+    # These data were emailed to the J40 team while first street got
+    # their official data sharing channels setup.
     SOURCE_URL = settings.AWS_JUSTICE40_DATASOURCES_URL + "/fsf_flood.zip"
     GEO_LEVEL = ValidGeoLevel.CENSUS_TRACT
     LOAD_YAML_CONFIG: bool = True
--- a/data/data-pipeline/data_pipeline/etl/sources/fsf_wildfire_risk/etl.py
+++ b/data/data-pipeline/data_pipeline/etl/sources/fsf_wildfire_risk/etl.py
@@ -14,6 +14,8 @@ class WildfireRiskETL(ExtractTransformLoad):
     """ETL class for the First Street Foundation wildfire risk dataset"""
 
     NAME = "fsf_wildfire_risk"
+    # These data were emailed to the J40 team while first street got
+    # their official data sharing channels setup.
     SOURCE_URL = settings.AWS_JUSTICE40_DATASOURCES_URL + "/fsf_fire.zip"
     GEO_LEVEL = ValidGeoLevel.CENSUS_TRACT
     PUERTO_RICO_EXPECTED_IN_DATA = False


esfoobar-usds (approved):
🍏

data/data-pipeline/data_pipeline/etl/base.py
```diff
@@ -115,56 +115,59 @@ class ExtractTransformLoad:
...
+        # and set the constants
+        cls.COLUMNS_TO_KEEP = [
+            cls.GEOID_TRACT_FIELD_NAME,  # always index with geoid tract id
+        ]
+        for field in dataset_config["load_fields"]:
+            cls.COLUMNS_TO_KEEP.append(field["long_name"])
+            setattr(cls, field["df_field_name"], field["long_name"])
+
+            # set the constants for the class
+            setattr(cls, field["df_field_name"], field["long_name"])
```
lucasmbrown-usds, line 169: duplicate line?

data/data-pipeline/data_pipeline/tests/sources/cdc_life_expectancy/test_etl.py
```diff
@@ -0,0 +1,112 @@
...
+            data_pipeline/tests/sources/cdc_life_expectancy/test_etl.py::TestClassNameETL::<testname>
+            --snapshot-update
+    """
+
+    _ETL_CLASS = CDCLifeExpectancy
+
+    _SAMPLE_DATA_PATH = pathlib.Path(__file__).parents[0] / "data"
+    _SAMPLE_DATA_FILE_NAME = "US_A.CSV"
+    _SAMPLE_DATA_ZIP_FILE_NAME = None
+    _EXTRACT_TMP_FOLDER_NAME = "CDCLifeExpectanc"
```
lucasmbrown-usds, line 27: nitty nit: you're missing a `y` 😂

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -124,22 +124,37 @@ def _setup_etl_instance_and_run_extract(
...
+                ) as file:
+                    file_contents = file.read()
             response_mock = requests.Response()
             response_mock.status_code = 200
             # pylint: disable=protected-access
             response_mock._content = file_contents
             # Return text fixture:
             requests_mock.get = mock.MagicMock(return_value=response_mock)
-
+            mock_get_state_fips_codes.return_value = [
```
lucasmbrown-usds, line 155: Hmm. Will this cause problems with the "expected states in data" logic? Or at least be redundant with it? 

We may want the mock to just return all states/DC/territories, but then have the code for `_get_instance_of_etl_class` continue to use a more limited list using all this code: 

```
# Find out what unique state codes are present in the test fixture data.
        states_expected_from_fixtures = {
            x[0:2] for x in self._FIXTURES_SHARED_TRACT_IDS
        }

        # Set values to match test fixtures
        etl_class.EXPECTED_MISSING_STATES = [
            x
            for x in TILES_CONTINENTAL_US_FIPS_CODE
            + TILES_ALASKA_AND_HAWAII_FIPS_CODE
            if x not in states_expected_from_fixtures
        ]
        etl_class.PUERTO_RICO_EXPECTED_IN_DATA = False
        etl_class.ISLAND_AREAS_EXPECTED_IN_DATA = False
        etl_class.ALASKA_AND_HAWAII_EXPECTED_IN_DATA = True
```

Alternatively, we can proceed with the approach in this line and mocking the return value of  `get_state_fips_codes` to just the codes we expect from the fixtures, but then we should delete all that code that's now redundant in `_get_instance_of_etl_class`.

data/data-pipeline/data_pipeline/tests/sources/example/test_etl.py
```diff
@@ -225,27 +240,33 @@ def test_sample_data_exists(self):
         """This will test that the sample data exists where it's supposed to as it's supposed to
         As per conversation with Jorge, here we can *just* test that the zip file exists.
         """
-        assert (
-            self._SAMPLE_DATA_PATH / self._SAMPLE_DATA_ZIP_FILE_NAME
-        ).exists()
+        if self._SAMPLE_DATA_ZIP_FILE_NAME is not None:
```
lucasmbrown-usds, line 243: nice

data/data-pipeline/data_pipeline/tests/sources/cdc_life_expectancy/test_etl.py
```diff
@@ -0,0 +1,112 @@
...
+    ]
+
+    def setup_method(self, _method, filename=__file__):
+        """Invoke `setup_method` from Parent, but using the current file name.
+
+        This code can be copied identically between all child classes.
+        """
+        super().setup_method(_method=_method, filename=filename)
+
+    def _setup_etl_instance_and_run_extract(
```
lucasmbrown-usds, line 41: is there a reason this needs to be re-implemented here, and we can't just use the method from the base class?

data/data-pipeline/data_pipeline/tests/sources/cdc_places/test_etl.py
```diff
@@ -0,0 +1,25 @@
...
+    _EXTRACT_TMP_FOLDER_NAME = "cdc_places"
+
+    def setup_method(self, _method, filename=__file__):
+        """Invoke `setup_method` from Parent, but using the current file name.
+
+        This code can be copied identically between all child classes.
+        """
+        super().setup_method(_method=_method, filename=filename)
+
+    def test_sample_data_exists(self):
```
lucasmbrown-usds, line 21: Any reason why this can't be just a base class test?

data/data-pipeline/data_pipeline/tests/sources/ejscreen/test_etl.py
```diff
@@ -0,0 +1,19 @@
+import pathlib
+from data_pipeline.tests.sources.example.test_etl import TestETL
+from data_pipeline.etl.sources.ejscreen.etl import EJSCREENETL
+
+
+class TestEJSCREENETL(TestETL):
```
lucasmbrown-usds, line 6: nice clean test inheriting everything!

data/data-pipeline/data_pipeline/tests/sources/dot_travel_composite/test_etl.py
```diff
@@ -0,0 +1,45 @@
...
+    def test_extract_produces_valid_data(self, mock_etl, mock_paths):
+        etl = self._setup_etl_instance_and_run_extract(
+            mock_etl=mock_etl,
+            mock_paths=mock_paths,
+        )
+        df = gpd.read_file(
+            etl.get_tmp_path() / self._SAMPLE_DATA_FILE_NAME,
+            dtype={etl.GEOID_TRACT_FIELD_NAME: str},
+        )
+        assert df.shape[0] == 30
```
lucasmbrown-usds, line 33: Idea: we could make expected rows and expected columns into class attributes on the child Test<Name>ETL classes, and then make this test part of the base class? Then we'd have this test coverage everywhere.

data/data-pipeline/data_pipeline/tests/sources/historic_redlining/test_etl.py
```diff
@@ -0,0 +1,66 @@
...
+        This code can be copied identically between all child classes.
+        """
+        super().setup_method(_method=_method, filename=filename)
+
+    def test_extract_produces_valid_data(self, snapshot, mock_etl, mock_paths):
+        etl = self._setup_etl_instance_and_run_extract(
+            mock_etl=mock_etl,
+            mock_paths=mock_paths,
+        )
+        tmp_df = pd.read_excel(
```
lucasmbrown-usds, line 30: What's going on in this test? You load the input data as an excel and check the length -- but don't you mean to check the output data? 

Also, see question above about, can we just make a single test for shape that uses class attributes for the expected shape.

data/data-pipeline/data_pipeline/tests/sources/historic_redlining/test_etl.py
```diff
@@ -0,0 +1,66 @@
...
+            etl.get_tmp_path() / self._SAMPLE_DATA_FILE_NAME,
+            dtype={etl.GEOID_TRACT_FIELD_NAME: str},
+        )
+        assert tmp_df.shape == (15, 5)
+
+    def test_load_base(self, snapshot, mock_etl, mock_paths):
```
lucasmbrown-usds, line 36: Why is this a specialized test and not inherited from the base class?

lucasmbrown-usds (commented):
I left a lot of questions, forgive me if I'm just a little dense today!

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -16,7 +16,12 @@ class CDCLifeExpectancy(ExtractTransformLoad):
     GEO_LEVEL = ValidGeoLevel.CENSUS_TRACT
     PUERTO_RICO_EXPECTED_IN_DATA = False
 
+    NAME = "cdc_life_expectancy"
+
     USA_FILE_URL: str = "https://ftp.cdc.gov/pub/Health_Statistics/NCHS/Datasets/NVSS/USALEEP/CSV/US_A.CSV"
+    LOAD_YAML_CONFIG: bool = False
```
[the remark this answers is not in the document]
mattbowen-usds: I made the YAML configs opt-in, since most sources don't use them, but I want the other refactor features around the tests.

data/data-pipeline/data_pipeline/tests/sources/doe_energy_burden/test_etl.py
```diff
@@ -59,3 +59,13 @@ def test_get_output_file_path(self, mock_etl, mock_paths):
             data_path / "dataset" / "doe_energy_burden" / "usa.csv"
         )
         assert output_file_path == expected_output_file_path
+
+    def test_tract_id_lengths(self, mock_etl, mock_paths):
```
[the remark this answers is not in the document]
mattbowen-usds: Good call --- I had thought about it just for the places where we specifically were modifying, but you are right; we can just check this everywhere.

data/data-pipeline/data_pipeline/tests/sources/historic_redlining/test_etl.py
mattbowen-usds: Here's the test it overrides:
```    def test_extract_produces_valid_data(self, snapshot, mock_etl, mock_paths):
        """Tests the extract method.

        Here we are verifying that the data that we extract is "readable". I added a snapshot to be thorough,
        but @Jorge -- do you think this is necessary?
        """
        etl = self._setup_etl_instance_and_run_extract(
            mock_etl=mock_etl,
            mock_paths=mock_paths,
        )
        tmp_df = pd.read_csv(
            etl.get_tmp_path() / self._SAMPLE_DATA_FILE_NAME,
            dtype={etl.GEOID_TRACT_FIELD_NAME: str},
        )
        snapshot.assert_match(
            tmp_df.to_csv(index=False, float_format=self._FLOAT_FORMAT),
            self._EXTRACT_CSV_FILE_NAME,
        )

```

So it's really trying to verify that extract in this case correctly gets us a parsable excel file that transform can work on. I needed to override tho because it's excel and not CSV. I check the shape to verify the extracted file is roughly what we expect

data/data-pipeline/data_pipeline/tests/sources/dot_travel_composite/test_etl.py
mattbowen-usds: In general, we do, because we check snapshots of extract for CSVs. In this case, since we're coming from shapefiles, and comparing a shapefile to a CSV didn't feel super useful, I just wanted to check that the file read correctly and was shaped correctly; other later tests verify that we'll get the transformations and ouput we expect.

data/data-pipeline/data_pipeline/tests/sources/historic_redlining/test_etl.py
mattbowen-usds: The base uses read_csv, but redlining reads an xlsx, so I needed to subclass to do change that out.

data/data-pipeline/data_pipeline/etl/base.py
mattbowen-usds: good catch!

PR #1902, opened 2022-09-15
1831 Follow up

lucasmbrown-usds:
It looks like I forgot to commit and push these changes that I made while wrapping up 1831. Pushing them now.

This code causes no functional change to the code. It does two things: 

1. Uses `difference` instead of `-` to improve code style for working with sets. 
2. Removes the line `EXPECTED_MISSING_STATES = ["02", "15"]`, which is now redundant because of the line I added (in a previous pull request) of `ALASKA_AND_HAWAII_EXPECTED_IN_DATA = False`.


esfoobar-usds (approved):
🍏

Issue #1900, opened 2022-09-15
Calculate % of land overlap for census tracts that overlap with tribal boundaries


BethMattern:
Users need to see how much land of a census tract overlaps with a tribal boundary on the CEJST map. 

Once this is calculated it needs to be available for every census tract so that it can be viewed on the map side panel.

It would be preferable to have the % overlap amount and the ID and Name of the overlapping tribal boundary for census tracts in the xls and csv download files.

This will only apply to the federally recognized tribal boundaries. The Alaskan Native Villages are points on the map, so we can't calculate a % of land overlap.

lucasmbrown-usds:
Put this information in the spreadsheet too.

BethMattern:
@lucasmbrown-usds if possible include the % overlap, tribal boundary id and tribal boundary name in the spreadsheet

KameronKerger:
boundaries if multiple pls!

PR #1903, opened 2022-09-16
Issue 1900: Tribal overlap with Census tracts

lucasmbrown-usds:
Issue #1900. 

This PR creates the ETL, and merges the data in `etl_score`, but it does not put the new fields in the tiles or the download files. That's left to a next PR after we clarify our intentions. 

Example output file: 
[usa.csv](https://github.com/usds/justice40-tool/files/9580142/usa.csv)


data/data-pipeline/data_pipeline/etl/sources/geo_utils.py
```diff
@@ -18,38 +19,66 @@ def get_tract_geojson(
...
+        tribal_etl.load()
+    tribal_data = gpd.read_file(
+        GEOJSON_PATH,
+    )
+    return tribal_data
+
+
 def add_tracts_for_geometries(
-    df: gpd.GeoDataFrame, _tract_data_path: Optional[Path] = None
+    df: gpd.GeoDataFrame, tract_data: Optional[gpd.GeoDataFrame] = None
```
lucasmbrown-usds, line 59: I refactored this a bit, so I could load the Tract geojson once in the tribal overlap ETL, and not just pass around a data path.

data/data-pipeline/data_pipeline/tests/score/test_utils/data/us.geojson
```diff
@@ -2,14 +2,14 @@
...
...2, 39.304984 ], [ -76.880733, 39.305174 ], [ -76.880735, 39.305263 ], [ -76.880547, 39.305504 ], [ -76.879848, 39.305738 ], [ -76.87922, 39.305965 ], [ -76.877891, 39.306433 ], [ -76.877241, 39.306642 ], [ -76.87623, 39.306966 ], [ -76.875149, 39.307313 ], [ -76.874586, 39.3075 ], [ -76.870653, 39.308758 ], [ -76.87045, 39.308843 ], [ -76.870392, 39.30884 ], [ -76.869161, 39.308786 ], [ -76.869153, 39.308799 ], [ -76.869122, 39.308816 ], [ -76.869013, 39.308862 ], [ -76.868485, 39.309055 ], [ -76.867978, 39.309239 ], [ -76.867705, 39.309347 ], [ -76.867261, 39.309502 ], [ -76.866102, 39.309894 ], [ -76.866101, 39.309894 ], [ -76.865669, 39.310066 ], [ -76.864234, 39.310568 ], [ -76.863039, 39.310937 ], [ -76.862618, 39.311062 ], [ -76.86255, 39.311114 ], [ -76.862538, 39.311172 ], [ -76.862581, 39.311225 ], [ -76.862677, 39.311297 ], [ -76.862787, 39.311386 ], [ -76.862929, 39.311549 ], [ -76.863046, 39.311741 ], [ -76.863144, 39.311933 ], [ -76.863206, 39.312082 ], [ -76.863267, 39.312264 ], [ -76.86331, 39.31248 ], [ -76.86334, 39.312713 ], [ -76.863359, 39.313026 ], [ -76.863314, 39.313502 ], [ -76.863203, 39.314109 ], [ -76.863106, 39.314569 ], [ -76.863049, 39.31484 ] ] ] } },
```
lucasmbrown-usds, line 5: Note: all I changed here is renaming the field `GEOID10_TRACT` as `GEOID10`. This is more realistic. And also, it will otherwise yell at us now that I set `errors="raise"` in 

```
tract_data = tract_data.rename(
        columns={"GEOID10": "GEOID10_TRACT"}, errors="raise"
    )
 ```

File: data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py [hunk 3]
```diff
@@ -0,0 +1,199 @@
...
+        # Second, calculate percentage overlap.
+        # Drop the points from the Tribal data (because these cannot be joined to a
+        # (Multi)Polygon tract data frame)
+        tribal_gdf_without_points = self.tribal_gdf[
+            self.tribal_gdf.geom_type != "Point"
+        ]
+
+        # Create a measure for the entire census tract area
+        with warnings.catch_warnings():
+            # Note: geopandas will (helpfully) warn that area is not a useful metric
```
mattbowen-usds, line 107: I ran the line and got this warning:

`UserWarning: `keep_geom_type=True` in overlay resulted in 348 dropped geometries of different geometry types than df1 has. Set `keep_geom_type=False` to retain all geometries`

That was not the warning I expected (I thought it would complain about projected vs geographic CRS. Both DFs have CRSes (they're both in WSG 84/EPSG:4326), so there wouldn't be a mismatch. Any idea what that warning is about? Also I feel like the comment suggests something else is going on.

mattbowen-usds (approved):
Overall this is quite nice. I stepped through the ETL and made sure I understood why it was going what it was doing, and I do have one legit question about the suppressed warning, but overall this is great.

data/data-pipeline/data_pipeline/etl/sources/geo_utils.py
mattbowen-usds: I like that change --- I had been passing the path just to make my tests faster, so this is a nice way of adding flexibility.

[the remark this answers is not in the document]
lucasmbrown-usds: it was awesome! loved reusable code!

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -469,12 +475,15 @@ def _prepare_initial_df(self) -> pd.DataFrame:
...
             field_names.PERCENT_AGE_10_TO_64,
             field_names.PERCENT_AGE_OVER_64,
+            field_names.PERCENT_OF_TRIBAL_AREA_IN_TRACT,
         ]
 
         non_numeric_columns = [
             self.GEOID_TRACT_FIELD_NAME,
             field_names.TRACT_ELIGIBLE_FOR_NONNATURAL_THRESHOLD,
             field_names.AGRICULTURAL_VALUE_BOOL_FIELD,
+            field_names.COUNT_OF_TRIBAL_AREAS_IN_TRACT,
```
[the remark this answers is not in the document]
lucasmbrown-usds: Good point. I'll move it back. I wasn't sure where integers should go. All the numerics get converted to `floats`.

data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py
lucasmbrown-usds: Oh shoot! That is an error I missed. I built this using a subset file of the tracts to improve speed of development, it must not have been present in the subset I used. 

Overlay requires the same geometry types, but it treats MultiPolygon and Polygon as compatible. 

Because we drop the points in line 101 to create `tribal_gdf_without_points`, I wasn't expecting warnings. It turns out that there are some Tribal areas with a "none" geography type. I'm updating them now to fix! 

I'm also more aggressively restricting the filter so it doesn't suppress other helpful warnings like this in the future.

Nice catch @mattbowen-usds !!

lucasmbrown-usds: @mattbowen-usds -- I honestly have no idea what's going on here. I pushed some new code that removes both the Points and the `None` from tribal_gdf. 

But, the warning is still there, even though it should be polygon/multipolygon to polygon/multipolygon, which should be fine.  

Further, it says `348 dropped geometries` -- which is more than the rows in `tribal_gdf_without_points`! 

If this is urgent to get in Monday, do you mind investigating a little bit while I'm out?

mattbowen-usds: From my commit

> I looked into this a bit, and in general the geometry type mismatch
> changes very little about the calculation; we have a mix of
> multipolygons and polygons. The fastest thing to do is just not keep
> geom type; I did some runs with it set to both True and False, and
> they're the same within 9 digits of precision. Logically we just want to
> overlaps, regardless of how the actual geometries are encoded between
> the frames, so we can in this case ignore the geom types and feel OKAY.


I changed it to ignore the geometry types because that is what makes sense to me after really digging into it, and also it basically doesn't matter; whatever was dropping before is a TINY (< .00000001%) amount of actual area.

data/data-pipeline/README.md
```diff
@@ -234,6 +234,15 @@ If you want to run tile generation, please install TippeCanoe [following these i
 - We use [Poetry](https://python-poetry.org/) for managing dependencies and building the application. Please follow the instructions on their site to download.
 - Install Poetry requirements with `poetry install`
 
+### Running tox 
+
+Our full test and check suite is run using tox. This can be run using commands such 
+as `poetry run tox`.
+
+Each run can take a while to build the whole environment. If you'd like to save time,
+you can use the previously built environment by running `poetry run tox -e lint`
```
esfoobar-usds, line 243: I didn't know this one!

data/data-pipeline/data_pipeline/etl/sources/geo_utils.py
```diff
@@ -18,38 +19,66 @@ def get_tract_geojson(
...
+        GEOJSON_PATH,
+        include_fields=["GEOID10"],
+    )
+    tract_data = tract_data.rename(
+        columns={"GEOID10": "GEOID10_TRACT"}, errors="raise"
+    )
     return tract_data
 
 
+@lru_cache()
```
esfoobar-usds, line 38: I am curious about the usage of `lru_cache` here. Is this function called multiple times?

data/data-pipeline/data_pipeline/etl/sources/tribal/etl.py
```diff
@@ -59,7 +60,10 @@ def _transform_bia_national_lar(self, tribal_geojson_path: Path) -> None:
         )
 
         bia_national_lar_df.rename(
-            columns={"LARID": "tribalId", "LARName": "landAreaName"},
+            columns={
+                "LARID": field_names.TRIBAL_ID,
+                "LARName": field_names.TRIBAL_LAND_AREA_NAME,
```
esfoobar-usds, line 65: Nice catch!

data/data-pipeline/data_pipeline/etl/sources/tribal/etl.py
```diff
@@ -87,7 +91,10 @@ def _transform_bia_aian_supplemental(
         )
 
         bia_aian_supplemental_df.rename(
-            columns={"OBJECTID": "tribalId", "Land_Area_": "landAreaName"},
+            columns={
+                "OBJECTID": field_names.TRIBAL_ID,
+                "Land_Area_": field_names.TRIBAL_LAND_AREA_NAME,
```
esfoobar-usds, line 96: Ditto!

data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py
```diff
@@ -0,0 +1,226 @@
...
+        "17",
+        "18",
+        "21",
+        "24",
+        "33",
+        "34",
+        "39",
+        "50",
+        "51",
+        "54",
```
esfoobar-usds, line 39: Can we document why this is?

esfoobar-usds (approved):
🍏 with some small comments

data/data-pipeline/data_pipeline/etl/sources/geo_utils.py
mattbowen-usds: I am guessing he was following the convention `get_tract_geojson`, and i was like "well, it won't hurt, since it is expensive. I believe at this time it isn't, so I'm good to drop the cache untel we need ot/

data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py
mattbowen-usds: Lol I honestly have no idea either, so I'll give that to @lucasmbrown-usds --- my best guess is those states have no tribal lands in them.

data/data-pipeline/README.md
lucasmbrown-usds: That's a @sverchdotgov discovery!

data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py
lucasmbrown-usds: Yes exactly. I'll add a comment though, awesome point @esfoobar-usds that we should be commenting these.

data/data-pipeline/data_pipeline/etl/sources/geo_utils.py
mattbowen-usds: It is actually useful on get_tract_geojson, so I vote keep it. And I'm about to merge so it's stayin'

PR #1916, opened 2022-09-22
Issue 1803: percent of zips (WIP)

lucasmbrown-usds:
Work in progress.


emma-nechamkin:
QQ here for @lucasmbrown-usds --> are you sure we need to be using ZIP rather than ZCTA? Zip and ZCTA are usually the same, and the census bureau publishes a tract:zcta crosswalk (https://www.census.gov/geographies/reference-files/time-series/geo/relationship-files.html). Just wondering!

vim-usds:
upshot: what % of a zip code is disadv? Some places do have data on zip codes. 

Random example:
If there's $1M to zip code X and if 75% are disadv., therefore 750k went to disadvantaged communities.

PR #1923, opened 2022-09-23
Backfill population in island areas (#1882)

### LINKED ISSUE #1882: Use 2010 Decennial Census population data for island areas AS, MP, GU, and VI instead of showing as N/A in new "Tract Information" side panel
**Description**
In the new side panel as seen using the latest staging links, when navigating to any tract in island areas American Samoa, Northern Mariana Islands, Guam, or Virgin Islands, selecting the tract to see the side panel including the new "Tract Information" with Population, it's always N/A. Per @emma-nechamkin this is because we're getting population from ACS. We could pull population at least for the island areas from 2010 Decennial Census so that we can show something for population numbers here instead of N/A.

update - for these island areas we're already using the 2010 Decennial Census data, for instance for American Samoa we're already showing Low median income, Unemployment, Poverty, and High school degree non-attainment (screenshot below)

**Screenshots**
image

image
 
image

mattbowen-usds:
Closes #1882


esfoobar-usds (approved):
🍏

mattbowen-usds:
This is now part of #1932

PR #1935, opened 2022-09-28
No issue number: Update front-end Readme

lucasmbrown-usds:


vim-usds (commented):
This is great!

I'm wondering if we can update the docs to include using both `brew` and `nvm` to install node? I've run into issues in the past with brew (maybe these are no longer an issue). See below for more info:

https://helpmonks.com/blog/switching-brew-nvm-node-install/

https://stackoverflow.com/questions/28017374/what-is-the-recommended-way-to-install-node-js-nvm-and-npm-on-macos-x

wdyt?

vim-usds:
Also it seems that the FE build failed to Cypress. The longest PR in history is open to address these flaky tests. Maybe try re-running this. Also what version of node are you using? We have a `.nvmrc` file that will auto set this (if you are using `nvm` and not `brew`). I guess another reason for `nvm`.

vim-usds:
@lucasmbrown-usds - congrats on getting the FE up and running

lucasmbrown-usds:
Sorry for the slow response to your comments, @vim-usds ! 

I'm happy with having two different versions of the README section on installing node -- one for brew and one for nvm. I switched from `brew` to `nvm` because the rest of that section assumes `nvm` (e.g., line 37 is `nvm use 14`). 

I don't know how to do a full install using brew. How about this: we merge this for now so the instructions are complete for `nvm`, and if you want to add a section on `brew` then you can? 

On your second question, I followed the instructions to `nvm use 14` and then node -v gives me `v14.20.1`. Is there anyway to get node to respect the client nvm version?

vim-usds (approved):

PR #1936, opened 2022-09-28
Remove markdown link from markdown checker

vim-usds:
This broken markdown link is causing the FE release branch to fail. Hoping to get this merged into main so that the FE release branch doesn't have to build the BE for each FE fix


lucasmbrown-usds (approved):
LGTM!

Issue #1921, opened 2022-09-23
New data fields in download files need to be in a different order in the file


BethMattern:
Attached is an xls file and all the columns highlighted in yellow are newly added columns that need to be moved to the position they are in in the attached. 

Restated, the order of the columns for the download files should be in the attached order. Those that need to be changed are in yellow.

Note - lack of plumbing is missing from the attached - another bug was logged for this issue.

BethMattern:
[https://app.zenhub.com/files/358242767/2e36696f-a1e3-4eef-92f0-de9023540927/download](https://app.zenhub.com/files/358242767/2e36696f-a1e3-4eef-92f0-de9023540927/download)

PR #1940, opened 2022-09-28
Re-add lack of plumbing fields

### LINKED ISSUE #1920: Lack of plumbing data fields are missing from the .xls and .csv
The lack of plumbing data fields are missing from the download files. They should be added and they should be added after the column called "Median value ($) of owner-occupied housing units"

mattbowen-usds:
Closes #1920

### DIFF
--- a/data/data-pipeline/data_pipeline/content/config/excel.yml
+++ b/data/data-pipeline/data_pipeline/content/config/excel.yml
@@ -165,6 +165,12 @@ sheets:
       - score_name: Median value ($) of owner-occupied housing units
         label: Median value ($) of owner-occupied housing units
         format: float
+      - score_name: Share of homes with no kitchen or indoor plumbing (percentile)
+        label: Share of homes with no kitchen or indoor plumbing (percentile)
+        format: float
+      - score_name: Share of homes with no kitchen or indoor plumbing (percent)
+        label: Share of homes with no kitchen or indoor plumbing (percent)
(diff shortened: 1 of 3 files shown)


mattbowen-usds:
I need to update the pickles it looks like, so switching this to draft

data/data-pipeline/data_pipeline/content/config/excel.yml
```diff
@@ -165,6 +165,12 @@ sheets:
       - score_name: Median value ($) of owner-occupied housing units
         label: Median value ($) of owner-occupied housing units
         format: float
+      - score_name: Share of homes with no kitchen or indoor plumbing (percentile)
```
lucasmbrown-usds, line 168: This is the opposite order of the other file -- is that intentional?

lucasmbrown-usds (changes requested):

data/data-pipeline/data_pipeline/content/config/excel.yml
mattbowen-usds: Good catch! It isn't. I also had to make a bunch of field order changes that are gonna end up causing conflicts, so will fix, but in #1921

mattbowen-usds:
I ended up just moving this into #1921 so that we can see where this column is with all the other correctly moved columns. Closing here and will open a PR for #1921 shortly.

PR #1942, opened 2022-09-28
Reorder download fields, add plumbing back

### LINKED ISSUE #1921: New data fields in download files need to be in a different order in the file

mattbowen-usds:
Closes #1921 and #1920


lucasmbrown-usds (approved):
LGTM! Will definitely be good to have Kameron or Beth spot-check the download files when they come out.

mattbowen-usds:
@BethMattern confirmed field order is correct. Merging!

PR #1960, opened 2022-09-29
Issue 1944: Changing source of Tribal data file

lucasmbrown-usds:
The old file had 24 missing geometries from the National LAR, for some reason. 

I uploaded two new zips to s3, refactored the ETL to point to them, and the problem is solved.


mattbowen-usds (approved):
Looks good! Nice, quick work.

esfoobar-usds (approved):
🍏 very cool!

PR #1976, opened 2022-10-04
Adding ADJLI_ET

emma-nechamkin:


lucasmbrown-usds (approved):
lgtm

vim-usds:
@emma-nechamkin - will this signal exist on all tiles? And if so, will it always have either a value of True or False?

emma-nechamkin:
> @emma-nechamkin - will this signal exist on all tiles? And if so, will it always have either a value of True or False?

i think yes?

PR #1993, opened 2022-10-06
Issue 1992: Do not impute income for null population tracts 

lucasmbrown-usds:
#1992. 

Weirdly, this doesn't seem to have affected any tracts? Because we already has this as a smoketest that was passing: 

```
# Make sure that any tracts with null population have null imputed income
    tracts_with_null_population_df = final_score_df[
        final_score_df[field_names.TOTAL_POP_FIELD].isnull()
    ]
    assert (
        tracts_with_null_population_df[
            field_names.POVERTY_LESS_THAN_200_FPL_IMPUTED_FIELD
        ]
        .isna()
        .all()
    )
```

So maybe the `null` population tracts got cast to `0` population somewhere in the scoring process?


emma-nechamkin (approved):
LGTM

PR #2004, opened 2022-10-12
Make tribal overlap set score N

### LINKED ISSUE #1998: Update score to account for 100% tribal overlap tracts (aka the great October 7th compromise of 2022)
1. New field called "Identified as disadvantaged due to tribal overlap" needs to be added. It will be True if the census tract overlaps with tribal boundaries by 99.5% or higher. This does not apply to the Alaska Native Villages.

2. If the "Identified as disadvantaged due to tribal overlap" field is True, then the "Definition N community, including adjacency index tracts" boolean field should be True.

3. There will be a new field called "Percentage of tract that is disadvantaged" and this will be 100% if "Definition N community, including adjacency index tracts" = True, will be 100% if the tract is 99.5% or higher inside tribal areas, and will be the partial % overlap number if the tract is partially inside tribal areas, and will be 0% of the tract is neither a DAC nor inside any tribal areas.

mattbowen-usds:
Closes #1998


data/data-pipeline/data_pipeline/content/config/csv.yml
```diff
@@ -59,9 +59,15 @@ fields:
 - score_name: Definition N (communities) (based on adjacency index and low income alone)
   label: Identified as disadvantaged based on neighbors and relaxed low income threshold only
   format: bool
+- score_name: Identified as disadvantaged due to tribal overlap
+  label: Identified as disadvantaged due to tribal overlap
+  format: bool
 - score_name: Definition N community, including adjacency index tracts
   label: Identified as disadvantaged
   format: bool
+- score_name: Percentage of tract that is disadvantaged
```
emma-nechamkin, line 68: can we make clear this is area-based?

data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py
```diff
@@ -72,16 +74,17 @@ def _create_string_from_list(series: pd.Series) -> str:
...
-    @staticmethod
+    @classmethod
     def _adjust_percentage_for_frontend(
+        cls,
         percentage_float: float,
     ) -> Optional[float]:
         """Round numbers very close to 0 to 0 and very close to 1 to 1 for display"""
         if percentage_float is None:
             return None
         if percentage_float < 0.01:
```
emma-nechamkin, line 85: should we have the same cutoff type here? like 0.005 in the vein of 0.995? There's a little bit of asymmetry?

data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py
```diff
@@ -246,6 +249,12 @@ def transform(self) -> None:
             field_names.COUNT_OF_TRIBAL_AREAS_IN_TRACT_CONUS
         ] = None
 
+        merged_output_df[field_names.IS_TRIBAL_DAC] = np.where(
```
emma-nechamkin, line 252: I think it might be a bit cleaner to do this with a fillna potentially?

````
merged_output_df[field_names.IS_TRIBAL_DAC] = merged_output_df[field_names.PERCENT_OF_TRIBAL_AREA_IN_TRACT]
            > self.TRIBAL_OVERLAP_CUTOFF
```

data/data-pipeline/data_pipeline/score/score_narwhal.py
```diff
@@ -997,6 +997,33 @@ def _mark_donut_hole_tracts(self) -> pd.DataFrame:
             ]
         )
 
+    def _mark_tribal_dacs(self) -> None:
+        """Per the October 7th compromise (#1988),
+        tracts that are approx 100% tribal are Score N communities.
+        """
+        self.df[field_names.SCORE_N_COMMUNITIES] = np.where(
```
emma-nechamkin, line 1004: might also be easier / make a bit more sense to just take the MAX of those two columns? and why overwrite SCORE_N_COMMUNITIES?

data/data-pipeline/data_pipeline/content/config/csv.yml
mattbowen-usds: @emma-nechamkin I asked Beth in the ticket; it's good by me, but I leave the actual words to Beth and others

data/data-pipeline/data_pipeline/score/score_narwhal.py
```diff
@@ -997,6 +997,33 @@ def _mark_donut_hole_tracts(self) -> pd.DataFrame:
...
+        tracts can be partially DACs if some portion of the tract is tribal land.
+
+        Rules are as follows:
+        If a tract is a SCORE_N_COMMUNITY, it is 100% a DAC
+        If a tract is not, but contains tribal land, the percent that is tribal land is a DAC.
+        """
+        tribal_percent = self.df[
+            field_names.PERCENT_OF_TRIBAL_AREA_IN_TRACT
+        ].fillna(0.0)
+        return np.where(
```
emma-nechamkin, line 1021: Why not just...

```
return self.df[[field_names.PERCENT_OF_TRIBAL_AREA_IN_TRACT, field_names.FINAL_SCORE_N_BOOLEAN]].max(axis=1)
````

File: data/data-pipeline/data_pipeline/score/score_narwhal.py [hunk 6]
```diff
@@ -1031,10 +1058,15 @@ def add_columns(self) -> pd.DataFrame:
         ]
         self.df[field_names.CATEGORY_COUNT] = self.df[factors].sum(axis=1)
         self.df[field_names.SCORE_N_COMMUNITIES] = self.df[factors].any(axis=1)
+        self._mark_tribal_dacs()
```
emma-nechamkin, line 1061: this is nice and clean

data/data-pipeline/data_pipeline/score/score_narwhal.py
```diff
@@ -1031,10 +1058,15 @@ def add_columns(self) -> pd.DataFrame:
...
         self.df[
             field_names.SCORE_N_COMMUNITIES
             + field_names.PERCENTILE_FIELD_SUFFIX
         ] = self.df[field_names.SCORE_N_COMMUNITIES].astype(int)
+
         self._mark_donut_hole_tracts()
+        self.df[
```
emma-nechamkin, line 1068: a quick style comment -- why not modify df in place here as per the mark functions?

emma-nechamkin: this is a question NOT a suggestion to be clear

emma-nechamkin (approved):
this looks really fantastic, just a few comments. i have NOT pulled the code down to rerun, but I think nothing here looks off and the 295 number buoys me

data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py
mattbowen-usds: The decisions were made at different times, so good catch. @BethMattern thoughts on changing the cutoff for "very close to 0%" smaller (.5% instead of 1%)?

data/data-pipeline/data_pipeline/score/score_narwhal.py
mattbowen-usds: I think because I noticed a pattern of marking tracts by mutating in place (e.g., with the donut hole calculation) but getting scores and things like that and assigning them (e.g., `self.df[field_names.N_CLIMATE] = self._climate_factor()` etc) so I was following that in this code.

data/data-pipeline/data_pipeline/score/score_narwhal.py
emma-nechamkin: totally valid!

data/data-pipeline/data_pipeline/score/score_narwhal.py
emma-nechamkin: no not a logical mistake at all! this makes sense to me! i am willy nilly with types

esfoobar-usds (approved):
🍏

data/data-pipeline/data_pipeline/etl/sources/tribal_overlap/etl.py
mattbowen-usds: If we change this, it'll have both a backend and frontend change, and it's not obvious to me we should, so I've pulled it into its own ticket to explicitly make the decision: #2006

Issue #1505, opened 2022-03-29
Add test for codebook to generate_score_post


emma-nechamkin:
I always think we need tests, obviously! Would those be easy to add? Or you can create a backlog ticket for it, but as mentioned on the other PR, tickets for tests alone can be a real drag to add at a later date. 

Snapshot of the whole codebook makes sense to me! And would be v easy to update with our snapshot code.

As for logging, regardless of how long it takes to run they can be helpful on a remote server for pinpointing how far a process got before it crashed, and can be helpful to someone reading that output to know exactly what is generally happening under the hood during a process even when it's successful.

_Originally posted by @lucasmbrown-usds in https://github.com/usds/justice40-tool/pull/1469#discussion_r837952396_

emma-nechamkin:
We can do this as a series of assertions in the score_geo code :) we chatted 8/5

PR #2014, opened 2022-10-13
Add assertions around codebook

mattbowen-usds:
#1505


emma-nechamkin (approved):
This looks fine to me. I'm honestly kind of confused by what we were originally planning here!

Issue #2020, opened 2022-10-17
Create an accurate 1:1 correspondence between BE signals and the FE


vim-usds:
Currently there a number of signals that are mis-matched between the FE and the BE. Cases being the BE sends them but they are not being used by the FE or vice-versa. 

Lets understand which BE signals are needed for the FE and remove all the one that are not needed.

mattbowen-usds:
Both PRs are closed!

PR #2022, opened 2022-10-19
Remove unused signals

mattbowen-usds:
Part of #2020


data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -240,14 +240,10 @@
     + field_names.PERCENTILE_FIELD_SUFFIX: "LIF_PFS",
     field_names.LOW_MEDIAN_INCOME_AS_PERCENT_OF_AMI_FIELD
     + field_names.PERCENTILE_FIELD_SUFFIX: "LMI_PFS",
-    field_names.MEDIAN_HOUSE_VALUE_FIELD
```
mattbowen-usds, line 243: @emma-nechamkin I would love your eyes on the changed I made to this file --- are any of these surprising to you? From what I could tell, none of these vars are used by the frontend, but if there are any that you're like "wait the frontend **really** should be using that, what is going on?!" it would be helpful for @vim-usds and me to know.

emma-nechamkin: yeah so this is fine -- we only use this for lead and i think it's abstracted!

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -240,14 +240,10 @@
...
-    + field_names.PERCENTILE_FIELD_SUFFIX: "MHVF_PFS",
     field_names.PM25_FIELD + field_names.PERCENTILE_FIELD_SUFFIX: "PM25F_PFS",
     field_names.HIGH_SCHOOL_ED_FIELD: "HSEF",
     field_names.POVERTY_LESS_THAN_100_FPL_FIELD
     + field_names.PERCENTILE_FIELD_SUFFIX: "P100_PFS",
-    field_names.POVERTY_LESS_THAN_200_FPL_FIELD
```
emma-nechamkin, line 249: this is the raw value; i think this is okay to remove

data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -276,30 +272,9 @@
     # temporarily update this so that it's the Narwhal score that gets visualized on the map
     # The NEW final score value INCLUDES the adjacency index.
     field_names.FINAL_SCORE_N_BOOLEAN: "SN_C",
-    field_names.SCORE_N_COMMUNITIES
```
emma-nechamkin, line 279: I am so behind what we decided to do on the front end so i cannot give you a good appraisal of this!

emma-nechamkin (approved):
i am really behind on front end stuff, but generally the things i know are reflected!

PR #2031, opened 2022-10-21
Disable file size limits on tiles

mattbowen-usds:
After initial analysis, we realized that tracts appear and disappear between zoom levels 5-7 because tippecanoe limits vector tiles to 500kb by default, and we were telling it to drop features to stay under that 500kb. For about 20% of tiles at zoom level 5, we can't stay easily under 500kb. To restore map functionality, we've removed these caps, and follow-on work will reduce the tile sizes as much as practical.


vim-usds (approved):
LGTM!

vim-usds:
Just tested this branch with the FE release branch and it's looking like the problem is solved!

https://screeningtool-staging.geoplatform.gov/1829-451fda/en/?flags=stage_hash=2031/842a507e543ba4a34ff33441097f9ccab5975edb#3/33.47/-97.5

On slower connections, this is still TBD, however I think we might be OK to move forward and wait until someone complains about our map being slow

PR #2039, opened 2022-10-26
Add action for creating new score versions

mattbowen-usds:
Add a new action that allows us to manually create and upload score versions to s3.


vim-usds (approved):
LGTM

PR #2040, opened 2022-10-26
Round ALL the float fields for the tiles

### LINKED ISSUE #2033: Round all floats for frontend
While trying to understand what drives the size of our pbf tiles, we determined that adding an unrounded float that exists in most tracts increases file sizes for zoom level 5 by about 20% over the minimum file size possible.

We looked, and all of the float fields in the tiles are acceptable to round to two digits of accuracy, so make that change.

mattbowen-usds:
Closes #2033


data/data-pipeline/data_pipeline/etl/score/constants.py
```diff
@@ -373,72 +373,3 @@
     field_names.PERCENT_OF_TRIBAL_AREA_IN_TRACT: "TA_PERC",
     field_names.PERCENT_OF_TRIBAL_AREA_IN_TRACT_DISPLAY: "TA_PERC_FE",
 }
-
-# columns to round floats to 2 decimals
-# TODO refactor to use much smaller subset of fields we DON'T want to round
```
emma-nechamkin, line 378: lol

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
```diff
@@ -231,10 +231,12 @@ def _create_tile_data(
...
-
-        score_tiles[constants.TILES_SCORE_FLOAT_COLUMNS] = score_tiles[
-            constants.TILES_SCORE_FLOAT_COLUMNS
-        ].apply(
+        float_cols = [
+            col
+            for col, col_dtype in score_tiles.dtypes.items()
+            if col_dtype == np.dtype("float64")
+        ]
+        score_tiles[float_cols] = score_tiles[float_cols].apply(
```
emma-nechamkin, line 239: i think since we are FLOORING everything to two decimal places we no longer need to floor it in this way but this is correct otherwise. 

A question -- why use `==`vs  `is` for the comparison (earnest question, not a gotcha)

emma-nechamkin (approved):
this is so much cleaner.

data/data-pipeline/data_pipeline/etl/score/etl_score_post.py
mattbowen-usds: Honestly because I have a heuristic for using "is" on singletons like None, True, False and my heuristic doesn't extend to dtypes :)

What way would you floor things now that we're flooring everything to two decimal places?

emma-nechamkin: i don't understand here why we use a separate floor series rather than applying `floor`, a standard function?

PR #2044, opened 2022-10-27
Change geojson to shapefile

### LINKED ISSUE #1931: There is no geojson with the Shapefiles download
**Describe the bug**
Need to update the copy on the downloads page to not say that the Shapefiles download includes geojson files.

there’s no geojson in that download — it’s just a codebook and the shapefiles

**To Reproduce**
1. Go to 'downloads'
2. Scroll down to 'Shapefiles download'
3. See error

**Expected behavior**
Update the copy (see kameron if you have questions)

**Screenshots**
Screen Shot 2022-09-27 at 1 04 44 PM

**Additional context**

mattbowen-usds:
Closes #1931

### DIFF
--- a/client/src/data/copy/downloads.tsx
+++ b/client/src/data/copy/downloads.tsx
@@ -504,7 +504,7 @@ export const DOWNLOAD_LINKS = {
   />,
   LINK3: <FormattedMessage
     id={'download.page.download.file.3'}
-    defaultMessage={`<link3>Shapefile</link3> (Codebook included with geojson {shapeFileSize} unzipped)`}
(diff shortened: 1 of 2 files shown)


vim-usds:
@mattbowen-usds - Awesome work!!!

What a pleasant surprise to see a FE PR! 

Quick question, I downloaded the link, and it wasn't obvious to me what a "codebook" looks like. The downloaded zip does have the shapefiles. Is the codebook called "columns.csv"?

mattbowen-usds:
@vim-usds  It is! To be honest, I'm not super familiar with that world either, but columns.csv is what I think of as the codebook --- it tells you what each of the short names actually expands to in human.

PR #2045, opened 2022-10-27
Update the dockerfile

### LINKED ISSUE #2026: Update the dockerfile
Our readme describes a dockerized version of running the pipeline. No one has really checked it against the narwhal changes, so, try it out with Narwhal and fix whatever is broken.

mattbowen-usds:
Closes #2026


vim-usds (approved):
Is the timezone requirement something new?

mattbowen-usds:
@vim-usds the builds were failing on that for me, and I found a lot of talk about it in that version of ubuntu, so I think so

PR #2052, opened 2022-10-28
Add checks to create version 

### LINKED ISSUE #2047: Add additional checks to score generation action
Add some last smoketests to the score generation action so it can be run without much fear:

1. For beta and 1.0 runs, make sure you actually generated the score you meant to, failing if you used a 1.0 branch but asked for beta or vice versa
2. For 1.0 runs, run the smoketests before uploading the score files
3. Ensure the files we link to are uploaded and publicly accessible at the end of the run; if not, fail

mattbowen-usds:
Closes #2047

### DIFF
--- a/.github/workflows/create-score-version.yml
+++ b/.github/workflows/create-score-version.yml
@@ -5,11 +5,12 @@ on:
       score_version:
         description: "Which version of the score are you generating?"
         required: true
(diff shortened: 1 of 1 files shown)


vim-usds (approved):
Awesome work!

PR #2114, opened 2022-12-01
Fix failing safety

mattbowen-usds:
https://nvd.nist.gov/vuln/detail/CVE-2022-42969 landed recently and there's no fix in py (which is maintenance mode). From my analysis, that CVE cannot hurt us (famous last words), so we'll ignore the vuln for now.

### DIFF
--- a/.github/workflows/data-checks.yml
+++ b/.github/workflows/data-checks.yml
@@ -22,9 +22,11 @@ jobs:
       # one execution of the tests per version listed above
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Upgrade wheel
+      - run: pip install -U wheel
       - name: Print variables to help debug
         uses: hmarr/debug-action@v2
       - name: Load cached Poetry installation
--- a/.github/workflows/data-checks.yml
+++ b/.github/workflows/data-checks.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Upgrade wheel
-      - run: pip install -U wheel
+        run: pip install -U wheel
       - name: Print variables to help debug
         uses: hmarr/debug-action@v2
       - name: Load cached Poetry installation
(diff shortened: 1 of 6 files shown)


.github/workflows/data-checks.yml
```diff
@@ -22,9 +22,11 @@ jobs:
       # one execution of the tests per version listed above
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Upgrade wheel
```
vim-usds, line 28: Seems like wheel is an "[optimization](https://pip.pypa.io/en/stable/cli/pip_wheel/)"

Seems safe to remove

data/data-pipeline/tox.ini
```diff
@@ -15,7 +15,8 @@ commands = black data_pipeline
 [testenv:checkdeps]
 # checks the dependencies for security vulnerabilities and open source licenses
 allowlist_externals = bash
-commands = safety check --ignore 44715  # known issue: https://github.com/pyupio/safety/issues/364
+commands = pip install -U wheel
```
vim-usds, line 18: Does this mean that we removed it from GHA and placed wheel into tox instead?

vim-usds (approved):
Looking great!

vim-usds:
@mattbowen-usds - also it seems that the map is loading nicely!

https://screeningtool.geoplatform.gov/en?flags=stage_hash=2114/4d9a8a574963aea16429e666d72450b57184076b#3/33.47/-97.5

This is the latest staging link. 

I'm investigating why there are so many comments labeled "** Map Deployed! **", I'm guessing one for every failure? Seems like it shoudn't have deployed the map if it failed. 🤔

.github/workflows/data-checks.yml
mattbowen-usds: This is not obvious if you're not a pip person, but `install -U` upgrades, not removes. Since basically everything depends on wheel somewhere now, I'm just making sure we have the most current one.

data/data-pipeline/tox.ini
mattbowen-usds: Tox sets up its own separate virtualenv from the main one, so I neded to replicate upgrading wheel for tox runs the way I did in the general action.

mattbowen-usds:
@vim-usds that is new! and i don't understand either!

mattbowen-usds:
@vim-usds my best guess as to why we got so many deploys was I  was pushing commits pretty quickly, and so some of the backend runs could have ended up still running; they wouldn't have had much reason to fail on their own, if the network didn't let them down, and maybe github just failed to kill them since I kept pushing new commits.

Issue #1531, opened 2022-04-05
Error using quickstart guide/docker


ejmiddleton:
**Describe the bug**
Error running docker via the quickstart guide instructions:
https://github.com/usds/justice40-tool/blob/main/QUICKSTART.md

**To Reproduce**
1. I cloned the git.
2. I opened a command prompt and changed to this directory: $ cd justice40-tool
3. I installed docker.
4. I typed $ docker-compose up

**Expected behavior**
All packages to be correctly installed.

**Desktop (please complete the following information):**
 - OS: Windows 10

[docker_error.txt](https://github.com/usds/justice40-tool/files/8419489/docker_error.txt)

PR #2137, opened 2023-01-23
Ianvnorrisusds/1531 fix docker quickstart

IanNorrisUSDS:
Addresses #1531 

Describe the bug

Was attempting to follow the quick start guide and noticed this issue which is a similar problem in terms of it being "quick" for folks unfamiliar to docker (needing to know to run docker-compose build before up).

Also modified the dockerfile to get the data server running, and noticed an annoying thing with port 5000 (future todo: change it away from port 5000 or make configurable?) so added that to the readme as well.


vim-usds (approved):
Awesome Ian!

vim-usds:
@IanNorrisUSDS, are you seeing any of these pipeline checks failing locally?

PR #2142, opened 2023-02-01
Update dependencies to fix safety check failures

travis-newby:
Update dependencies causing safety check to fail. There were two dependencies updated:

1. Pillow was updated to a version that is not [susceptible to DoS](https://pyup.io/v/51886/f17) (9.3.0). Pillow was previously pinned, so I left it as a pinned dependency; there are _upcoming_ breaking changes specified in the release notes, so this seems like a safe choice. There are some new deprecations, but I didn't see any _current_ breaking changes in the release notes. 
2. Certifi was updated to a version that [removes TrustCor certificates](https://pyup.io/v/52365/f17/) from the root store. (>= 2022.12.07).

During this process, setuptools was upgraded automatically. That takes care of a potential [DoS issue](https://pyup.io/v/52495/f17).

Finally, I removed the nb-black dev dependency. It's not being used, and it doesn't play nice with poetry on certain Mac devices.


travis-newby:
I'd like to do a bit more testing before merging, but I first wanted to see if this resolved issues we saw in the checks.

data/data-pipeline/data_pipeline/ipython/compare_tiles_and_geoJson_files.ipynb
```diff
@@ -332,7 +331,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3.8.10 ('data-pipeline-WziHKidv-py3.8')",
+   "display_name": "Python 3 (ipykernel)",
```
vim-usds, line 334: I'm a noob with notebooks, what is the significance of the display name and do we know when it's used?

vim-usds: Any reason to not include the semantic versioning?

lucasmbrown-usds (approved):
LGTM!

travis-newby:
Spot checked around the map. Everything looks good. 🚀

PR #2147, opened 2023-02-03
fix: update ua-parser

vim-usds:
This will update a transitive dependency to mitigate a security vulnerability flagged by dependabot.

For those that are able to do a PR review:

This will require a quick test on mobile. Warning: the mobile experience is far from great.  This test is mainly to increase confidence that regressions weren't introduced.

The direct dependency that was updated is used for mobile detection. Which is being used to render the map on mobile and also to handle the case when the open source community member isn't using Mapbox (paid). 

- [ ] Given the staging link, see if the map renders and operates in the same manner as production. 

Regressions would be obvious like
1. map doesn't render
2. map is not clickable

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -16781,11 +16781,11 @@
       }
     },
     "react-device-detect": {
-      "version": "1.17.0",
-      "resolved": "https://registry.npmjs.org/react-device-detect/-/react-device-detect-1.17.0.tgz",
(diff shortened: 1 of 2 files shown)


vim-usds:
QA using staging link on iPhone seems to work well 
- [ ] map renders
- [ ] map is clickable

travis-newby (approved):
I checked on macOS Safari and mobile Safari. Map renders and is clickable / tappable. There are no extra JS errors. So LGTM!

We should consider filing bugs to clean up the existing JS errors.

PR #2151, opened 2023-02-04
Add pre-cleaning to a couple of zip files during geo-score

travis-newby:
Two zip files were not getting cleared prior to running geo score; one of those was also not getting cleared before the full run. This resulted in them growing between runs until the application ground to a halt. This fix deletes those two zip files.


travis-newby:
We should have a larger conversation around pre- and post-run cleanup. We have multiple ETLs writing to the same directory, or a single ETL writing to multiple directories, and it gets a little wonky (this is a technical term).

vim-usds (approved):
This looks good. I haven't run it locally. Let's chat if a new score needs to be pushed to S3

PR #2155, opened 2023-02-07
User Story 2152 – Clean up logging

travis-newby:
This is a high touch change to clean up logging across the data pipeline. This change,

0. Fixes a bug that was causing all log messages to be displayed twice
1. Streamlines the number of INFO log messages displayed during typical data pipeline runs
2. Categories the logging level of every log message in the data pipeline, so that we can better use the ability to only show certain log levels
3. Removes extra log messages that were not adding value to any level of logging


travis-newby:
Since this is such a high touch change, I took care to ensure the pipeline was not changed. However, I can't test my last few changes locally due to being over the census API limit. Hopefully the build portion of this PR will help illuminate any issues.

travis-newby:
Spot-checked the map for any changes. Seems GTG. 🚀

data/data-pipeline/data_pipeline/etl/score/etl_score.py
```diff
@@ -56,8 +56,6 @@ def __init__(self):
         self.ISLAND_DEMOGRAPHIC_BACKFILL_FIELDS: List[str] = []
 
     def extract(self) -> None:
-        logger.info("Loading data sets from disk.")
```
mattbowen-usds, line 59: For a second I was like "wait but this is true," but then I remembered you'd instead put these centrally as debug in `_run_one_dataset`, which makes a ton more sense. This is a not for a future reviewer who also goes "wait!?"

data/data-pipeline/data_pipeline/etl/sources/census/etl.py
```diff
@@ -211,31 +213,35 @@ def _load_national_geojson(self):
...
+        logger.debug("Loading census data")
+
+        logger.debug("Loading individual state csv files")
         for fips_code in self.TRACT_PER_STATE:
             self._load_into_state_csvs(fips_code)
+
         self._load_national_csv()
         self._load_national_geojson()
+
+        logger.debug("Census data complete")
```
mattbowen-usds, line 247: Does your goodbye not handle this?

mattbowen-usds (approved):
I feel like I should have more to say about a PR that touches 63 files, but I agree with pretty much every change. Nicely done.

data/data-pipeline/data_pipeline/etl/score/etl_score.py
travis-newby: Yep, I tried to delete these if they didn't add value.

data/data-pipeline/data_pipeline/etl/sources/census/etl.py
travis-newby: Because there are multiple ways the census etl gets run (by itself, or as part of the full run but separate from the other etls) I left that particular debug statement in. I thought it may help in some rando circumstance I've not thought of yet. Happy to take it out if you don't think it'll add any value.

vim-usds (approved):

data/data-pipeline/data_pipeline/etl/sources/census/etl.py
mattbowen-usds: nah no strong feeling, and that makes sense. i'd say keep it.

Issue #2168, opened 2023-02-15
ADR: Add ability to cache the data pipeline's external data sources


travis-newby:
# Speed data pipeline time and conserve network resources by caching external data sources

## Context and Problem Statement
As designed, the data pipeline downloads and unpacks all external data sources with every pipeline run. This is because the original developers believed that the data pipeline's data sources, and subsequently the score, would change often. However, that's not the case. And in fact the opposite is true: there is a desire for the score to remain stable until new external data sources are vetted and incorporated into the product.

However, the code contains no mechanism to cache and reuse external data sources.

## Decision Drivers 

- The downloaded data sources are approximately 5.6GB in size, and are downloaded with each pipeline run
- A non trivial amount of time is spent downloading those data sources with each run of the data pipeline, adversely impacting developer productivity
- Some data sources, such as the census bureau, throttle the number of requests over a time period; running the pipeline multiple times per day can get developers banned from making calls for 24 hours
- Because data sources may change between runs, it's impossible to know if a code change impacted the score or if a data source changing impacted the score

## Considered Options

- Maintaining the status quo
- Aggressively caching data sources by default with each run
- Caching data sources in a well-known location inside the project when a flag is sent to application.py

## Decision Outcome

We picked "Caching data sources in a well-known location inside the project when a flag is sent to application.py", because this option allows us to speed up our pipeline, reduce the required bandwidth, and maintain consistency between runs. By requiring a flag to turn on caching, this option has the aforementioned benefits without fundamentally changing the way the application works. We therefore won't have to update any GitHub Actions or change the way the build works; it does, however, give us the freedom to take advantage of the cache more broadly by adding the data sources to source control in the future (should we choose to do so).

PR #2169, opened 2023-02-15
Add ability to cache ETL data sources

travis-newby:
This change adds data set caching to the ETL pipeline, saving the downloading and unpacking of over 5.6GB of data per pipeline run! (See: related [ADR](https://github.com/usds/justice40-tool/issues/2168).)

### Usage

Callers of certain features in application.py now have the option to use cached data for ETLs. Simply pass `--use-cache` or `-u`, and the system will use cached data if it exists. Those features include,

1. `census-data-download`
2. `etl-run`
3. `score-full-run`
4. `data-full-run`

For example, `poetry run python3 data_pipeline/application.py etl-run -u`.

Callers can also: pre-fetch the data sources by calling application.py with `extract-data-sources`; clear cached data sources by calling application.py with `clear-data-source-cache`; and print the data sources by calling application.py with `print-data-sources`.

Cached files are stored in a new folder called data_pipeline/data/sources. This involved a fair bit of standardization work in the ETLs, because not all files were stored consistently or in a single location prior to this change (and some files were used directly from memory after being downloaded).

The caching algorithm is pretty simple: if there's something in the cache folder for the ETL, the system tries to use that something. If not, the ETL will likely fail. To refresh an ETLs data source, all you have to do is call `extract-data-sources` on that ETL without the caching option.

### Why not make caching the default?

I don't want this change to impact the Github Actions or anyones existing workflow. I do think this may be an incremental change to caching more broadly, but that those iterations should be done as a separate PR. This one is already big enough.

We may decide, in the future, to check the data_pipeline/data/sources folder into source control, or to allow the folder to be downloaded in zipped form from S3; this will give us consistency between _every_ pipeline run, and allow us to vet and incorporate new data sources while understanding their impact on the score. In that case, caching would likely be turned on by default (if you can't tell, this would be my ultimate recommendation, with this PR simply introducing the caching mechanism into the code to be better used later).

### But what about the results of the pipeline?

I think we're good! I took time to diff the results of main vs the results of this branch. Specifically, I looked at,

* data_pipeline/data/dataset to make sure the ETL results are consistent. They are.
* data_pipeline/data/score/csv, downloadable, and geojson to make sure those are consistent. They are.
* data_pipeline/data/score/shapefile to make sure those are consistent. They are with a single exception: an extra "newline" character is in the .dbf file of my branch. I have no idea why, but it doesn't seem to impact anything.
* data_pipeline/data/score/tiles to make sure those are consistent. The tiles are the same.

I don't have the ability to diff binary files, so a few of those were skipped. But the important stuff, like the CSV files and the tiles themselves, are identical. 

### And the tests? Bet you forgot about those...

I didn't forget! Previously, because there was no concept of caching, tests used a variety of directories to store their temporary contents. Now, like in the main application, we've standardized on a location for mocked test downloads (you can find it under data_pipeline/data/tmp/tests). Due to the way the tests work (magic constants!), this mean mocking the `get_sources_path` call in the ETL base class, and modifying a few tests to use that path. I believe I got all this right, and the tests run and pass with everything ending up in the right place, but would love a second set of eyes.

### Any other comments on the way the code is structured?

Yep! To make it easier to review, here are a few things to keep in mind.

1. I introduced a new concept to the ETLs called "data sources." Each ExtractTransformLoad subclass supplies a list of data sources it requires to run by implementing the `get_data_sources` method. The superclass implementation of `extract` automatically downloads those data sources, and child classes can call `super().extract(...)` to inherit that behavior. 
2. Data sources are implemented in `data_pipeline.etl.datasource.py` and come in three flavors: a `FileDataSource` used to download single files, a `ZIPDataSource` used to download and extract zip files, and a `CensusDataSource` that calls the Census API. Each of those data sources knows how to retrieve its own content. 
3. I also created a class called `Downloader` to isolate the downloading behavior for ZIP and regular files. 
4. I tried to standardize the work that was performed in the `extract` methods. It was not at all consistent, with some ETLs performing almost no work in extract and some performing most of their work in extract. I'm not sure the balance is perfect yet, but it's better.

Thanks for making it this far. 🚀


data/data-pipeline/data_pipeline/etl/datasource.py
```diff
@@ -0,0 +1,112 @@
...
+from dataclasses import dataclass
+from abc import ABC, abstractmethod
+
+from data_pipeline.etl.downloader import Downloader
+from data_pipeline.etl.sources.census_acs.etl_utils import (
+    retrieve_census_acs_data,
+)
+
+
+class Fetchable(Protocol):
```
mattbowen-usds, line 12: I love protocols, but this isn't used anywhere and I think the ABC would fulfill the same purposes for typechecking?

data/data-pipeline/data_pipeline/etl/downloader.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        # file_id allows us to evade race conditions on parallel ETLs
+        file_id = uuid.uuid4()
+        dir_id = uuid.uuid4()
+
+        zip_download_path = (
+            settings.DATA_PATH
+            / "tmp"
+            / "downloads"
+            / f"{dir_id}"
```
mattbowen-usds, line 82: double-uuid! no one is finding that file. For real tho, do you actually need double UUIDs?  like the file could always be called `download.zip`

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
```diff
@@ -17,59 +18,74 @@
...
 
-    raw_df: pd.DataFrame
-    output_df: pd.DataFrame
+    raw_df: pd.DataFrame  # result of extraction
+    output_df: pd.DataFrame  # result of transformation
 
     def __init__(self):
+
+        # fetch
+        if settings.DATASOURCE_RETRIEVAL_FROM_AWS:
```
mattbowen-usds, line 43: So FWIW, this setting is the previous attempt to "cache". I sorta think asking for a cached version of the data implies also grabbing from our AWS FWIW.

data/data-pipeline/data_pipeline/etl/datasource.py
```diff
@@ -0,0 +1,112 @@
...
+            unzipped_file_path=self.destination,
+            verify=True,
+        )
+
+    def __str__(self):
+        return f"Zip – {self.source}"
+
+
+@dataclass
+class CensusDataSource(DataSource):
```
mattbowen-usds, line 84: It would be fun and nice if, in your cache design, a change to the variables here automatically invalidated the cache. I think in your current design, that would require you to bake them in (maybe as a hash) to the directory name?

mattbowen-usds (commented):
I haven't reviewed this with the level of detail it deserves, but I wanted to say (a) I like and support the approach, (b) I would like it even more if it was usable in prod, and (c) I do have some specific feedback on the code I wanted to give sooner than later. Nice job tho! I like this overall a lot.

data/data-pipeline/data_pipeline/etl/datasource.py
travis-newby: I think you're right. I had bigger plans but scrapped them for simplicity. This is a vestigial tail that can be removed.

data/data-pipeline/data_pipeline/etl/datasource.py
```diff
@@ -0,0 +1,112 @@
+from pathlib import Path
```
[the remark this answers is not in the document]
travis-newby: Good point. Thanks.

data/data-pipeline/data_pipeline/etl/downloader.py
travis-newby: Prooooobably. 😄 I will say the point is for no one to find the file. This is a temporary path for downloading, and the file is moved to the actual destination later. The first UUID is fixing the fact that previously you had to pass in the temp directory, and I thought "nobody cares what the temp directory is, they just care that the file makes it to the final destination." The second UUID is inherited and can just be removed. So good catch.

data/data-pipeline/data_pipeline/etl/sources/cdc_life_expectancy/etl.py
travis-newby: I view this PR as the first in a series of potential PRs to clear up a lot of this. Because you're right... we have an AWS cache-ish-type-thing and a local cache. I actually think the right thing to do is to get rid of the AWS cache and source control these data sources.

This isn't a normal ETL where you don't care what happens to your sources after you ETL them. And I think the architecture and implementation should ultimately reflect the fact that score consistency between runs and understanding the impact of each data source change are some of the more important parts of this process.

data/data-pipeline/data_pipeline/etl/datasource.py
travis-newby: See discussion above. TLDR – I'd like to leave that to a future PR.

vim-usds:
I'd like to re-run this with the score-comparator bot :)

travis-newby:
> I'd like to re-run this with the score-comparator bot :)

Me, too! Just kicked it off!

travis-newby:
Looks like this branch all the same deltas from the 1.0 score as main, which are small variances in the column `Percent of the Census tract that is within Tribal areas (percentile) `.

vim-usds (approved):

travis-newby:
Running this with the new comparator one last time, then I'll merge.

PR #2179, opened 2023-03-01
Add several factors to comparator, including number of disadvantaged tracts and population

travis-newby:
This is a small update to the comparator to include the number of disadvantaged tracts calculated in both scores, and the populations (as a % of the total population) of those disadvantaged tracts. It also includes a count of the census tracts newly marked as disadvantaged, as well as those that were marked as disadvantaged but have now changed.


vim-usds (approved):
LGTM!

PR #2183, opened 2023-03-06
Update IPython dependency to >= version 8.11

travis-newby:
This PR updates the IPython dependency from 7 to 8.11. The upgrade from 7 to 8 was mainly a minimum Python version upgrade, with the project dropping support for everything below Python 3.8. That has no impact on us, since we were using >= 3.8 already.

I spot-checked some of our Jyputer notebooks, and they appeared to be working (those that were not working were failing for reasons unrelated to this change, such as Census API changes, and were failing before this upgrade).


vim-usds (approved):
LGTM!

PR #2186, opened 2023-03-06
remove release banner

### LINKED ISSUE #2184: Take down the press release banner
**Describe the bug**
Now that the release is a few months out, CEQ is asking to take down the informational banner that links to the press release.

**To Reproduce**
1. Go to 'any page'
2. See banner

**Expected behavior**

**Screenshots**

**Additional context**

vim-usds:
- closes #2184

### DIFF
--- a/client/src/components/J40Header/J40Header.tsx
+++ b/client/src/components/J40Header/J40Header.tsx
@@ -1,7 +1,7 @@
 import React, {useEffect, useState} from 'react';
 import {Link, useIntl} from 'gatsby-plugin-intl';
 import {
-  Alert,
+  // Alert,
   Header,
   NavMenuButton,
   PrimaryNav,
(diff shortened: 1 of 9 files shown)


travis-newby (approved):
Approved, with a minor nit – feel free to address or ignore: why just comment out the Alert, and not remove it? We could always go back in history and retrieve it if we need to.

vim-usds:
> Approved, with a minor nit – feel free to address or ignore: why just comment out the Alert, and not remove it? We could always go back in history and retrieve it if we need to.

This is a valid point. I go back and forth on this. I believe I did remove this in the past, and then forgot which file had Alert (as it could in the Header or Layout). I started re-implementing it (this might be because I may not know how to search the history effectively). I then decided that for things like an Alert (which is often brought in and out) commented out allows me to quickly find and make the change.

Out of curiosity, is searching for this change easy to do if one doesn't know the file name that made the change?

travis-newby:
That makes more sense for something that comes and goes frequently. And you're right, git history search stinks.

Issue #2189, opened 2023-03-07
EJScreen Link Broken in CEJST


travis-newby:
**Describe the bug**
The EPA site https://www.epa.gov/ejscreen/technical-documentation-ejscreen is referenced multiple times on the CEJST Methodology and Data page. Following that link results in a 404 (page does not exist).

**To Reproduce**
1. Open the CEJST
2. Navigate to the Methodology and Data page.
3. Select `Fusion of model and monitor data`, `RMP database`, `National Air Toxics Assessment`, `Traffic Data`, or `Risk-Screening Environmental Indicators (RSEI) model`.
4. Notice the 404.

**Expected behavior**
All links on that page should point to an active website. I suggest we link to something less likely to go away, such as https://www.epa.gov/ejscreen or https://www.epa.gov/ejscreen/technical-information-about-ejscreen.

travis-newby:
@KameronKerger curious where you think we should link.

KameronKerger:
they must have moved it. what do you think about linking here - https://www.epa.gov/ejscreen/download-ejscreen-data

travis-newby:
I think that's fine.

PR #2190, opened 2023-03-07
Update minor frontend dependencies

travis-newby:
This PR updates _minor_ FE dependencies to address issues pointed out by dependabot and by `npm audit`. (It does not tackle the larger ones, such as Gatsby.) It also fixes a few broken Cypress tests that popped up while updating the dependencies (the outstanding Cypress test failure is addressed in #2189).

### DIFF
--- a/client/cypress/integration/Explore.feature
+++ b/client/cypress/integration/Explore.feature
@@ -2,15 +2,15 @@ Feature: The Explore the map page will open from all other pages
   
   Scenario: Explore the map page open when navigating from About page
     Given I am on the "About" page
-    When I click on the "Explore the map" page in the navigation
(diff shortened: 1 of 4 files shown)


vim-usds:
> This PR updates _minor_ FE dependencies to address issues pointed out by dependabot and by `npm audit`. (It does not tackle the larger ones, such as Gatsby.) It also fixes a few broken Cypress tests that popped up while updating the dependencies (the outstanding Cypress test failure is addressed in #2189).

Great catch! Being away from J40, I forgot about these cypress tests, even though some took quite bit of time to write. Thanks for bringing this back on the radar for me. 

PS - they used to be part of the CI/CD pipeline, however there was some error causing me to remove them. I'll try and find the ticket that captures this.

PR #2194, opened 2023-03-08
Update create-score-version Github Action

travis-newby:
This PR updates the create-score-version Github Action to,

* Separate the action into a few logical steps, and organize the action around those logical steps. This fixes a potential issue where the backend score could be deployed then fail a smoke test, resulting in no tiles being built or uploaded.
* Use the `test` input option to upload everything to a new test folder and run smoke tests from there. That way we can test this script without deploying version 1.0 again.
* Separate out the "extra" uploaded files (from `data_pipeline/files`) to be in their own folder uniquely tied to the version being deployed. So there's now a `beta`, `1.0`, and `test` folder under `files`, and we don't need to have special logic to upload files for each new version of the tool (we just have to update the files in the files directory).


.github/workflows/create-score-version.yml
```diff
@@ -115,11 +76,64 @@ jobs:
...
           poetry run s4cmd put ./data_pipeline/data/score/tiles/ s3://justice40-data/data-versions/${{env.J40_VERSION_LABEL_STRING}}/data/score/tiles --recursive --force  --API-ACL=public-read --num-threads=250
+
+      # Run more smoke tests to ensure we deployed everything correctly
+      - name: Confirm we deployed downloadables to AWS
+        if: ${{ env.J40_VERSION_LABEL_STRING != 'beta' }}
+        run: |
+          curl "https://static-data-screeningtool.geoplatform.gov/data-versions/${{env.J40_VERSION_LABEL_STRING}}/data/score/downloadable/1.0-shapefile-codebook.zip" -s -f -I -o /dev/null && \
+          curl "https://static-data-screeningtool.geoplatform.gov/data-versions/${{env.J40_VERSION_LABEL_STRING}}/data/score/downloadable/1.0-communities.xlsx" -s -f -I -o /dev/null && \
+          curl "https://static-data-screeningtool.geoplatform.gov/data-versions/${{env.J40_VERSION_LABEL_STRING}}/data/score/downloadable/1.0-communities.csv" -s -f -I -o /dev/null && \
+          curl "https://static-data-screeningtool.geoplatform.gov/data-versions/${{env.J40_VERSION_LABEL_STRING}}/data/score/downloadable/1.0-shapefile-codebook.zip" -s -f -I -o /dev/null && \
```
vim-usds, line 127: We may have repeated the curl smoketest for the `1.0-shapefile-codebook.zip` file

Also I believe the following files are used by 1.0 and could be added to the smoketest:

1.0-cejst-technical-support-document.pdf
CEQ-CEJST-Instructions.pdf
1.0-communities-list.pdf
1.0-climate-and-environmental-burden-tool-comparison.pdf
1.0-codebook.csv
1.0-communities-csv.zip
1.0-communities-xls.zip
1.0-data-documentation.zip

.github/workflows/create-score-version.yml
```diff
@@ -115,11 +76,64 @@ jobs:
...
+          curl "https://static-data-screeningtool.geoplatform.gov/data-versions/${{env.J40_VERSION_LABEL_STRING}}/data/score/downloadable/cejst-technical-support-document.pdf" -s -f -I -o /dev/null && \
```
vim-usds, line 128: These two files namely, `cejst-technical-support-document.pdf` and `draft-communities-list.pdf` can be removed as they are part of beta

.github/workflows/create-score-version.yml
```diff
@@ -115,11 +76,64 @@ jobs:
...
+      - name: Run smoketest
+        if: ${{ env.J40_VERSION_LABEL_STRING != 'beta' }}
+        run: |
+          poetry run pytest data_pipeline/ -m smoketest
+      
+      # Deploy to AWS
+      - name: Deploy Score to Geoplatform AWS
+        run: |
+          poetry run s4cmd put ./data_pipeline/data/score/csv/ s3://justice40-data/data-versions/${{env.J40_VERSION_LABEL_STRING}}/data/score/csv --recursive --force --API-ACL=public-read
+          poetry run s4cmd put ./data_pipeline/files/${{env.J40_VERSION_LABEL_STRING}}/ s3://justice40-data/data-versions/${{env.J40_VERSION_LABEL_STRING}}/data/score/downloadable --recursive --force --API-ACL=public-read
```
vim-usds, line 112: It seems that the [`files`](https://github.com/usds/justice40-tool/tree/main/data/data-pipeline/data_pipeline/files) folder is for _static files_ (ie files that are to be downloadable from the client and not created by the pipeline) while the [`downloadable`](https://github.com/usds/justice40-tool/tree/main/data/data-pipeline/data_pipeline/data/score/downloadable) folder is for files that _are_ created by the pipeline. These folder names don't make it obvious to me that this is how they operate. I wonder if changing them makes sense?

These are the _static files_ _for 1.0_:

[1.0-cejst-technical-support-document.pdf](https://static-data-screeningtool.geoplatform.gov/data-versions/1.0/data/score/downloadable/1.0-cejst-technical-support-document.pdf)
[CEQ-CEJST-Instructions.pdf](https://static-data-screeningtool.geoplatform.gov/data-versions/1.0/data/score/downloadable/CEQ-CEJST-Instructions.pdf)
[1.0-communities-list.pdf](https://static-data-screeningtool.geoplatform.gov/data-versions/1.0/data/score/downloadable/1.0-communities-list.pdf)
[1.0-climate-and-environmental-burden-tool-comparison.pdf](https://static-data-screeningtool.geoplatform.gov/data-versions/1.0/data/score/downloadable/1.0-climate-and-environmental-burden-tool-comparison.pdf)

The files currently in `files` are for `beta` only ( I think). Matter of fact they may not be used anywhere anymore. I believe these files are strictly for the client and since the client will not likely revert to the beta version, maybe it's safe to remove these and replace with the 1.0 static files? In which case my next suggestion of changing `files` to `files/beta` and `files/1.0` may be moot.

Lastly, given that these CLI are PUTs, _not_ having these 1.0 files won't break anything, however if S3 bucket contents are lost for any reason, maybe it's a good thing to have these 1.0 static files in source control. 

What do you think?

vim-usds (commented):
Thanks for getting this started and improving the flow! Wrote up a few ideas on ways to potentially improve. Let's discuss :)

.github/workflows/create-score-version.yml
travis-newby: There are beta versions and 1.0 versions. I added the 1.0 version to source control under /files/1.0/.

.github/workflows/create-score-version.yml
travis-newby: Yes, and. 🤣 Just for the historical record, _some_ of the files in the files folder are repackaged into {version}-data-documentation.zip, and are uploaded to S3 (e.g. communities-list.pdf and cejst-technical-support-document.pdf).

.github/workflows/create-score-version.yml
travis-newby: Great catch! I'll update the test.

vim-usds:
@travis-newby -  I don't believe the cache invalidation is in the script. We could place it in another ticket if you think that's better. 

Just set this up [here](https://github.com/usds/ej-scorecard/blob/main/.github/workflows/deploy-prod.yml#L91).

As an FYI, the BE distribution ID is E1324VDMNCO97N

vim-usds (approved):
LGTM!

vim-usds:
Also a fast follow will be to update the FE links once these files are on S3 :)

PR #2197, opened 2023-03-10
Consolidate columns.csv data into codebook

travis-newby:
This PR consolidates the data that was in `/data/score/shapefile/columns.csv` into `/data/score/downloadable/xxx-codebook.csv`. We now have a single codebook with column names (the long ones), shapefile names (the short ones), and column definitions.

To avoid confusing people, we no longer include the `columns.csv` file in our zips (e.g. `xxx-shapefile-codebook.zip`); in fact, we don't even write it to disk anymore. We instead include the full codebook in our zip files. This should result in a more consistent and easier experience for consumers of our data.

On the technical side, the regular codebook is written during `generate-score-post`. During `geo-score` we add the shortened column names (the data that was previously in `columns.csv`) to the codebook, then package it all up in zip files. I updated the logs to help developers follow along with the process.

I don't think this will require any client changes, since these files were just included inside other zip files on the CEJST site.

### DIFF
--- a/data/data-pipeline/data_pipeline/application.py
+++ b/data/data-pipeline/data_pipeline/application.py
@@ -268,7 +268,10 @@ def geo_score(data_source: str):
     Returns:
         None
     """
-    log_title("Generate GeoJSON", "Combine Score and GeoJSON")
+    log_title(
+        "Generate GeoJSON",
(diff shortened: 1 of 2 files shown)


vim-usds (approved):
Awesome comments! Thanks!

PR #2201, opened 2023-03-15
Update Client packages

travis-newby:
This PR updates a handful of client packages to fix dependabot alerts. The long and short of it is that I ran `npm audit fix`, which addressed the two minor alerts (webpack and json5) that don't involve Gatsby. It also bumped minor versions of 5 other packages, which I'm sure dependabot would be complaining about soon.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -3062,9 +3062,9 @@
       }
     },
     "@types/eslint-scope": {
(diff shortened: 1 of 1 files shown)


travis-newby:
I spot checked some communities and everything looks the same.

vim-usds (approved):
LGTM!

PR #2254, opened 2024-01-31
Update survey links

lscharen:
Update to switch away from SurveyMonkey to Qualtrics survey links.  The survey URLs have been pulled into the constants file to easier reuse moving forward.


lscharen:
@vim-usds This PR now includes an updated Spanish data survey link from @KameronKerger

